PR #10, opened 2016-01-16
Use POSIX paths for image directive

### LINKED ISSUE #9: Maplotlib figures do not render
After building, it seems my plots for matplotlib are not rendering.  In place of the images are paths to files in the doctrees folder, e.g:

_build%5Cdoctrees%5Cnbsphinx%5Cshowcase_121_1.png

Those images do exist at these paths, but are not being rendered somehow `make html`.  The errors says these files are not readable, which is odd since they are generated from the notebook.  Any thoughts?

mgeier:
This hopefully fixes #9.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -98,7 +98,7 @@ RST_TEMPLATE = """
 {{ outputdata.strip(\n) | indent }}
 {%- elif datatype in ['image/svg+xml', 'image/png', 'image/jpeg'] %}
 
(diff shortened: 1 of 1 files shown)


jhconning:
That fixes it! 

This is what locally served html documents with links to .PNG files looked like in Windows before that fix:
image

After the fix, it renders correctly:
image

mgeier:
Thanks @jhconning and @par2 for testing!

Issue #3, opened 2015-12-23
SVG support for LaTeX output


mgeier:
This will hopefully be fixed soon in Sphinx itself:
https://github.com/sphinx-doc/sphinx/issues/1907
https://github.com/sphinx-doc/sphinx/pull/2166

mgeier:
See also https://github.com/ipython-contrib/jupyter_contrib_nbextensions/pull/587.

PR #18, opened 2016-02-01
hack to handle pdf figures in notebook

aebrahim:
This may not be the correct way to handle it, but it was the easiest way to fix this bug:

In some of my notebooks, I use this code block, which causes issues:

```python
from IPython.display import set_matplotlib_formats
set_matplotlib_formats('png', 'pdf')
```

Specifically, it causes the notebook output to look like this:
```json
"outputs": [
    {
     "data": {
      "application/pdf": "JVBERi0xLjQKJazcIKu6CjEgMCBvYmoKPD<etc>==\n",
      "image/png": "iVBORw0KGgoAAAANSUhEUgAAAV0AAADtCAYAAAAc<etc>=\n",
      "text/plain": [
       "<matplotlib.figure.Figure at 0x7f7c2c22f3d0>"
      ]
     },
     "metadata": {},
     "output_type": "display_data"
    }
   ],
```

When I run this with sphinx, I get the following in my built html file:
image

This basically just changes the template so instead of having an extension on the written filename of the image, it leaves it with an asterix (i.e. my_image.* instead of my_image.png). This lets sphinx choose the appropriate image based on the backend ([sphinx docs link](http://www.sphinx-doc.org/en/stable/rest.html#images))

This assumes the extension is 3 characters, which I thought was probably a fairly good assumption (png, pdf, jpg, svg), but maybe there will be exceptions I haven't thought of.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -96,9 +96,9 @@ RST_TEMPLATE = """
 {{ insert_empty_lines(outputdata) }}
 
 {{ outputdata.strip(\n) | indent }}
-{%- elif datatype in ['image/svg+xml', 'image/png', 'image/jpeg'] %}
+{%- elif datatype in ['image/svg+xml', 'image/png', 'image/jpeg', 'application/pdf'] %}
 
-    .. image:: {{ output.metadata.filenames[datatype] | posix_path }}
+    .. image:: {{ output.metadata.filenames[datatype].rsplit('.', 1)[0] + '.*' | posix_path }}
 {%- elif datatype in ['text/markdown'] %}
 
 {{ output.data['text/markdown'] | markdown2rst | indent }}


mgeier:
Thanks for your PR!
I think that a robust solution to this might be a bit more involved, though.

I think the actual bug in that case (i.e. the WARNING is shown instead of using the available PNG data) is that `filter_data_type` is used in the template. The example for that was the [RST template](https://github.com/jupyter/nbconvert/blob/master/nbconvert/templates/rst.tpl) and the [RST exporter](https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/rst.py) from `nbconvert`, which just use the default `display_data_priority`: `['text/html', 'application/pdf', 'text/latex', 'image/svg+xml', 'image/png', 'image/jpeg', 'text/markdown', 'text/plain']`, which lists PDF before PNG.

The problem for `nbsphinx` is that HTML and LaTeX output need different display priorities. BTW, `nbconvert` uses this for HTML: `['application/javascript', 'text/html', 'text/markdown', 'image/svg+xml', 'text/latex', 'image/png', 'image/jpeg', 'text/plain']` and this for LaTeX: `['text/latex', 'application/pdf', 'image/png', 'image/jpeg', 'image/svg+xml', 'text/markdown', 'text/plain']`.

I thought about using the asterisk for the image directive (as you are suggesting), but I've not gone through with it because it wasn't necessary for the use cases I was trying.

I didn't know about `set_matplotlib_formats()`, which looks interesting. Do you actually want to use the PDF data from matplotlib in the LaTeX output of Sphinx?

Or do you just want to avoid the warning in the HTML output?

mgeier:
A solution using `.*` should also consider the conversion of SVG images to PDF and vice versa. But probably we'll have to wait until this is solved in Sphinx, see #3 and #19.

aebrahim:
Thanks for taking the time for the detailed reply (and for starting this project... it's going to be really nice to be able to remove the nbconvert step from building my sphinx docs)

To answer your questions:

My issue is not just that it displays a warning in the HTML output... it doesn't give an output image at all! So I can't actually use it on the current notebooks.

I personally don't need the pdf figure in the latex sphinx output. It would be nice, but it's not really necessary. But I do need the html output to have the figures (I was using ```set_matplotlib_formats``` for one of my custom templates, not for sphinx).

Also conversion of pdf->svg and vice versa will likely have errors in the conversion. Would it be easier to simply pick the best format from those available in the notebook?

mgeier:
Yes, I think it's definitely better to take the best data from the notebook, if a reasonable candidate is available. But if not, conversion is still better than nothing, right?

Anyway, I think your little hack is innocuous enough to use it for now and do the "real" solution later.

But I think the splitting of the path could be a bit more robust, because `.jpeg` is quite common ...

What about using `output.metadata.filenames[datatype].rsplit('.', 1)[0]`?

aebrahim:
I agree that would be better. I have updated the pull request.

mgeier:
Thanks a lot, I've merged your commit.

There is now an error when running LaTeX on the docs because of the YouTube video, but this didn't work really work correctly before either.

Now do you also want to tackle the "proper" solution?

aebrahim:
Is the "proper" solution to use different display priorities depending on which sphinx backend which will be used? I'm aware of any mechanisms to do that.

mgeier:
Well, I don't know if it's really the proper solution, but my idea was to check for the "best" display format for HTML and LaTeX separately (not using `filter_data_type` but probably something similar). This could be done with a function similar to `get_empty_lines()`.
Then, the template code could check if both are the same and emit reST code for the `nboutput` directive similar to what we have now. If they are not the same, it could create separate cases with `.. only:: html` and `.. only:: latex`.
To avoid duplicate code, the actual reST commands could be created with a Jinja macro somewhat similar to `insert_empty_lines()`.

What do you think about that?
Do you have a better idea?

BTW, we don't know the backend during parsing, so we have to take care of all possible backends at the same time.

aebrahim:
I'm not exactly a sphinx expert, but that sounds like a great way to handle it to me.

mgeier:
I've created a separate issue for that (#25), feel free to submit a PR!

mgeier:
@aebrahim I've added an implementation to #25, can you please check if that works for you?

aebrahim:
Works great for HTML. However, I'm getting errors for PDF:

```
LaTeX Error: Unknown graphics extension: .only::html.
```

This is with ```sphinx.__version__ == '1.4.1'``` and nbsphinx at commit a91f8b5

aebrahim:
It looks like for the latex case, it inserted this:

```latex
\includegraphics{{../documentation/doctrees/nbsphinx/phenotype_phase_plane_5_0.pdf.}.only::html}

\includegraphics{{phenotype_phase_plane_5_1}.png}

\includegraphics{{phenotype_phase_plane_5_1}.pdf}
```

No idea why, because for the previous image, it inserted (as expected)
```latex
\includegraphics{{phenotype_phase_plane_3_0}.pdf}
```

aebrahim:
I guess this implies the trigger case is when there are two output images in the same cell?

mgeier:
@aebrahim Thanks for testing and for finding the bug!

I think this was just a problem with missing empty lines in the intermediate RST string.
I've updated #25 with d002d051be9f3c026b2be748620167251527485f, can you please check if it works now?

mgeier:
I've merged #25, if it doesn't work for you, please tell me!

mgeier:
@aebrahim, you might be interested in this: #48.

PR #30, opened 2016-04-01
Fix emtpy -> empty typo

tbekolay:
Hi, thanks for this project! We had been doing something similar, but not as robust, in our own project, so I'm very excited to switch to using `nbsphinx`. There are a few things that we need that aren't currently done in this project, so I have a series of PRs coming in. Since I notice that your repo history is linear (which we do in our project as well) I've ordered the PRs so that they can be rebased/merged easily.

This is the first PR; I noticed a typo in `nbsphinx.py` so this fixes the typo. It also adds `.ipynb_checkpoints` to `.gitignore` as these are likely to crop up when editing the notebooks in the `docs` folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__/
 build/
 dist/
 nbsphinx.egg-info/
+.ipynb_checkpoints/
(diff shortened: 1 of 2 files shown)


mgeier:
Thanks for the kind words and thanks for this PR!

I have `.ipynb_checkpoints/` in my global gitignore file, and I would recommend that to everyone, but I guess it doesn't hurt to have it in here, too.

PR #31, opened 2016-04-01
Add nbsphinx_timeout option

tbekolay:
This is the second PR. In our project, we have several cells that take quite a while to compute, so we encountered a `TimeoutError` when generating our docs.

I tried modifying the timeout through a config files in `~/.jupyter/jupyter_nbconvert_config.py`, but that does not appear to be read when import a preprocessor directory, so I had to add an option through `nbsphinx`. Not sure if there's another way to handle this? In any case, this works for us.

I also added a new notebook to the docs describing how to use the option, similar to the `allow_errors` option.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -13,11 +13,16 @@ extensions = [
 # Exclude build directory and Jupyter backup files:
 exclude_patterns = ['_build', '**.ipynb_checkpoints']
 
-# If True, the build process is continued even if an exception occurs:
-#nbsphinx_allow_errors = True
-
 # Default language for syntax highlighting (e.g. in Markdown cells)
 highlight_language = 'none'
+
+# -- These set defaults that can be overridden through notebook metadata --
+
+# See http://nbsphinx.readthedocs.org/en/latest/allow-errors.html
+# for more details.
+
+# If True, the build process is continued even if an exception occurs:
+#nbsphinx_allow_errors = True
 
 # -- The settings below this line are not specific to nbsphinx ------------
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -19,10 +19,13 @@ highlight_language = 'none'
 # -- These set defaults that can be overridden through notebook metadata --
 
 # See http://nbsphinx.readthedocs.org/en/latest/allow-errors.html
-# for more details.
+# and http://nbsphinx.readthedocs.org/en/latest/timeout.html for more details.
 
 # If True, the build process is continued even if an exception occurs:
 #nbsphinx_allow_errors = True
+
+# Controls when a cell will time out (defaults to 30; use -1 for no timeout):
+#nbsphinx_timeout = 60
 
 # -- The settings below this line are not specific to nbsphinx ------------
 
--- a/doc/timeout.ipynb
+++ b/doc/timeout.ipynb
@@ -1,0 +1,62 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "nbsphinx": "hidden"
+   },
+   "source": [
+    "This notebook is part of the `nbsphinx` documentation: http://nbsphinx.readthedocs.org/."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Long-Running Cells\n",
+    "\n",
+    "By default, `nbconvert` will give a cell 30 seconds to execute before it times out.\n",
+    "\n",
+    "If you would like to change the amount of time given for a cell, you have three options:\n",
+    "\n",
+    "1. Manually execute the notebook in question and save the results, see [the pre-executed example notebook](pre-executed.ipynb#Long-Running-Cells).\n",
+    "\n",
+    "2. Change the timeout length for all notebooks by setting this option in [conf.py](conf.py):\n",
+    "```\n",
+    "nbsphinx_timeout = 60  # Time in seconds; use -1 for no timeout\n",
+    "```\n",
+    "\n",
+    "3. Change the timeout length on a per-notebook basis by adding this to the notebook's JSON metadata:\n",
+    "```json\n",
+    "\"nbsphinx\": {\n",
+    "  \"timeout\": 60\n",
+    "}\n",
+    "```"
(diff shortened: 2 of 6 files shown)


nbsphinx.py
```diff
@@ -386,6 +388,8 @@ class NotebookParser(rst.Parser):
 
     """
 
+    timeout = 30  # Default timeout
```
mgeier: What does this do?
Is there some magic in the `rst.Parser` class?

tbekolay: Oh whoops, this was leftover from an earlier attempt and I meant to take it out. Will amend!

tbekolay:
Amended!

nbsphinx.py
```diff
@@ -360,8 +361,9 @@ def from_notebook_node(self, nb, resources=None, **kw):
                 not any(c.outputs for c in nb.cells if 'outputs' in c)):
             allow_errors = (self._allow_errors or
                             nbsphinx_metadata.get('allow_errors', False))
+            timeout = nbsphinx_metadata.get('timeout', self._timeout)
```
mgeier: Hmmm, should the `nbsphinx_timeout` value take precedence or the notebook metadata if both are given?

This line sure looks nicer than my 2 clumsy lines directly above, but they also have slightly different behavior, if I'm not mistaken ...

Either way, I think both `allow_errors` and `timeout` should behave in the same way, right?

doc/timeout.ipynb
```diff
@@ -0,0 +1,60 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Long-running cells\n",
+    "\n",
+    "By default, `nbconvert` will give a cell 30 seconds to execute before it times out.\n",
+    "\n",
+    "If you would like to change the amount of time given for a cell, you have three options:\n",
+    "\n",
+    "1. Manually execute the notebook in question and save the results, see [the pre-executed example notebook](pre-executed.ipynb#Exceptions).\n",
```
mgeier: Could you please add a link in the other direction, too?

doc/timeout.ipynb
```diff
@@ -0,0 +1,60 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
```
mgeier: Can you please hide this cell with `"nbsphinx": "hidden"`?

nbsphinx.py
tbekolay: Yes, you're right, they are different. I'm happy to change, though we can't use the `or` idiom here since the default is an integer.

Perhaps the notebook metadata should take precedence? It feels to me like the notebook-local setting should take precedence over the global setting so that even when the global setting is in place notebooks can have specific behavior. What do you think?

doc/conf.py
```diff
@@ -16,6 +16,9 @@
 # If True, the build process is continued even if an exception occurs:
 #nbsphinx_allow_errors = True
 
+# Controls when a cell will time out:
+#nbsphinx_timeout = 30  # Defaults to 30
```
mgeier: I think I would prefer to mention the default value in the line before and use a different (non-default) value in the commented-out example (and no further comment on the commented-out line).

doc/timeout.ipynb
```diff
@@ -0,0 +1,60 @@
...
+   "source": [
+    "This notebook is part of the `nbsphinx` documentation: http://nbsphinx.readthedocs.org/."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Long-running cells\n",
```
mgeier: cells -> Cells?

I'm not a native speaker, so I'm not sure it that's proper, but that's what I used in all other section titles ...

doc/timeout.ipynb
mgeier: Also, I guess you wanted to link to `#Long-Running-Cells`, right?

nbsphinx.py
mgeier: I don't know.

I think cases can be made for both.

In your argumentation, `conf.py` sets "default" values.

However, one could also argue that those should "override" the notebook-local settings.

Either way, I think typically people will use either one or the other, but seldom both. So I guess in most cases it won't actually matter.

I think we should use your suggestion because the implementation is cleaner.
To make the behavior very clear, we should probably mention in the comments in `conf.py` that the settings can be overridden on a per-notebook basis.
We could also include those links:
http://nbsphinx.rtfd.org/en/latest/allow-errors.html
http://nbsphinx.rtfd.org/en/latest/timeout.html

If you want to, you can make those changes, if not, I'll make them once your PRs are merged.

tbekolay: I agree on both comments: cases can be made for both, and people will tend to use one or the other. I'll update the behavior and update the `conf.py` and the notebooks :)

doc/timeout.ipynb
tbekolay: Yep, you're right. Will fix!

doc/timeout.ipynb
tbekolay: My apologies, I didn't look at the capitalization on the other titles. I'll update this to `Long-Running Cells` to be consistent!

mgeier: I had a look at the other titles and I'm wondering how I should handle capitalization after hyphens ...

Currently, I have "Built-in" and "3rd-party" on the one side and "Pre-Executed", "Long-Running", "Sub-Directory" and "Sub-Section" on the other side.

Which do you think is better?

tbekolay: Googling around a bit, there are a few style guides recommending various things... In one style, the test seems to be "would this word be understandable on its own?" and if so capitalize it. Another style says "no capital unless it would be capitalized in a normal sentence (e.g., proper nouns)."

The second style seems simpler to me, but I don't have a real preference, just to be consistent :) For now, I'll do `Long-Running` to be consistent, but we can change it later on.

tbekolay:
OK, I think I've addressed all your comments. I change the precedence for notebook-local and global settings in a separate commit (7243014002) to make that change more prominent.

mgeier:
Thanks a lot!

PR #32, opened 2016-04-01
Add option for changing default language

tbekolay:
This is the third PR. Our project has been building IPython notebooks in docs for quite a while, so currently our notebooks are still in version 3 of the notebook format. Generally this is fine and they get converted successfully in the build process, but syntax highlighting doesn't work because there is no `language_info` metadata generated in the conversion process. Maybe this is an nbconvert bug, but in any case, this PR makes it possible to set a default language that will be used if no language is found in the notebook metadata.

The implementation of this is definitely more of a hack than the other PRs, so if you have any thoughts of other ways to do this, I'm happy to change it! We will update our notebooks to v4 eventually, but this strikes me as a useful thing in other weird cases anyhow.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -13,7 +13,7 @@ extensions = [
 # Exclude build directory and Jupyter backup files:
 exclude_patterns = ['_build', '**.ipynb_checkpoints']
 
-# Default language for syntax highlighting (e.g. in Markdown cells)
+# Default language for syntax highlighting in reST and Markdown cells
 highlight_language = 'none'
 
 # -- These set defaults that can be overridden through notebook metadata --
@@ -26,6 +26,9 @@ highlight_language = 'none'
 
 # Controls when a cell will time out (defaults to 30; use -1 for no timeout):
 #nbsphinx_timeout = 60
+
+# Default Pygments lexer for syntax highlighting in code cells
+#nbsphinx_codecell_lexer = 'ipython3'
 
 # -- The settings below this line are not specific to nbsphinx ------------
 
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -65,6 +65,8 @@ RST_TEMPLATE = """
 {% block input -%}
 .. nbinput:: {% if nb.metadata.language_info -%}
 {{ nb.metadata.language_info.pygments_lexer }}
+{%- else -%}
+{{ resources.codecell_lexer }}
 {%- endif -%}
 {{ insert_empty_lines(cell.source) }}
 {%- if cell.execution_count %}
@@ -336,10 +338,11 @@ class Exporter(nbconvert.RSTExporter):
 
     """
 
-    def __init__(self, allow_errors=False, timeout=30):
+    def __init__(self, allow_errors=False, timeout=30, codecell_lexer='none'):
         """Initialize the Exporter."""
         self._allow_errors = allow_errors
         self._timeout = timeout
+        self._codecell_lexer = codecell_lexer
         loader = jinja2.DictLoader({'nbsphinx-rst.tpl': RST_TEMPLATE})
         super(Exporter, self).__init__(
             template_file='nbsphinx-rst', extra_loaders=[loader],
@@ -354,6 +357,9 @@ class Exporter(nbconvert.RSTExporter):
             resources = {}
         else:
             resources = copy.deepcopy(resources)
+        # Set default codecell lexer
+        resources['codecell_lexer'] = self._codecell_lexer
+
         nbsphinx_metadata = nb.metadata.get('nbsphinx', {})
 
         # Execute notebook only if there are code cells and no outputs:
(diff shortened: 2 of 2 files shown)


mgeier:
Thanks for this PR!

I was thinking about this before, but didn't do anything because I wanted to wait until this actually comes up. And alas, now it does!

I think this can be done in the Jinja template in a much easier way, by extending:

```jinja
.. nbinput:: {% if nb.metadata.language_info -%}
{{ nb.metadata.language_info.pygments_lexer }}
{%- endif -%}
```

... with an "else" section. Or am I missing something?

The default value would have to be passed into the template via `resources`.

But probably this isn't simpler after all? I didn't try it ...

tbekolay:
That was my original thought as well, but I wasn't sure how to pass things into the template. But, I've done some reading now, so I'll change this accordingly :)

tbekolay:
OK, I've updated this PR to pass in the default language through `resources`. Is this what you had in mind? It works for me locally.

I added a line to `conf.py` to document this, but didn't add it anywhere in the docs. Any idea where we should document this option?

mgeier:
Thanks for the update, this looks good!

I'm just not sure if `nbsphinx_language` is a good name ...

There are many options containing the word "language", maybe we should be more specific here?

Is there a way to make the distinction between `nbsphinx_language` and `highlight_language` clearer?

About the docs:
I consider `conf.py` part of the documentation and I think it is sufficient to document it only there.

doc/conf.py
```diff
@@ -27,6 +27,9 @@
 # Controls when a cell will time out (defaults to 30; use -1 for no timeout):
 #nbsphinx_timeout = 60
 
+# Default language for syntax highlightin in notebooks
```
mgeier: highlighting

tbekolay:
Fixed the typo! I agree, the word language is quite ambiguous and we should be more specific.

`highlight_language` is defined by sphinx, so we can't change it, correct? Here are a few alternatives for `nbsphinx_language` off the top of my head:

- `nbsphinx_codecell_language` / `nbsphinx_codecells` (makes it more clear what it affects)
- `nbsphinx_pygments_lexer` / `nbsphinx_lexer` (since this is what it's called in notebook metadata)
- `nbsphinx_kernel_language` / `nbsphinx_kernel` (since we're basically figuring out what kernel was used -- this could be useful for things other than syntax highlighting too maybe?)

I think of these I like `nbsphinx_codecells` the best... seems the most explicit. What do you think?

mgeier:
Yes, `highlight_language` [is defined by Sphinx](http://www.sphinx-doc.org/en/stable/config.html#confval-highlight_language).

I like "codecell" as part of the name, since this differentiates from "code in Markdown cells".

> we're basically figuring out what kernel was used

I don't think the Kernel name is necessarily the same as the `pygments_lexer`. There are surprisingly many language-related names stored in the notebook metadata, e.g.:

```json
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.5.1+"
  }
 }
```

I guess the actual kernel name is `metadata.kernelspec.name` ...?

If we need an option for selecting a default kernel, this should be a separate option.
But I hope we don't need that ...

In the notebook format version 3, the information seems to be in each code cell:

```json
{
 "cell_type": "code",
 "collapsed": false,
 "input": [
  "# Create the simulator\n",
  "with nengo.Simulator(model) as sim:\n",
  "    # Run it for 5 seconds\n",
  "    sim.run(5)"
 ],
 "language": "python",
 "metadata": {},
 "outputs": []
}
```

Therefore, I think we won't need a way to specify the actual kernel to be used.

Anyway, I don't care if the option name is long and ugly, since it will be hardly ever typed by anyone.

E.g. `nbsphinx_codecell_lexer` or even `nbsphinx_codecell_pygments_lexer` would be OK for me.

doc/conf.py
```diff
@@ -27,6 +27,9 @@
 # Controls when a cell will time out (defaults to 30; use -1 for no timeout):
 #nbsphinx_timeout = 60
 
+# Default language for syntax highlighting in code cells
+#nbsphinx_language = 'python'
```
mgeier: You should probably use `'ipython'` or `'ipython2'` or `'ipython3'` for the example?

tbekolay:
Agreed, this should be rarely used. I went with `nbsphinx_codecell_lexer` and changed the example!

nbsphinx.py
```diff
@@ -336,10 +338,11 @@ class Exporter(nbconvert.RSTExporter):
 
     """
 
-    def __init__(self, allow_errors=False, timeout=30):
+    def __init__(self, allow_errors=False, timeout=30, lexer='none'):
         """Initialize the Exporter."""
         self._allow_errors = allow_errors
         self._timeout = timeout
+        self._lexer = lexer
```
mgeier: I hope I'm not too picky, but wouldn't it be better for symmetry to name the variables like this:

    self._codecell_lexer = codecell_lexer

?

nbsphinx.py
```diff
@@ -837,6 +844,7 @@ def setup(app):
 
     app.add_config_value('nbsphinx_allow_errors', False, rebuild='')
     app.add_config_value('nbsphinx_timeout', 30, rebuild='')
+    app.add_config_value('nbsphinx_codecell_lexer', 'none', rebuild='')
```
mgeier: When this changes, the notebooks have to be re-parsed, right?

I think this should be `rebuild='env'`.

mgeier:
Thanks a lot!

PR #33, opened 2016-04-01
Add notebook directive

tbekolay:
This is the fourth and final PR. In our project, the notebooks that we include in the documentation are examples of using our library. We have the examples in the top-level of the repository so that people visiting the repo see them front-and-center and can look through them. Unfortunately, since Sphinx doesn't allow `toctree` entries that are not in the root folder of the documentation, we can't include these examples simply by adding them to `toctree` entries.

Similar to how files like `README.rst` can be added with the `include` directive, this PR adds a `notebook` directive that you can point at a notebook outside of the `doc` directory. It also allows you to have one `.rst` file that includes multiple notebooks, which I can imagine being useful.

The actual implementation of the directive is basically how `docutils`'s `include` directive works, but then uses the `NotebookParser` to insert blocks into the document. 4fd8fe813399d1 shows how this works by eliminating the duplication between `doc/index.rst` and `doc/index.ipynb`. It seems like bi-directional communication works perfectly; now reST files can have notebooks in them, while notebooks can have raw reST cells in them! It's very nice, and definitely not something we were able to do with our previous implementation.

One issue with the last commit is that when I build the Sphinx docs, I don't get links in the left sidebar, so I may be missing something critical in this PR. Similarly, the `subdir/toctree.ipynb` notebook doesn't show up, even though it is included in `index.ipynb`. Any idea why that might be?

### DIFF
--- a/doc/rst.rst
+++ b/doc/rst.rst
@@ -36,6 +36,26 @@ These links were created with:
     * "``../``" is not allowed, you have to specify the full path even if the
       current source file is in a subdirectory!
 
+Sphinx Directive for Including Notebooks
+----------------------------------------
+
+Notebooks can also be included directly in an RST file
+using the ``nbinclude`` directive.
+Below, we include :ref:`subdir2/included.ipynb`.
+
+.. nbinclude:: subdir2/included.ipynb
+
+The notebook was included with:
+
+.. code-block:: rst
+
+    .. nbinclude:: subdir2/included.ipynb
+
+.. note::
+
+    * "``../``" is allowed here. In fact, you can include notebooks that are
+      outside of the documentation directory!
+
 Sphinx Directives for Jupyter Notebook Cells
 --------------------------------------------
 
--- a/doc/rst.rst
+++ b/doc/rst.rst
@@ -36,26 +36,6 @@ These links were created with:
     * "``../``" is not allowed, you have to specify the full path even if the
       current source file is in a subdirectory!
 
-Sphinx Directive for Including Notebooks
-----------------------------------------
-
-Notebooks can also be included directly in an RST file
-using the ``nbinclude`` directive.
-Below, we include :ref:`subdir2/included.ipynb`.
-
-.. nbinclude:: subdir2/included.ipynb
-
-The notebook was included with:
-
-.. code-block:: rst
-
-    .. nbinclude:: subdir2/included.ipynb
-
-.. note::
-
-    * "``../``" is allowed here. In fact, you can include notebooks that are
-      outside of the documentation directory!
-
 Sphinx Directives for Jupyter Notebook Cells
 --------------------------------------------
 
@@ -93,3 +73,23 @@ This was created with
         :execution-count: 42
 
         42
+
+Sphinx Directive for Including Notebooks
+----------------------------------------
+
+Notebooks can be included directly in an RST file
+using the ``nbinclude`` directive.
+Below, we include :ref:`subdir2/included.ipynb`.
+
+.. nbinclude:: subdir2/included.ipynb
+
+The notebook was included with:
+
+.. code-block:: rst
+
+    .. nbinclude:: subdir2/included.ipynb
+
+.. note::
+
+    * "``../``" is allowed here. In fact, you can include notebooks that are
+      outside of the documentation directory!
--- a/doc/subdir2/included.ipynb
+++ b/doc/subdir2/included.ipynb
@@ -1,0 +1,59 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "nbsphinx": "hidden"
+   },
+   "source": [
+    "This notebook is part of the `nbsphinx` documentation: http://nbsphinx.readthedocs.org/."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### An Included Notebook\n",
+    "\n",
+    "This notebook is included with the `nbinclude` directive.\n",
+    "\n",
+    "The cell below is stored unexecuted, but will be executed when included, just like notebooks included in a `:toctree:`. Notebooks with at least one executed cell will not be executed when included."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "6 * 7"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.5.1+"
+  },
+  "nbsphinx": {
+   "orphan": true
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
--- a/doc/subdir2/included.ipynb
+++ b/doc/subdir2/included.ipynb
@@ -17,7 +17,11 @@
     "\n",
     "This notebook is included with the `nbinclude` directive.\n",
     "\n",
-    "The cell below is stored unexecuted, but will be executed when included, just like notebooks included in a `:toctree:`. Notebooks with at least one executed cell will not be executed when included."
+    "The cell below is stored unexecuted,\n",
+    "but will be executed when included,\n",
+    "just like notebooks included in a ``toctree``.\n",
+    "Notebooks with at least one executed cell will not be executed when included\n",
+    "(see [A Pre-Executed Notebook](../pre-executed.ipynb) for more details)."
    ]
   },
   {
@@ -29,6 +33,37 @@
    "outputs": [],
    "source": [
     "6 * 7"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Local images work like notebooks in `toctree`s:\n",
+    "![Jupyter notebook icon](../images/notebook_icon.png)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "from IPython.display import Image\n",
+    "Image(filename='../images/notebook_icon.png')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A link to a notebook in the parent directory: [link](../markdown-cells.ipynb).\n",
+    "\n",
+    "A link to a specific section in that notebook: [link](../markdown-cells.ipynb#Links-to-Other-Notebooks).\n",
+    "\n",
+    "A link to a local file: [link](../images/notebook_icon.png)."
    ]
   }
  ],
--- a/doc/subdir2/included.ipynb
+++ b/doc/subdir2/included.ipynb
@@ -49,9 +49,6 @@
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
    "version": "3.5.1+"
-  },
-  "nbsphinx": {
-   "orphan": true
   }
  },
  "nbformat": 4,
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -575,6 +575,13 @@ class NbInclude(rst.Directive):
             raise self.severe(
                 u'Problems with "nbinclude" directive:\n%s.' % error)
 
+        # Add the docname to the `files_to_rebuild` dictionary,
+        # which maps from files to files that include that file in a toctree.
+        # This suppresses the warning that is usually raised when a file
+        # is not in any toctree.
+        name, _ = os.path.splitext(os.path.relpath(path, settings.env.srcdir))
+        settings.env.files_to_rebuild.setdefault(name, set())
+
         # Use the NotebookParser to get doctree nodes
         nbparser = NotebookParser()
         node = docutils.utils.new_document(path, settings)
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -27,7 +27,9 @@ __version__ = '0.2.5'
 
 import copy
 import docutils
+from docutils import io
 from docutils.parsers import rst
+from docutils.parsers.rst import directives
 import jinja2
 import nbconvert
 import nbformat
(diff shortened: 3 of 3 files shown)


mgeier:
Thanks for this PR! I hope it isn't actually your final PR. I'm looking forward to receiving more ...

I'll have a closer look later, but first I want to comment on your suggested changes regarding `index.rst` and `index.ipynb`:

The reason why I initially added `index.ipynb` was to show a somewhat accidental and quite obscure feature.
It is possible to have source files with the same name but different extensions. Sphinx just uses the one whose extension comes first in `source_suffix` (which is typically `.rst`).
Since `nbsphinx` rewrites Markdown links to `*.ipynb` files to ``:doc:`basename` `` links, this can be used for a little trick that is shown in [doc/orphan.ipynb](http://nbsphinx.readthedocs.org/en/0.2.5/orphan.html) with the "Back to the main page" link. This link leads to `index.ipynb` when viewed in the Jupyter Notebook and on [nbviewer](http://nbviewer.jupyter.org/github/spatialaudio/nbsphinx/blob/master/doc/orphan.ipynb). However, after it's processed by Sphinx, it magically links to the Sphinx index page generated from `index.rst`.

However, all this became obsolete when I added the `"nbsphinx-toctree"` metadata which allows to use the same `index.ipynb` for everything. Also, in the meantime it became possible to use Markdown links to `*.rst` files which get turned into links to the generated Sphinx pages automatically.

Now I had the choice to use either `index.ipynb` or `index.rst` for the `nbsphinx` docs. I decided to use the more traditional `index.rst` in order to not frighten away users who are used to Sphinx.
Also, the `toctree` metadata is somewhat obscure and probably harder to use, so I think `index.rst` is easier to understand for newcomers, even if they don't know Sphinx yet. And I guess all Sphinx tutorials mention `index.rst`.

Anyway, I think that both are equally valid ways to use `nbsphinx`, but I think it's very confusing to use both at the same time.
I think the best solution here is not to link the two with your new `notebook` directive but instead remove `index.ipynb` completely. I added #34 as a reminder.

Does that sound reasonable?

tbekolay:
Yes, definitely reasonable! I agree that sticking with `index.rst` and removing `index.ipynb` is the best way to go.

Do you think it's necessary to show an example of the notebook directive, or is adding a notebook as documentation sufficient? I think the most compelling example would be including a notebook that is not in the `doc` folder (as that's why we're using it in our project), but I don't want to litter the main directory with a notebook if it doesn't need to be there. What do you think?

mgeier:
I think an example in the docs would be good, both as documentation and as test if it even works.

I would add this to the end of [doc/rst.rst](http://nbsphinx.readthedocs.org/en/0.2.5/rst.html) and include a short notebook there.
I'm not sure how links from the notebook to other places should be handled and if the links have to be rewritten if it's included from a different directory.
Or is this already handled?

Either way, I think it would be useful to put the example notebook in a subdirectory, e.g. `doc/subdir2/`.

nbsphinx.py
```diff
@@ -848,6 +889,7 @@ def setup(app):
 
     app.add_directive('nbinput', NbInput)
     app.add_directive('nboutput', NbOutput)
+    app.add_directive('notebook', Notebook)
```
mgeier: I'm not sure if it's good to call this `notebook`, since that's already used by `notebook_sphinxext.py`.

What about calling it `nbinclude` to show the similarity with the built-in `include` directive?

tbekolay: We previously used a modified version of `notebook_sphinxext.py`, hence why I went with `notebook` :) But `nbinclude` makes a lot of sense and is a natural extension of the `include` directive, and also the `NbInput` and `NbOutput` directives defined in nbsphinx, so I'll make that change!

tbekolay:
I added an example in `doc/rst.rst` in d1da027c596bd301c. And it was a good thing! When I did this initially, the included document was inserted at the bottom of the document, even though the `nbinclude` was in the middle. So, I reworked the first commit (731e177b44257bb1013ba) to use `insert_lines` instead of parsing directly into the document. This necessitated a bit of a refactor of `NotebookParser`, is that change OK?

mgeier:
Thanks! I'll have a closer look at this in the next few days ...
Am 05.04.2016 18:07 schrieb "Trevor Bekolay" <notifications@github.com>:

mgeier:
OK, I had a look ...

I don't think it's necessary to make the detour over the reST string. Also, if #36 comes through, there won't be a reST string anymore.

Do you still have the hash of your original commit?
I'd be interested to see what was the problem there.

mgeier:
I found it: 212ea7354af5cccbe62f329d65aaa995158c132a

tbekolay:
Yep, that's it :) The difference is that I was parsing directly into the document (`nbparser.parse(rawtext, self.state.document`) as opposed to adding input lines to `self.state_machine`. I don't know enough about Spinx / docutils internals to know why parsing into the document placed the tree at the end of the document, but adding input lines to `self.state_machine` is what the `include` directive does, so I think it's likely the right thing to do.

mgeier:
You shouldn't parse into `self.state.document`, you should rather return the list of directives which should be inserted by `NbInclude` (I wonder why `include` doesn't do that). To do that, you'll have to parse into a temporary document and then return the content of that temporary document.

I'm not sure if that's the best way, but it should work somewhat like this:

```python
node = docutils.utils.new_document('somename', self.state.document.settings)
nbparser.parse(rawtext, node)
return node.children
```

tbekolay:
Yes, I also tried that before settling on this current solution.

The issue with making a new document, parsing into it, and returning the parsed doctree is that other parts of the document (specifically the Sphinx sidebar) gets included in `node`, and therefore ends up nested in the body of the doctree where it should be outside of it. For some reason, doing this also prevents the sidebar from showing up where it should show up (so, you end up with one misplaced sidebar rather than two, which I would have expected).

That said, when I tried it returned `[node]`; maybe returning `[node.children]` would fix the issue?

mgeier:
I didn't test this extensively, but the sidebar looked OK ...

`node.children` is already a sequence, so you shouldn't put it into another list.

If `node.children` turns out to be wrong, there has to be something similar that does the right thing.

tbekolay:
I'll try it out!

tbekolay:
`node.children` does indeed do the right thing! There's no sidebar issue with it. The updated PR to no longer change `NotebookParser`.

One issue that I have now (that isn't related to the change to `node.children`; this happens with the previous version of the PR) is that I get a weird "**orphan:**" label rendered in `rst.html`. See [here](http://www.bekolay.org/nbsphinx/rst.html).

I added the orphan metadata to the `included` notebook so that a warning isn't generated in the Sphinx build process, but for some reason when parsing, a table is added to the doctree rendering the orphan label. However, this doesn't appear in the rendered versions of those notebooks alone (see [included](http://www.bekolay.org/nbsphinx/subdir2/included.html) and [orphan](http://www.bekolay.org/nbsphinx/orphan.html). Any idea what's going on there?

mgeier:
This should be the explanation: http://www.sphinx-doc.org/en/stable/markup/misc.html#file-wide-metadata

I don't know if that's the proper solution, but you can try this:

```python
if isinstance(node.children[0], docutils.nodes.field_list):
    return node.children[1:]
return node.children
```

... or something along those lines ...

This theoretically also removes `:tocdepth:` and `:nocomments:`, but since there is no way to set those in a notebook, this shouldn't be a problem. Also, I'm not sure if those even make sense in an included notebook.

doc/rst.rst
```diff
@@ -36,6 +36,26 @@ These links were created with:
     * "``../``" is not allowed, you have to specify the full path even if the
       current source file is in a subdirectory!
 
+Sphinx Directive for Including Notebooks
```
mgeier: I would prefer if you could move this section after the section about `nbinput` and `nboutput`.

doc/subdir2/included.ipynb
```diff
@@ -0,0 +1,59 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### An Included Notebook\n",
+    "\n",
+    "This notebook is included with the `nbinclude` directive.\n",
+    "\n",
+    "The cell below is stored unexecuted, but will be executed when included, just like notebooks included in a `:toctree:`. Notebooks with at least one executed cell will not be executed when included."
```
mgeier: `toctree` doesn't have colons.

BTW, I try to use line breaks in Markdown cells because I think this makes the text easier to edit.

mgeier: Do you think it would be good to add a link to `pre-executed.ipynb` here?
Notebook execution is supposed to be explained there ...

doc/subdir2/included.ipynb
```diff
@@ -0,0 +1,59 @@
...
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.5.1+"
+  },
+  "nbsphinx": {
+   "orphan": true
```
mgeier: Regardless of the problem in https://github.com/spatialaudio/nbsphinx/pull/33#issuecomment-206923109, I think it would be nice if including the notebook would avoid the orphan warning without forcing people to add notebook metadata.

nbsphinx.py
```diff
@@ -27,7 +27,9 @@
 
 import copy
 import docutils
+from docutils import io
 from docutils.parsers import rst
+from docutils.parsers.rst import directives
```
mgeier: Since `rst` is already available, you can drop this import and instead use `rst.directives` in the code.

nbsphinx.py
```diff
@@ -541,6 +543,50 @@ def run(self):
         return [container]
 
 
+class NbInclude(rst.Directive):
+    """Include a notebook in a reST document.
+
+    This directive uses :class:`NotebookParser`, which means that evaluated
+    notebooks will be included without modification, while notebooks with
+    no output cells will be evaluated before being included.
```
mgeier: You should move this explanation to the docstring of `NotebookParser`.

nbsphinx.py
```diff
@@ -541,6 +543,50 @@ def run(self):
...
+
+    """
+    required_arguments = 1
+    optional_arguments = 0
+    final_argument_whitespace = True
+    option_spec = {}
```
mgeier: I think it would be good to add this here, just to be explicit:

```python
has_content = False
```

nbsphinx.py
```diff
@@ -541,6 +543,50 @@ def run(self):
...
+
+    def run(self):
```
mgeier: What about this:

```python
settings = self.state.document.settings
```

This would slim down the following code.

nbsphinx.py
```diff
@@ -541,6 +543,50 @@ def run(self):
...
+        if not self.state.document.settings.file_insertion_enabled:
+            raise self.warning('"%s" directive disabled.' % self.name)
```
mgeier: I'd hard-code the directive name here.

nbsphinx.py
```diff
@@ -541,6 +543,50 @@ def run(self):
...
+
+        # Read the notebook to a string
+        path = directives.path(self.arguments[0])
+        rst_file = self.state_machine.document.attributes['source']
+        rst_dir = os.path.abspath(os.path.dirname(rst_file))
```
mgeier: I'm not sure if that's cleaner, but those 2 lines could be written as:

```python
rst_dir = os.path.dirname(settings.env.doc2path(settings.env.docname))
```

nbsphinx.py
```diff
@@ -541,6 +543,50 @@ def run(self):
...
+                                        encoding='utf-8',
+                                        error_handler=e_handler)
+        except IOError as error:
+            raise self.severe(u'Problems with "%s" directive path:\n%s.' %
+                              (self.name, error))
+        try:
+            rawtext = include_file.read()
+        except UnicodeError as error:
+            raise self.severe(u'Problem with "%s" directive:\n%s' %
+                              (self.name, error))
```
mgeier: I'd prefer to use the way how it's done in `nbformat` here.

```python
import io

try:
    with io.open(path, encoding='utf-8') as f:
        rawtext = f.read()
except IOError ...
    ...
except UnicodeError ...
    ...
```

nbsphinx.py
```diff
@@ -541,6 +543,50 @@ def run(self):
...
+
+        # Use the NotebookParser to convert to reST and evaluate if needed
```
mgeier: Conversion to reST is an implementation detail and will likely go away with #36.

mgeier:
Can you please add links and images to the example notebook, similar to `subdir/another.ipynb`?

Somehow you'll have to manipulate the paths to get this working correctly.

tbekolay:
Sorry for the delay! I think I've incorporated all the feedback now, and where it made sense, I made FIXUP commits so that you can see what I've changed compared to what was in this PR before. Before merging, I can fixup those commits in an interactive rebase to keep the history clean.

I was able to remove the `toctree` warning in commit b1255d18fbf877c8e2b. Now, the included notebook doesn't have the `orphan` metadata, though of course it could have it if desired. Your fix to not render the field list worked in my testing.

The only thing that I think is not really resolved thus far is adding links / images to the included notebook. The issue is that relative paths depend on the directory in which the notebook is executed, and there are two sets of relative paths to worry about: those written by the notebook creator, and those generated by `nbconvert`. There are four options that I can see, and none of them is a clear winner.

1. Execute the notebook in the directory that it exists in. I did this in 387df6f7991bc3d81e1 but it does not render correctly because relative paths are wrong, including the relative path generated by the `IPython.display.Image`, even though the actual image resources are saved to the correct place on disk.

2. Execute the notebook in the directory that it exists in, but rewrite any relative paths in the notebook to point to their relative path from the new directory. It seems like this should be possible through augmenting `ProcessLocalLinks`? Or perhaps some different transformation of the doctree is necessary? The main downside here is that it might be difficult to find all cases in which relative paths are used and rewrite them appropriately.

3. Execute the notebook in the directory that includes it. This is what happens by default (i.e., without 387df6f7991bc3d81e1). This saves resources and references them properly; however, it means that opening up the notebook and executing it with Jupyter behaves differently than when it is executed when included in a reST file, essentially meaning that the notebook doesn't really work _unless_ it's nbincluded. If you have relative paths in the notebook, you write them as though the notebook was in the directory that it is included in, not the directory that it is actually in. If the notebook doesn't use relative paths (which is probably true of a lot of notebooks; it's true of the notebook in our project that is using `nbsphinx`) then this works properly. But for documents that link to other documents (which is likely the case for notebooks in a documentation tree) this is likely not true.

4. Raise an error if the notebook is in a different directory than the RST file that includes it, and uses relative paths. These can likely be detected in non-code cells; if the relative paths are in code cells, then running those cells is likely to raise a `CellExecutionError`, which we can catch and reraise.

Whichever option we go with will need to be documented, in any case!

mgeier:
Thanks for the changes!

I think option 2 is the right choice.

I'll have a closer look tomorrow ...

mgeier:
OK, I had a closer look ...
In general, it seems to work nicely, but the problems are in the details.

Most code assumes that the files are in the source directory, `ProcessLocalLinks` even explicitly raises a warning (and ignores the link) if a local link points outside of the source directory.
This sure could be solved somehow, but it would mean creating a whole parallel implementation of this stuff "around" Sphinx.

Links to notebooks outside of the source directory would also be quite hard to get working.

The implementation would definitely become much more complicated and therefore harder to maintain in the future.

Now let me take a step back and ask you this:

Do you really need this?

I see a few alternatives for your case:

1. move `examples/` to `doc/examples/`. You said you don't want that, and I understand your reasons, but it would be a *really* simple solution
1. create a symbolic link `doc/examples` that points to `examples/`. I'm not sure if that works on Windows, but at least theoretically, symbolic links should be supported since Windows Vista.
1. if you don't want to move the examples into the source directory, move the source directory around the examples. Moving `conf.py` to the main directory should do it. All other files can stay at their old places. You'd just have to specify the correct `master_doc`.

I think I'd prefer alternative 2 in you case. All three of them would be hugely simpler than implementing (and probably even using) the `nbinclude` directive.

What do you think?

jgosmann:
Unfortunately, option 2 is not a complete solution as this causes problems with images in output cells as detailed in #49. Also, this is problematic on Windows. Symbolic links have been supported for a while, but up to Windows 10 required elevated privileges to be created.

Option 1 is in my case currently not possible (the notebooks need to be part of the installed Python package).

mgeier:
@jgosmann Thanks for your comment. I know that each of the alternatives has its own weaknesses. I'm open to suggestions for solutions to the problems of this PR, probably we can solve them ...

Another thing you could try is to use your `setup.py` to copy around the notebooks to wherever you need them.
You can probably also use the [package_dir](https://docs.python.org/3.6/distutils/setupscript.html#listing-whole-packages) option to manipulate where your notebooks are installed (since you say that they are supposed to be installed with the package). You might also be able to do some path manipulations with [package_data](http://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files)?

You could also try https://github.com/ngoldbaum/RunNotebook, which I think lets you include notebooks from wherever, but AFAICT its integration with Sphinx is weaker (which may actually be what you want) and it has no LaTeX/PDF support.

mgeier:
I've just stumbled upon the PyPI package `nbsphinx-link`, see https://github.com/vidartf/nbsphinx-include.

tbekolay:
Hadn't seen nbsphinx-include, thanks @mgeier, will check it out :)

Seanny123:
@tbekolay we used `nbsphinx-link` in `nengo_extras`, so should we close this PR?

tbekolay:
Yes, `nbsphinx-link` is a better solution than what's in this PR. Thanks for pointing us to it @mgeier!

Issue #1, opened 2015-12-07
Move CSS string from <body> to <head>


mgeier:
There was a related question on the sphinx-users mailing list, but no solution:

https://groups.google.com/forum/?nomobile=true#!searchin/sphinx-users/css/sphinx-users/iG0fB9-JdNM/Le-rz_zbRHUJ

Also, the CSS string should only be included in the pages which actually need it.

mgeier:
This can probably done with the `extrahead` block in `layout.html`, but it would be nice to be able to do that programmatically (without adding a separate template file).

mgeier:
Since cdfd4f789b63c2817ba627eb42d26ac658f288ec, the CSS is only included for HTML files generated from notebooks (except if Sphinx version < 1.3.6 is used).

The CSS could also be added to `context['metatags']` instead of `context['body']`, but then it would end up very close to the top of the HTML file, which isn't too great either.
And I guess it is better to define the `nbsphinx` CSS *after* other CSS files are included ...

tbekolay:
I just rebased the nbinclude stuff to master and cdfd4f7 kind of breaks it, in that the CSS doesn't get loaded so the included notebook looks bad. The `rst.rst` file in general doesn't work, as you can't manually use the `nbinput` or `nboutput` directives anymore without the CSS.

mgeier:
Oh, sorry, I completely missed that!

I created a different solution in #37, can you please try if that works for you?

I'm not sure if that's a proper way to do that (see my doubts in #37), probably you know some improvements?

PR #37, opened 2016-04-13
Fix inclusion of CSS

mgeier:
This hopefully fixes the bug described in https://github.com/spatialaudio/nbsphinx/issues/1#issuecomment-209125641.

I'm not sure if it's allowed to add new fields to the `document` and I don't know if `True` is a meaningful value for an attribute of `document` ...

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -513,6 +513,7 @@ class NbInput(rst.Directive):
         _set_empty_lines(node, self.options)
         node.attributes['latex_prompt'] = latex_prompt
         container += node
+        self.state.document['nbsphinx_include_css'] = True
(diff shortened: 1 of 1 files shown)


tbekolay:
I'll test this out today!

mgeier:
I just had a closer look, and boolean values are actually used in the `document` (e.g. `linenos="False"`). The values apparently are somehow stored with type information, there is a difference between `"True"` and `True`. Therefore, I think my use of `True` is fine.

As for adding arbitrary attributes, I still don't know if that's officially allowed, but since it works fine, and collisions are avoided by prefixing `nbsphinx_`, I think that's fine, too.

PR #39, opened 2016-04-14
Add don't execute explicit flag

jdfreder:
I have a lot of notebooks that don't make sense to include outputs and yet I want them built into RTD.  This lets me do that in a less kludgey way.

cc @willingc

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,8 +6,7 @@ Jupyter Notebook Tools for Sphinx
 Custom Sphinx directives are used to show `Jupyter Notebook`_ code cells (and of
 course their results) in both HTML and LaTeX output.
 Un-evaluated notebooks -- i.e. notebooks without stored output cells -- will be
-automatically executed during the Sphinx build process unless disabled via
-config.
+automatically executed during the Sphinx build process.
 
 .. _Sphinx: http://sphinx-doc.org/
 .. _Jupyter Notebook: http://jupyter.org/
--- a/README.rst
+++ b/README.rst
@@ -6,7 +6,8 @@ Jupyter Notebook Tools for Sphinx
 Custom Sphinx directives are used to show `Jupyter Notebook`_ code cells (and of
 course their results) in both HTML and LaTeX output.
 Un-evaluated notebooks -- i.e. notebooks without stored output cells -- will be
-automatically executed during the Sphinx build process.
+automatically executed during the Sphinx build process unless disabled via
+config.
 
 .. _Sphinx: http://sphinx-doc.org/
 .. _Jupyter Notebook: http://jupyter.org/
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -361,10 +361,10 @@ class Exporter(nbconvert.RSTExporter):
 
     """
 
-    def __init__(self, allow_errors=False, dont_execute=False, timeout=30, codecell_lexer='none'):
+    def __init__(self, allow_errors=False, execute='auto', timeout=30, codecell_lexer='none'):
         """Initialize the Exporter."""
         self._allow_errors = allow_errors
-        self._dont_execute = dont_execute
+        self._execute = execute
         self._timeout = timeout
         self._codecell_lexer = codecell_lexer
         loader = jinja2.DictLoader({'nbsphinx-rst.tpl': RST_TEMPLATE})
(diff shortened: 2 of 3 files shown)


mgeier:
Thanks for this PR, I was already waiting for this to come up!

I don't really like the negated option name, why not just call it `nbsphinx_execute` and `execute`?

README.rst
```diff
@@ -6,7 +6,8 @@ Jupyter Notebook Tools for Sphinx
 Custom Sphinx directives are used to show `Jupyter Notebook`_ code cells (and of
 course their results) in both HTML and LaTeX output.
 Un-evaluated notebooks -- i.e. notebooks without stored output cells -- will be
-automatically executed during the Sphinx build process.
+automatically executed during the Sphinx build process unless disabled via
+config.
```
mgeier: Is it really necessary to mention this here?
I'd like to keep this simple.

nbsphinx.py
```diff
@@ -925,6 +928,7 @@ def setup(app):
     _add_notebook_parser(app)
 
     app.add_config_value('nbsphinx_allow_errors', False, rebuild='')
+    app.add_config_value('nbsphinx_dont_execute', False, rebuild='')
```
mgeier: I think this should be `rebuild='env'`, right?

mgeier:
I guess I know why you called it `dont_execute` ... if you call it `execute` this might be interpreted as "force execution even if output cells are already there" ...

Should `dont_execute=False` (or `execute=True`) force execution?

Probably there should be a string for differentiating the use cases?

* never execute
* only execute if no outputs are present (a.k.a. "auto"?)
* always execute

tbekolay:
This PR looks great, making this configurable makes a lot of sense! I am +1 on a string (enumeration) of `execute='false'` or `'true'` or `'auto'` where the default is `'auto'`.

nbsphinx.py
jdfreder: I have no idea, I was just mimicking the code for `nbsphinx_allow_errors`

jdfreder:
Comments addressed

mgeier:
Thanks for the changes! I'm still not quite satisfied, though.

I don't really like @tbekolay s suggestion with `'true'` and `'false'`. Why should those be strings instead of `True`/`False` (that's probably a rhetorical question)?

I don't think I've seen that before, but I've seen `None` vs. `'none'` and found it very strange ...

What about `execute='always'` and `execute='never'`?
What about `execute='yes'` and `execute='no'`?
Any other suggestions?

tbekolay:
You're right, `'true'` and `'false'` is probably too close to `True`/`False`. I like `'always'` / `'never'` as it seems the most explicit and least similar to true/false... nothing else comes to mind!

nbsphinx.py
```diff
@@ -387,8 +388,10 @@ def from_notebook_node(self, nb, resources=None, **kw):
         nbsphinx_metadata = nb.metadata.get('nbsphinx', {})
 
         # Execute notebook only if there are code cells and no outputs:
-        if (any(c.source for c in nb.cells if c.cell_type == 'code') and
-                not any(c.outputs for c in nb.cells if 'outputs' in c)):
+        execute_option = nbsphinx_metadata.get('execute', self._execute)
+        has_unexecuted = (any(c.source for c in nb.cells if c.cell_type == 'code') and
+            not any(c.outputs for c in nb.cells if 'outputs' in c))
+        if execute_option == 'true' or (execute_option == 'auto' and has_unexecuted):
```
mgeier: Can you please check for all possible values of `execute_option` and raise a `NotebookError` if it's invalid (similar to the handling of "orphan" below)?

nbsphinx.py
```diff
@@ -387,8 +388,10 @@ def from_notebook_node(self, nb, resources=None, **kw):
         nbsphinx_metadata = nb.metadata.get('nbsphinx', {})
 
         # Execute notebook only if there are code cells and no outputs:
-        if (any(c.source for c in nb.cells if c.cell_type == 'code') and
-                not any(c.outputs for c in nb.cells if 'outputs' in c)):
+        execute_option = nbsphinx_metadata.get('execute', self._execute)
+        has_unexecuted = (any(c.source for c in nb.cells if c.cell_type == 'code') and
```
mgeier: `has_unexecuted` (in the sense of "has at least one un-executed cell") isn't quite the right name, it's rather "has no executed cells" or "no outputs"

nbsphinx.py
```diff
@@ -387,8 +388,10 @@ def from_notebook_node(self, nb, resources=None, **kw):
         nbsphinx_metadata = nb.metadata.get('nbsphinx', {})
 
         # Execute notebook only if there are code cells and no outputs:
-        if (any(c.source for c in nb.cells if c.cell_type == 'code') and
-                not any(c.outputs for c in nb.cells if 'outputs' in c)):
+        execute_option = nbsphinx_metadata.get('execute', self._execute)
```
mgeier: I think for symmetry with the other options this should just be called `execute`.

nbsphinx.py
jdfreder: I like no_outputs

nbsphinx.py
jdfreder: Definitely

nbsphinx.py
jdfreder: Sure

nbsphinx.py
mgeier: great

mgeier:
@jdfreder Are you planning to do the last few discussed changes?
Or do you want me to do them?

dchudz:
I'm looking forward to this being merged.

willingc:
Pinging @jdfreder. Let me know if you want to pair on this or let @mgeier finish it off ;-)

jdfreder:
Sorry, I've been preoccupied- yeah @mgeier you can finish it off, or @willingc .  I can give either of you r/w access to my repo if you'd like.

mgeier:
@jdfreder I've squashed your commits into 971a76c099718f96ca5531a039e854bc13aa16a3 and added a follow-up commit (e758d6772de786c18c7dbc784806792f1de927c7) with the changes discussed here.

Thanks for your contribution, I'm looking forward to more!
Thanks @tbekolay and @willingc for helping!

dchudz:
thanks for this, everyone! This change will let me use nbsphinx in my team's docs. Any idea when it'll be on PyPI?

dchudz:
The versioning is a bit confusing. Master branch has `__version__ = '0.2.7'`, and `0.2.7` is what's on PyPI... so I thought that would mean this is on PyPI (but it's not).

mgeier:
@dchudz I've just made a new release on PyPI.

The `__version__` is only valid and meaningful for releases, once you use the Git version, you should use Git's version information. You can see on the [release page](https://github.com/spatialaudio/nbsphinx/releases) how far ahead of each release the `master` branch is.

I was simply too lazy to make another commit after each release setting the version to `x.y.z.dev` or something. Do you think it would be worth it?

dchudz:
Oh, I see. Sorry about that! Thanks so much.

David

PR #42, opened 2016-04-16
add scrollbar to overly long output lines

### LINKED ISSUE #40: wrapping long output
How can long output be wrapped correctly?

In some notebooks I print long lines, which extend over the screen (see rendered html here: http://wradlib.org/wradlib-docs/latest/simple_plot.html). How can it be done, that the lines are wrapped at the "end" of the code-boxes?

JFTR, the notebook is rendered with `runipy` and then sphinxed `nbsphinx`. Great tool, btw.

kmuehlbauer:
This uses the CSS `white-space` property (http://www.w3schools.com/cssref/pr_text_white-space.asp).

Set to `pre-wrap` it has the following outcome:

- Whitespace is preserved by the browser. 
- Text will wrap when necessary, and on line breaks 

This works for my use-cases. I did not encounter any problems so far. 

Closes #40.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -236,6 +236,14 @@ div.nboutput div[class^=highlight] pre {
     box-shadow: none;
 }
 
+/* avoid overly long output lines 
+   Whitespace is preserved by the browser. 
+   Text will wrap when necessary, and on line breaks. 
+*/
+div.nboutput div[class^=highlight] pre {
+    white-space: pre-wrap;
+}
+
 /* avoid gaps between output lines */
 div.nboutput div[class^=highlight] pre {
     line-height: normal;
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -236,14 +236,6 @@ div.nboutput div[class^=highlight] pre {
     box-shadow: none;
 }
 
-/* avoid overly long output lines 
-   Whitespace is preserved by the browser. 
-   Text will wrap when necessary, and on line breaks. 
-*/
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR!

I don't like how wrapped pre-formatted text looks like, though. I think I'd prefer scrollbars.

What about using `overflow: auto;` instead of the wrapping stuff?
You could add this to the "input/output area" section.

kmuehlbauer:
Yeah, I think this is more appropriate for the most cases.

But if you have very(!) long output lines, you have to scroll and scroll. Don't know how to tackle this (without using `white-space`). One would have to adapt the print statement, to wrap the lines. Suggestions?

mgeier:
Thanks!

If you are talking about Python, you could try the [textwrap](https://docs.python.org/3/library/textwrap.html) module.

jaladh-singhal:
@mgeier Is this enabled by default, or is it required to change some setting because it doesn't work in [our docs](https://tardis-sn.github.io/tardis/quickstart/quickstart.html#Running-the-simulation-(long-output)) ?

jaladh-singhal:
@mgeier Thanks for your reply. Now that you have mentioned, I realized my mistake, - I'm actually looking for vertical scrollbar in long outputs (with so many lines). Sorry I must have confused between the words. 

I understand this PR doesn't implement that but is there any way to achieve that?

mgeier:
#417, #372

PR #56, opened 2016-07-05
Add dependency on nbconvert "execute" extra requirement

mgeier:
This in turn depends on `jupyter-client`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     install_requires=[
         'docutils',
         'jinja2',
-        'nbconvert',
+        'nbconvert[execute]',
         'nbformat',
         'sphinx',
     ],


mgeier:
I'm not sure if that's a good idea because if notebooks are not executed the Jupyter Notebook doesn't have to be installed at all.

A hard dependency on `jupyter-client` would also add a dependency on PyZMQ (and probably more stuff).

Since many people seem to be using `nbsphinx` without auto-executing their notebooks, I'm reluctant to add that dependency ...

mgeier:
I've mentioned this in the docs: http://nbsphinx.readthedocs.io/en/latest/usage.html#Jupyter-Kernel

I think it's better not to force this dependency ... I'm closing this issue.

PR #75, opened 2016-12-05
Custom mime types

SylvainCorlay:
This enables outputs of type `vnd.jupyter.widget-state+json` and `vnd.jupyter.widget-view+json` in output areas.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -48,6 +48,8 @@ _ipynbversion = 4
 # See nbconvert/exporters/html.py:
 DISPLAY_DATA_PRIORITY_HTML = (
     'application/javascript',
+    'application/vnd.jupyter.widget-view+json',
+    'application/vnd.jupyter.widget-state+json',
     'text/html',
     'text/markdown',
     'image/svg+xml',
@@ -155,6 +157,11 @@ RST_TEMPLATE = """
         var element = document.getElementById('{{ div_id }}');
 {{ output.data['application/javascript'] | indent | indent }}
         </script>
+{%- elif datatype.startswith('application') and datatype.endswith('+json') %}
+
+    .. raw:: html
+
+        <script type="{{ datatype }}">{{ output.data[datatype] }}</script>
 {%- elif datatype == 'ansi' %}
 
     .. rst-class:: highlight


mgeier:
Thanks!

SylvainCorlay:
Would you be ok with me transforming the widget state from the notebook metadata to a similar script tag in the html?

This would enable things like embedding live widgets from notebooks into sphinx documentation...

SylvainCorlay:
I just made a sphinx extension to do this manually: https://github.com/jupyter/jupyter-sphinx bu

sphinx-screencast

But it would be nice if nbsphinx did this too out of the box.

mgeier:
If the information is already in the notebook metadata, I don't see see anything that speaks against it.

Can you please point me to an example notebook which contains such metadata?

How would nbconvert and nbviewer handle such a case?

SylvainCorlay:
@mgeier these mime types will be new in ipywidgets 6.0.

I have specified the spec for the mime types in 

https://github.com/ipython/ipywidgets/tree/master/jupyter-widgets-schema

So at the moment, it is not handled by nbconvert and all.

My goal is to have it working in the ipywidgets documentation for the next ipywidgets release....

mgeier:
OK, sure, I'm looking forward to a PR!

It would still be helpful for me to see a concrete example, since I don't understand (and don't want to learn) JSON schema.
And I guess apart from the JSON data, you will also have to inject some JavaScript code into the output page, right?

SylvainCorlay:
Sure, expect a PR + example tonight or tomorrow.

The javascript code is done here: https://github.com/jupyter/jupyter-sphinx/blob/master/jupyter_sphinx/embed_widgets.py#L222

So if we enable the `jupyter-sphinx` extension it should work.

I am not sure of the behavior is the script tag is inserted twice, so I will make some tests before I propose adding this line to nbsphinx.

SylvainCorlay:
@mgeier I posted an example notebook in the following gist:

https://gist.github.com/SylvainCorlay/d6d8cd44e032523d07e102f5dc0bb1df

The crux of it is that the widget manager state is stored in the notebook metadata:

```json
"metadata": {
    ...
    "widgets": {
        "application/vnd.jupyter.widget-state+json": {
        ...
        }
    }
}
```

This would only be replaced with

````html
<script type="application/vnd.jupyter.widget-state+json">
{
    ...
}
</script>
```

nbsphinx.py
```diff
@@ -155,6 +157,11 @@
         var element = document.getElementById('{{ div_id }}');
 {{ output.data['application/javascript'] | indent | indent }}
         </script>
+{%- elif datatype.startswith('application') and datatype.endswith('+json') %}
+
+    .. raw:: html
+
+        <script type="{{ datatype }}">{{ output.data[datatype] }}</script>
```
mgeier: In https://github.com/jupyter/nbconvert/pull/482 @michaelpacer uses the same `element` trick as in `application/javascript` blocks. Should this also be added here?

PR #76, opened 2016-12-07
Adding a footer template handler to add the widget state script tag

SylvainCorlay:
With this change , we can render interactive widgets in documentation generated from nbsphinx:

Thanks a lot to @takluyver for his help in person here and @michaelpacer for the work in progress in the html convert in nbconvert.

- Adding a json.dumps step in the serialization of `+json` types
- Adding a footer script tag for the widget state if present.

nbsphinx-screencast

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -162,7 +162,7 @@ RST_TEMPLATE = """
 
     .. raw:: html
 
-        <script type="{{ datatype }}">{{ output.data[datatype] }}</script>
+        <script type="{{ datatype }}">{{ output.data[datatype] | json_dumps }}</script>
 {%- elif datatype == 'ansi' %}
 
     .. rst-class:: highlight
@@ -240,6 +240,20 @@ RST_TEMPLATE = """
 {{ cell.source }}
 {% endif %}
 {% endblock rawcell %}
+
+
+{% block footer %}
+
+{% if 'application/vnd.jupyter.widget-state+json' in nb.metadata.widgets %}
+
+.. raw:: html
+
+    <script type="application/vnd.jupyter.widget-state+json">
+    {{ nb.metadata.widgets['application/vnd.jupyter.widget-state+json'] | json_dumps }}
+    </script>
+{% endif %}
+{{ super() }}
+{% endblock footer %}
 """
 
 
@@ -485,6 +499,7 @@ class Exporter(nbconvert.RSTExporter):
                 'get_empty_lines': _get_empty_lines,
                 'extract_toctree': _extract_toctree,
                 'get_output_type': _get_output_type,
+                'json_dumps': json.dumps,
             })
 
     @property


SylvainCorlay:
@mgeier  if we merge and make a quick release, this should enable the live widgets on https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html

mgeier:
Thanks, I'll look into this in the next few days.

In the meantime, can you please help me improve my understanding of widgets:

What is the relation between http://jupyter.org/widgets.html and `ipywidgets`?

The JSON data is embedded in the notebooks, why is the JS import not there, too?

This whole thing only works if the [jupyter-sphinx](https://github.com/jupyter/jupyter-sphinx) extension is loaded, right?

AFAICT, the extension loads the JS on each HTML output page, wouldn't it be better to load it just if a widget is actually used on a page?

Why is there is no Python command to load the JS into the notebook, like it's done e.g. with [bokeh.io.output_notebook()](http://bokeh.pydata.org/en/latest/docs/reference/io.html#bokeh.io.output_notebook)?

SylvainCorlay:
> What is the relation between http://jupyter.org/widgets.html and ipywidgets?

ipywidgets is the base framework upon which all the widgets libraries on http://jupyter.org/widgets.html are built.

Actually, this PR also enables (as a byproduct) the rendering of widgets from bqplot, pythreejs, ipyleaflet, nglview, and autodesk's molecular design toolkit from the notebook in sphinx documentation.

> The JSON data is embedded in the notebooks, why is the JS import not there, too?
> This whole thing only works if the jupyter-sphinx extension is loaded, right?

I consider the now well-specified schema for the two mime types (`application/vnd.jupyter.widget-state+json` and `application/vnd.jupyter,widget-view+json`) final.

The actual rendering indeed requires a script tag to be added to transform those script tags into a rendered widget.

`jupyter-sphinx`'s only purpose is to render the widgets so I already added the script tag in that case. It has a `0.0.x` version number and I will release `0.1.0` at the same time as `ipywidgets 6.0`  with the script tag for the final version of the javascript.

I imagine that we can decide to add the javascript as part of nbsphinx when ipywidgets 6.0 is released... I intended to do so in another PR. The current PR only adds static content which I thought was less likely to be an issue.

> AFAICT, the extension loads the JS on each HTML output page, wouldn't it be better to load it just if a widget is actually used on a page?

That would probably be premature optimization. We can simply load the js as part of the header for all pages of the documentation with `add_javascript` like https://github.com/jupyter/jupyter-sphinx/blob/master/jupyter_sphinx/embed_widgets.py#L222 which would be less bug prone than maintaining a logic to detect whether there was any widget on the page...

> Why is there is no Python command to load the JS into the notebook, like it's done e.g. with bokeh.io.output_notebook()?

That is because ipywidgets and other widgets libraries use the proper jupyter extension points (notebook extensions) to add JavaScript permanently. Bokeh does not.

SylvainCorlay:
@mgeier let me know if you would be ok to get this in with a minor version release.

I would love to be able to iterate with that feature enabled before getting ipywidgets 6 out!

mgeier:
Sure, I'll include it in the next minor release which I can make in a few days, if necessary.
I just wanted to give it a try first ...

When is the ipywidgets 6 deadline?

nbsphinx.py
```diff
@@ -487,6 +501,11 @@ def __init__(self, execute='auto', execute_arguments=[],
                 'get_output_type': _get_output_type,
             })
 
+    def default_filters(self):
```
mgeier: I've defined my other filters a few lines above.

Is there a reason to prefer `default_filters()`?

I would like to define all the filters the same way.

takluyver: Sorry, I missed that when suggesting that Sylvain add this method. In the case of nbsphinx, it probably doesn't matter which way you define filters.

In general, for custom exporters, it's slightly better to define this method. The `filters` variable you're passing in also allows users to specify extra filters in config, and if both the exporter and the user's config specify `filters`, one of them will clobber the other - I'm not even sure which way round. So the `default_filters()` method is an extension point specifically for authors of custom exporters.

SylvainCorlay:
> When is the ipywidgets 6 deadline?

We are sprinting on fixing the main issues that are remaining with the 6.0 milestone.

I think that we will make a RC release by the end of the year.

SylvainCorlay:
@mgeier merge?

nbsphinx.py
```diff
@@ -240,6 +240,20 @@
 {{ cell.source }}
 {% endif %}
 {% endblock rawcell %}
+
+{% block footer %}
+
+
```
mgeier: Please remove the superfluous empty lines here (if they are indeed superfluous).

nbsphinx.py
```diff
@@ -485,6 +499,7 @@ def __init__(self, execute='auto', execute_arguments=[],
                 'get_empty_lines': _get_empty_lines,
                 'extract_toctree': _extract_toctree,
                 'get_output_type': _get_output_type,
+                'json_dumps': json.dumps
```
mgeier: Please add a trailing comma.

SylvainCorlay:
Rebased and made changes as per your comments.

mgeier:
Thanks!

SylvainCorlay:
🎉 Awesome! 🎉 

Check out some live widgets on RTD: 

http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html

Not everything is perfect but we are getting there.

mgeier:
Cool!

PR #77, opened 2016-12-11
Force docutils <0.13.1

SylvainCorlay:
Docutils 0.13.1 which was just released breaks nbsphinx because of  https://github.com/sphinx-doc/sphinx/issues/3212

The issue seems to have been fixed in sphinx master since https://github.com/sphinx-doc/sphinx/pull/3217 but until a new version of sphinx is released, we need to force `docutils<0.13.1`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ setup(
     version=__version__,
     py_modules=['nbsphinx'],
     install_requires=[
-        'docutils',
+        'docutils<0.13.1',
         'jinja2',
         'nbconvert',
         'traitlets',


mgeier:
Would the long-term solution (after a few new Sphinx releases) be to force `sphinx>=1.5.1` and remove the limitation on `docutils`?

mgeier:
Note that there is a bug in https://github.com/sphinx-doc/sphinx/pull/3217, see https://github.com/sphinx-doc/sphinx/pull/3217#discussion_r91851109.

mgeier:
Would it work if we (after the release of Sphinx 1.5.1) use `sphinx!=1.5`?

There is another breaking change (see #78) which will hopefully be fixed in https://github.com/sphinx-doc/sphinx/pull/3224.

SylvainCorlay:
> Would the long-term solution (after a few new Sphinx releases) be to force sphinx>=1.5.1 and remove the limitation on docutils?

That is right, but in the mean time, the current setup breaks RTD builds with sphinx.

SylvainCorlay:
> Would it work if we (after the release of Sphinx 1.5.1) use sphinx!=1.5?

No idea :)

mgeier:
I'll undo this in #79. If there are objections, please comment there.

michaelpacer:
Sphinx 1.5.1 doesn't include the fix, it won't be available until Sphinx 1.5.2.

mgeier:
@michaelpacer Are you talking about https://github.com/sphinx-doc/sphinx/pull/3217 not being available in Sphinx 1.5.1?

Or rather about https://github.com/sphinx-doc/sphinx/pull/3224?

michaelpacer:
Ah, I'd thought they were related.

But I was referring to the discussion in sphinx-doc/sphinx#3224
<https://github.com/sphinx-doc/sphinx/pull/3224> describing the fix as not
available in 1.5.1.

mgeier:
Thanks for the clarification.

AFAIK, the two issues are not related, they just happened to break at the same Sphinx version.

The first one should be resolved in Sphinx 1.5.1, the second one will hopefully be fixed with #85.

Issue #78, opened 2016-12-11
LaTeX build breaks with Sphinx version 1.5


mgeier:
That's because I'm assuming that the whole `sphinxVerbatim` environment is contained in a single list element in `self.body` here: https://github.com/spatialaudio/nbsphinx/blob/4f50816d1252cecc28e3472a99c02eb0a3706704/nbsphinx.py#L1162

Sadly, https://github.com/sphinx-doc/sphinx/pull/3022 broke that assumption.

mgeier:
https://github.com/sphinx-doc/sphinx/issues/3224 will hopefully fix this, and #79 will disable Sphinx 1.5.

michaelpacer:
Ok, sphinx 1.5.1 does not fix this(I think); but I can't be sure because I can't locally reproduce this issue in any case :/.

mgeier:
Does #85 fix this?

michaelpacer:
I can confirm, this fixes the error on both 1.5.0 and 1.5.1 (with docutils 1.12.0).

PR #79, opened 2016-12-11
Avoid using the broken Sphinx version 1.5

mgeier:
See #77 and #78.

This should be merged as soon as Sphinx 1.5.1 (which hopefully fixes https://github.com/sphinx-doc/sphinx/pull/3224) is released.


SylvainCorlay:
+1

michaelpacer:
1.5.1 is released!

michaelpacer:
Unfortunately, I don't think it solves the problem :/ as when I explicitly use 1.5.1 in RTD, I'm still getting an error from `depart_code_latex`.

michaelpacer:
Oh, that update is delayed until 1.5.2 now, which means this will remain broken for another month.

mgeier:
If #85 fixes #78, I can merge this now, right?

mgeier:
... and then I could make a new `nbsphinx` release in the next few days.

michaelpacer:
Yes!

Issue #81, opened 2016-12-13
embed markdown tables automatically generated


gpagliuca:
I would like to embed tables and plots generated on the fly in markdown cells based on some (hidden) initialization cells.  

This is possible in the jupyter notebook thanks to this extension [python-markdown](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/wiki/python-markdown). 

According to the extension documentation it is possible to have the same behavior exporting via nbconvert using a pre-processor (here the [preprocessor source code](https://github.com/paulovn/nbextensions/blob/master/pre_pymarkdown.py)) :

> Exporting
> In order to have nbconvert show the computed Python output when exporting to another format, use the pre_pymarkdown.py preprocessor. If you used the python setup.py install command to install the IPython-contrib extension package, this will already be installed.
> For manual setup, you need to copy this file to a location within the Python path (or extend PYTHONPATH). Additionally, you need to add these two lines to your jupyter_nbconvert_config.py configuration file:
> c = get_config()
> c.Exporter.preprocessors = ['pre_pymarkdown.PyMarkdownPreprocessor'] 

This works when I manually convert a notebook to html using nbconvert, but not when I try to build the doc with sphinx. 

Any suggestion? Is this the best approach? (related in end to issue #15)

Thanks ;-)

mgeier:
For now, `nbsphinx` doesn't support adding custom preprocessors, but it shouldn't be too hard to implement this.

Would you like to make a PR for this?

gpagliuca:
I am more a user... but I'll try.

Thanks for the answer.

mgeier:
If you need help, don't hesitate to ask!

gpagliuca:
Created a pull request with the hide_input option as cell metadata. It works but it is ugly.

I need your help for the pre-processor instead.
The pre-processor I'd like to add extends the nbconvert Preprocessor class and it is automatically called when using `nbconvert` with the jupyter nbextensions installed.

Now, in `nbsphinx`, if I have well understood, the only preprocessor used is `HighlightMagicsPreprocessor` in the `Exporter `class but if I simply try to add there `PyMarkdownPreprocessor` I see no change or error.

mgeier:
AFAIK, the thing with the `HighlightMagicsPreprocessor` only works because it is in the list of default preprocessors:
[...]

PR #86, opened 2016-12-21
Added the hide_input option

gpagliuca:
This partially covers #81 and #6. 
With `"hide_input": true` as cell metadata the cell input is hidden but not the output.

This solution is not very elegant since the "Out[10]" is still there
untitled

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -91,6 +91,12 @@ RST_TEMPLATE = """
 {% endif %}
 {%- endblock any_cell %}
 
+{% block input_group -%}                                                        
+{%- if cell.metadata.hide_input -%} 
+{%- else -%} 
+{{ super() }} 
+{%- endif -%} 
+{% endblock input_group %}
 
 {% block input -%}
 .. nbinput:: {% if cell.metadata.magics_language -%}


mgeier:
Thanks for the PR.
Did you try to set `nbsphinx_prompt_width` to zero?
There is also a CSS class named `prompt`, which you can probably use.
Note that those won't affect LaTeX output.

BTW, what HTML theme are you using?
The prompts are not supposed to have a scrollbar.
But probably this should somehow be disabled with CSS?

About your PR: I think it's better to wait a bit to see how the "hiding" discussion in Jupyter turns out.

You should join the discussion!

There is some discussion in #65, which has links to more discussions in the Jupyter repos.

gpagliuca:
Sorry for the late reply, I was offline.

The theme is alabaster and what you see in the image is a pandas dataframe (not sure if this matters)

mgeier:
Thanks for the update!
I've changed the CSS in 51ca141e089227d5647a3a3f8a84918fccf2d118, can you please show how it looks like when you use the `master` branch?

gpagliuca:
Better, thanks!

untitled

Is it possible to remove also the "Out[10]" and maybe align to the left (or center)?

mgeier:
Thanks for trying it again, but could you please try it with a two-digit execution count?

> Is it possible to remove also the "Out[10]" and maybe align to the left (or center)?

Sure, everything is possible.
But I don't know yet what's the best way to do that.
I'd have to look into that, but I don't know when I'll have time for that.

If someone else wants to give it a shot, go ahead!

gpagliuca:
> could you please try it with a two-digit execution count?

untitled

mgeier:
Thanks.
So the scrollbar is still there ... are you sure you were using the `master` branch?

gpagliuca:
> are you sure you were using the master branch?

I think so...

untitled

mgeier:
Can you please check if you have this CSS in the output HTML file?

```css
div.nbinput div.prompt pre,
div.nboutput div.prompt pre {
    overflow: hidden;
}
```

gpagliuca:
> Can you please check if you have this CSS in the output HTML file?

```
div.nbinput div.prompt pre,
div.nboutput div.prompt pre {
    overflow: hidden;
}
```

This is not present.

An interesting thing: in my test case I have essentially only 1 notebook called "preprocessor". Because of that my index points only at the preprocessor.html file.
Now, in the picture below on the right side there is the preprocessor.html open directly with the browser, on the left side the same file open from the index:

screenshot_2017-01-18_10-47-11

the horizontal scroll bar is present only on the left.

[Here the 2 html files and the original notebook](https://github.com/spatialaudio/nbsphinx/files/713553/test_files.zip)

mgeier:
If the mentioned CSS is not present, you are not using the `master` branch.

In the screenshot you show two files with the same name (`preprocessor.html`) but different paths.
Are they supposed to be the same file?

The right one seems to be missing the external CSS (it only has the one from `nbsphinx`, which is embedded in the HTML).
You should check for error messages in the Chromium Developer Tools.

rddaz2013:
Q: 'nbsphinx_prompt_width' in the conf.py from sphinx or where to set?

mgeier:
A: yes

Issue #38, opened 2016-04-13
rst_prolog and rst_epilog break notebook loading


wkerzendorf:
Any ideas. Here is the log:
[sphinx-err-glebxw.txt](https://github.com/spatialaudio/nbsphinx/files/216808/sphinx-err-glebxw.txt)

mgeier:
`nbformat` seems to have problems reading the notebook.
Can you try to upgrade the `nbformat` package?

If that doesn't change anything, can please you provide the notebook in question?
Or a stripped-down version that still produces the error?

wkerzendorf:
```bash
docs git:(master) ✗ conda install nbformat
Fetching package metadata: ....
Solving package specifications: ..............
# All requested packages already installed.
# packages in environment at /Users/wkerzend/anaconda3/envs/xtool:
#
nbformat                  4.0.1                    py27_0 
```

wkerzendorf:
Thanks for your help

https://github.com/eso/xtool/blob/master/docs/notebooks/cookbook_fitting.ipynb

mgeier:
Your notebook works for me in both Python 2.7 and 3.5, I could simply include it in the `nbsphinx` docs.

Can you try to uninstall `nbformat` and re-install it again?

Or probably try everything in a fresh virtual environment?

I couldn't run your Sphinx build though, because there was no `setup.cfg`.

mgeier:
You could also try to isolate the problem with this:

```python
import nbformat
nbformat.read('cookbook_fitting.ipynb', 4)
```

wkerzendorf:
That works: 

```python
nbformat.read('notebooks/cookbook_fitting.ipynb', 4)
Out[4]: 
{u'cells': [{u'cell_type': u'markdown',
   u'metadata': {},
   u'source': u'### Modelling XShooter data with xtool ###'},
  {u'cell_type': u'code',
```

wkerzendorf:
Here is the conf: https://github.com/eso/xtool/blob/master/docs/conf.py

mgeier:
I get `NoSectionError: No section: 'metadata'` and I suspect that's because there is no `setup.cfg`.

But anyway, can you please try to make a new Sphinx project just including this one notebook and see if that works?

wkerzendorf:
Found this now: *** ValueError: Extra data: line 295 column 1 - line 298 column 1 (char 5900 - 5934)
[...]

PR #88, opened 2017-01-10
Add 'supported' attribute to NotebookParser

adrn:
This adds an empty attribute `supported` to the `NotebookParser` class so that Sphinx's `get_parser_type()` uses this parser for .ipynb files instead of thinking they are restructured text.

See also https://github.com/spatialaudio/nbsphinx/issues/38#issuecomment-271595602


mgeier:
Thanks!

Issue #90, opened 2017-01-17
'no kernel named ...', problem with conda env's


arsenovic:
when i try to execute notebooks during the sphinx build with   `nbsphinx_execute = 'always'` in `conf.py` i get 
                                                                        
    Notebook error:
    NoSuchKernel in tutorials/Calibration.ipynb:
    No such kernel named conda-env-py35-py
    make: *** [html] Error 1

i am using  conda, and am running sphinx build from a conda env. i have tried to change the kernel setting in the offending notebook to many different options, the obvious first choice was the env i am running sphinx build in. they all give errors. 

it seems to indicate the attempt to set the kernel is failing.

mgeier:
The first thing you should try is if you can execute it with `nbconvert`:

```
python3 -m nbconvert --to html --execute tutorials/Calibration.ipynb
```

(or something similar, I didn't try this command)

Then you should look at the notebook metadata, especially the `kernelspec` section.

arsenovic:
this mentions the `kernel_name` argument, http://nbconvert.readthedocs.io/en/latest/execute_api.html#executing-notebooks-using-the-python-api-interface

mgeier:
This should be resolved by #98.

PR #91, opened 2017-01-19
add kernel name

arsenovic:
adds `kernel_name` as a parameter to nbsphinx. 

you may want to cherry pick this, as i did not mean to modify the notebooks.


mgeier:
Thanks. But wouldn't it be better to fix the actual problem?

Notebooks have already metadata for kernel selection, why should we have a separate mechanism for `nbsphinx`?

Apart from that, your PR breaks all non-`python` notebooks.

mgeier:
For reference, see #90 and https://github.com/jupyter/nbconvert/issues/515 and the Jupyter mailing list: https://groups.google.com/forum/#!msg/jupyter/h7yb1nA99pc/dKFYFLy3BAAJ.

arsenovic:
not sure i get your point,   given that manually setting the kernel is a option for jupyter why shouldnt nbsphinx support it? the use case is precisely to 'overide' the kernel setting per notebook.

mgeier:
I didn't know about this option. Can you please provide a link to the docs where it is described?

An option in `conf.py` (analogous to the jupyter option) would probably make sense, but why additional notebook metadata?

As you've implemented it now, it overwrites the kernel by default, I guess that's not the intended behavior (i.e. the one from jupyter)?

mgeier:
For reference: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jupyter/RYoVU314oyM/B8GZBrB9CAAJ

mgeier:
@arsenovic Thanks for the update.
You still haven't answered my question why you are checking for notebook metadata.
Can you please elaborate?

arsenovic:
the idea is to allow two methods of operation: 
* either use the kernel defined in the notebook , or 
* use a kernel defined in the conf.py. 

i need the later to work around the conda kernel /notebook bug as you linked.

mgeier:
Yes, but the kernel is already defined in the notebook metadata (in `kernelspec`), so why do you need to define it as well in the `nbsphinx`-specific metadata?

arsenovic:
are commenting on my implementation or the general idea of having nbsphinx override what the notebook has for a kernelspec?

mgeier:
I'm talking about your implementation, about this very PR here. I would like to get it merged, but for now it doesn't make sense to me.

I think the config option `nbsphinx_kernel_name` makes sense, and since you changed the overwriting behavior, this also makes sense.

What I was trying to ask you, is why you are using `nbsphinx_metadata.get('kernel_name')`?

arsenovic:
i thought that was accessing the per-notebook kernel settings in the notebook metadata, is this incorrect?

arsenovic:
it might be faster for you to implement it yourself, rather than us debating the implementation of  a small feature.

mgeier:
Don't worry, I like debating.
The implementation is simple, it's much more important to find the right features to implement and how they should be used.

There might be a good reason why you would want to add `nbsphinx`-specific notebook metadata to select a kernel. In that case we should enable it.
But up to now you still didn't tell me why that would make sense.

> i thought that was accessing the per-notebook kernel settings in the notebook metadata

Yes it does.
My question is still: why?

arsenovic:
the notebook file supports a kernel_name in its metadata. nbsphinx currently uses this to determine the kenerl.  i was trying to add an option to allow this feature to remain.  im not sure ive implemented it correctly based on your responses.

mgeier:
As you have implemented it (if I understood correctly), there are two different ways to specify a kernel in the notebook metadata:

1. `kernelspec.name` (?)
1. `nbsphinx.kernel_name`

My question is: why do we need two different ways to do the same thing?
Or am I missing something?

arsenovic:
you can call me on google hangouts if you want to discuss this further.

mgeier:
@arsenovic I couldn't reach you and didn't hear back from you, so I gave it a shot myself: #98.

Does that work for you?

PR #94, opened 2017-01-22
Make in memory template name match it's declaration

michaelpacer:
With nbconvert 5.1.1+ in-memory templates need to be declared with the same "file name" as they are then invoked with. This relates to the changes made to allow in memory templates more generally because it's fine now for them to have no extension while being loaded. 

Right now nbsphinx is looking for `nbsphinx-rst` when it has loaded `nbsphinx-rst.tpl`. 

This fix should address the issue with nbsphinx and nbconvert 5.1.1 that is currently stopping docs that use nbsphinx from building. I've tested it locally on python2 and 3 and it makes the nbconvert docs build again.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -497,7 +497,7 @@ class Exporter(nbconvert.RSTExporter):
         self._allow_errors = allow_errors
         self._timeout = timeout
         self._codecell_lexer = codecell_lexer
-        loader = jinja2.DictLoader({'nbsphinx-rst.tpl': RST_TEMPLATE})
+        loader = jinja2.DictLoader({'nbsphinx-rst': RST_TEMPLATE})
         super(Exporter, self).__init__(
             template_file='nbsphinx-rst', extra_loaders=[loader],
             config= traitlets.config.Config(


mgeier:
Thanks, but your solution doesn't seem to work with nbconvert 5.1.0, it gives me:

```
Notebook error:
TemplateNotFound in allow-errors.ipynb:
nbsphinx-rst.tpl
```

It works when I add the extension `.tpl` to both strings.

michaelpacer:
That's extremely strange. 

I just tried it with 5.1.1 and 5.1 our docs build with and without this one commit.

It fails with that error if I _don't_ include this commit (w/ `git checkout bc383ec0a`) and it succeeds if I do include this commit (w/ `git checkout a4886338`). 

Are you sure you're not using nbconvert 5.0.0? Because without the changes to in memory template loading in 5.1.1/5.1.0 this won't fix things.

mgeier:
Sorry, I was very imprecise!
I was simply using the version I had checked out at that time (which happened to be https://github.com/jupyter/nbconvert/commit/d7c996c87db98c2bb80e72a0c22511a802064bc4) and I just wanted to say that it's something lower than the new release ...

Anywho, I was wondering why you removed the extension from one string instead of adding it to the other. Is there a reason?

Because if there's not, you could just add it to both, then it would even work with my randomly chosen commit ...

michaelpacer:
I'd approached it that way because to me, in-memory templates don't need to have an "extension" if they're not being saved to the file system. It makes it clearer to developers when they run into an issue that they don't need to look for a template file being shadowed on the file system (a problem that can occur easily) versus an object that's being created in-memory and so it may need to be tracked down in an interpreted fashion. This doesn't matter too much in your case, since the declaration and use are so near each other.

I can see a use where they should disagree: if you also had a similar template that was directed toward LaTeX  (just with a `.tplx` extension) and you wanted subclasses of the exporter class to be able to choose either of them from memory as needed. But https://github.com/jupyter/nbconvert/pull/518 should address that use case.

takluyver:
#95 takes the other approach. I agree that it shouldn't be necessary to specify an extension here, but as it is in some versions of nbconvert, let's go for what works.

mgeier:
Thanks @michaelpacer for the explanations. As you say, the problem you described isn't that critical in my case. I merged @takluyver's PR because it works in at least one known commit where yours doesn't.

Thanks again!

Issue #93, opened 2017-01-22
NBSphinx builds broken with nbconvert-5.1.1 (TemplateNotFound ... nbsphinx-rst)


timstaley:
Hi folks,

just a heads-up that nbsphinx seems to be incompatible with the latest release of nbconvert, [5.1.1](https://github.com/jupyter/nbconvert/releases/tag/5.1.1). The error looks something like this:

    Notebook error:
    TemplateNotFound in tutorial/01-parsing.ipynb:
    nbsphinx-rst
    Makefile:45: recipe for target 'html' failed

I've not dug into the underlying cause but I'd hazard a guess it's something to do with changes to nbconvert's template handling, cf https://github.com/jupyter/nbconvert/issues/490

A simple workaround is to specify `nbconvert==5.0.0` in the requirements file.

michaelpacer:
An easier fix is to merge https://github.com/spatialaudio/nbsphinx/pull/94 which fixes the issue by getting the file names to match.

michaelpacer:
also trying to fix in nbconvert itself: https://github.com/jupyter/nbconvert/pull/518

PR #95, opened 2017-01-24
Specify template name including extension

takluyver:
I think this should work on all versions of nbconvert.

Closes gh-93
Alternative to gh-94


mgeier:
Thanks!

PR #99, opened 2017-03-01
Add jupyter-js-widgets embedder javascript

### LINKED ISSUE #84: How to include the JavaScript for widgets?
Is it really necessary to load `jupyter_sphinx.embed_widgets`?

Since #76, only adding the JS file is necessary, nothing else from `jupyter_sphinx.embed_widgets` is needed, right?

What about adding a config option like `nbsphinx_widgets`?
This could be `False` by default and `True` if widgets should be displayed.
It could even be set to an alternative path to the JS file (e.g. for offline use).

See also https://github.com/jupyter/jupyter-sphinx/issues/3.

SylvainCorlay:
We have just released ipywidgets 6.0.

This complete the work on the rendering of jupyter widgets in nbsphinx-generated docs.

Closes #84.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -1278,6 +1278,9 @@ def setup(app):
     app.add_node(AdmonitionNode,
                  html=(visit_admonition_html, depart_admonition_html),
                  latex=(visit_admonition_latex, depart_admonition_latex))
+
+    app.add_javascript('https://unpkg.com/jupyter-js-widgets@^2.1.4/dist/embed.js')
+
     app.connect('builder-inited', builder_inited)
     app.connect('html-page-context', html_page_context)
     app.connect('html-collect-pages', html_collect_pages)


mgeier:
Thanks!

I have 2 questions though (which might or might not be related):

1. Should the JS really be added unconditionally?

2. Wouldn't it be good to enable offline use of the resulting docs?

Both points could probably be solved with a config setting, see #84 (which I guess should be closed by this PR).

SylvainCorlay:
> Should the JS really be added unconditionally?

On this point, I think that the complication of the code to add it conditionally on whether a page has widgets is not worth the benefit. However, I am +1 on providing a general opt-out or to make it an opt-in.

> Wouldn't it be good to enable offline use of the resulting docs?

Sure, however, the script currently may try to load other resources on unpkg for custom widgets.

Making a pure offline version of the script is totally in roadmap for the next ipywidgets release.

mgeier:
@SylvainCorlay 

> I think that the complication of the code to add it conditionally on whether a page has widgets is not worth the benefit.

How much complication would this actually be?
Is there a reliable way to check if any widgets are used in a given notebook?

Notebooks might only be a small part of a Sphinx project, so if it's reasonable, I would like to try to avoid loading the JS on all pages, same as I try to include CSS where it is not needed.

In this case we wouldn't need an opt-in/opt-out mechanism, but it would still be good to provide a setting for offline use (or an alternative CDN?).
Even if a full-featured offline version is not yet available, we could already implement this setting now, right?

mgeier:
@SylvainCorlay ping?

PR #104, opened 2017-03-25
A few CSS manipulations for the 'cloud' theme

mgeier:
This makes more space for the info and warning logo.

It also creates a nice title line when multiple paragraphs are used in an admonition.


mgeier:
For the record: On http://readthedocs.org, this currently doesn't seem to work for some unknown reason when using `requirements.txt`, but it works fine when using `conda`.

PR #106, opened 2017-03-28
add line break after the adjacent ANSI cell

### LINKED ISSUE #105: Latex: a line break should be added at the end of an adjacent ANSI cell
Hi,

Thank you for this useful project. But when I built latex from a notebook, I found that adjacent ANSI cell wouldn't start a new line, so it would make something like this:

```latex
\begin{OriginalVerbatim}[commandchars=\\\{\}]
 some code
\end{OriginalVerbatim}\begin{itemize}
 some other code
\end{itemize}
```

In this example, the ``itemize `` cell won't be compiled correctly.

Harlan-pku:
Hi, all,

I have made a simple commit to fix #105


mgeier:
Thanks!

Can you please provide an example where this bug shows itself?

Also, I'm having a hard time understanding why your fix works and what's the difference to inserting a literal new line.
Can you please elaborate?

Harlan-pku:
An example is available [here](https://github.com/Harlan-pku/nbsphinx/blob/master/example/example.ipynb). When converted to latex, it will be like this:

``` latex
% This comment is needed to force a line break for adjacent ANSI cells
\begin{OriginalVerbatim}[commandchars=\\\{\}]
/var/matlab/ifog_analysis/get_tems_data.m
\end{OriginalVerbatim}\begin{itemize}
\item {} 
Help information of get\_tems\_data.

\end{itemize}
```

When I use xelatex to compile this tex file, some errors occur:

``` log
! FancyVerb Error:
  Extraneous input `\begin {itemize}\end {}' between \end{OriginalVerbatim} and
 line end
.
\FV@Error ...ncyVerb Error:^^J\space \space #1^^J}
                                                  
l.1188 \end{OriginalVerbatim}\begin{itemize}
                                            

! LaTeX Error: Lonely \item--perhaps a missing list environment.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.1189 \item {
              }

Underfull \hbox (badness 10000) detected at line 1189
[][]

! LaTeX Error: \begin{document} ended by \end{itemize}.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.1192 \end{itemize}
                    
! Extra \endgroup.
<recently read> \endgroup 
                          
l.1192 \end{itemize}
```
Please ignore the line number, cause they are not correct.

mgeier:
@Harlan-pku Thanks! I can reproduce the problem.

However, your fix doesn't seem to work for me (on Linux).
Do you happen to use Windows?

Do you have another idea how to fix this?

Harlan-pku:
Yes, I use Windows and the fix worked for me. And I think it's sphinx's bug instead of nbsphinx's bug. I wrote a rst file and a similar error occurred.


Source code:
```rst
hello
=====

some code

.. raw:: latex

	\begin{OriginalVerbatim}[commandchars=\\\\\\{\\}]
		hello too
	\end{OriginalVerbatim}

			
* item1
* item2
```

The LaTex output:
```latex

\chapter{hello}
\label{\detokenize{test:hello}}\label{\detokenize{test::doc}}\label{\detokenize{test:welcome-to-my-test-s-documentation}}
some code
\begin{OriginalVerbatim}[commandchars=\\\\\\{\\}]
        hello too
\end{OriginalVerbatim}\begin{itemize}
\item {} 
item1

\item {} 
item2

\end{itemize}
```

Harlan-pku:
I have reported it to sphinx
https://github.com/sphinx-doc/sphinx/issues/3594

Harlan-pku:
@mgeier the contributor of Sphinx @jfbu said this is not Sphinx's bug, and Sphinx always treat ``..raw:: latex`` as inline text. So, I think we can add a ``\relax`` into the raw latex cell to fix this bug.

mgeier:
Thanks for resolving this!

At https://github.com/sphinx-doc/sphinx/issues/3594 there is now actually some discussion if the behavior of Sphinx should be changed ... anyway, with our `\relax` we should be safe, even if they decide to insert a newline for raw LaTeX directives.

PR #108, opened 2017-04-01
LaTeX: Improve handling of raw cells.

mgeier:
* Raw cells with "None" type will now be on their own paragraph in LaTeX
* Raw cells with "LaTeX" will not, but the user can add blank lines
* More docs


mgeier:
This is mostly obsoleted by Sphinx 1.6.

I'll wait some time and then merge the remaining changes.

PR #111, opened 2017-05-02
Use refdomain='std' for pending_xref() on Sphinx >= 1.6

mgeier:
See https://groups.google.com/forum/#!topic/sphinx-dev/98bvia7i7II


mgeier:
This seems to be a backward-incompatible change and it will hopefully be fixed in Sphinx: https://github.com/sphinx-doc/sphinx/issues/3698.

mgeier:
For now, this is fixed in https://github.com/sphinx-doc/sphinx/pull/3699, but it will break again in Sphinx 2.0.
Some time before that (but not too soon), we should merge this PR.

mgeier:
I've added a check for the Sphinx version. Now it should work on old and new versions.

PR #112, opened 2017-05-02
Use sphinx.util.status_iterator()

mgeier:
... instead of app.status_iterator(), which was deprecated in Sphinx 1.6


mgeier:
This will be removed in Sphinx 1.7.

jschueller:
+1 for this

jschueller:
maybe this could be implemented in a way that it remains compatible with older sphinx ?

mgeier:
Fixed in #115.

PR #113, opened 2017-05-24
quick fix for deprecation warning

mpacer:


mgeier:
Thanks, I have this already queued up in #112.

I just didn't know which older Sphinx versions this breaks, so I didn't apply it yet.
Do you happen to know that?

mpacer:
Not sure, sorry!

Issue #36, opened 2016-04-05
Allow conversion via markdown instead of RST


bnaul:
Currently there are some markdown formatting difficulties that can arise when converting to HTML. For example, a link like ``[`nbsphinx`](https://github.com/spatialaudio/nbsphinx)`` which should render as [`nbsphinx`](https://github.com/spatialaudio/nbsphinx) (cf. [nbsphinx](https://github.com/spatialaudio/nbsphinx)) won't be displayed properly because this text can't easily be represented in the intermediate RST.

Since `sphinx` can easily produce documents directly from markdown (e.g. using `recommonmark`), it would be nice if it were possible to skip the RST altogether as this would more reliably reproduce the desired formatting from the notebook. Currently I am using `nbconvert` to get markdown output and then producing HTML with Sphinx instead of using `nbsphinx` because of such formatting issues.

cc: @fperez @takluyver

takluyver:
Thanks @bnaul . Also pinging @willingc, who works quite a bit on our own docs.

willingc:
Thanks @takluyver. @bnaul If you have a sample notebook, I would be happy to look at it. In general, there are pros and cons with different conversion workflows. FWIW, we've been using ``recommonmark`` and ``nbsphinx`` with good success that past few months on Jupyter projects. Also pinging @mgeier for input.

willingc:
@bnaul Another thing to look at is the exact versions that you are using in Sphinx, nbsphinx, and recommonmark. Certain combinations have more issues than others.

takluyver:
To clarify, Fernando and I were just talking to Brett in person, and asked him to open this issue. While there might be workarounds for any specific case, that's a game of whack-a-mole: the non-code parts of notebooks are markdown, so translating to rst before loading notebooks into Sphinx always runs the risk of messing something up.

We suggested that maybe nbsphinx should depend on recommonmark or something similar, and cut the rst stage out of the loading process. I may be able to spare some time to look into this, though I don't know much (anything?) about the Sphinx API.

willingc:
Ahh...thanks for the context @takluyver. I would be happy to have a direct conversion from markdown in
notebook cells to HTML. The less conversions the better.

mgeier:
Thanks @bnaul for bringing this up and thanks all for the comments!

[...]

Issue #29, opened 2016-03-23
Check if cell attachments work


mgeier:
... and include an example in the docs.

http://nbformat.readthedocs.org/en/latest/format_description.html#cell-attachments

mgeier:
This will be available in The Notebook version 5.0, see https://github.com/jupyter/notebook/pull/1965.

mgeier:
They don't work currently, I've opened #114 to eventually implement this.

PR #114, opened 2017-06-09
Support for cell attachments

mgeier:
For now, only the documentation is there, the implementation is missing.
Currently, we only have access to the cell metadata from within the Jinja template, and I don't know how to save the attachment data there.
I guess it has to be saved to disk for Sphinx to be able to use it.

I'm open for suggestions!

Once #36 is implemented, this should be easier.

See also #29.


mgeier:
Thanks to ideas from @seidlr in #214 I've implemented a hacky solution for this: 69057ef000ef235459436929c5b04394eba9270a.

PR #115, opened 2017-06-26
Use sphinx.util.status_iterator if available

jschueller:
My proposition for #112, doesn't require changes in setup.py and allows to get a release before sphinx 1.7.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -1174,8 +1174,12 @@ def html_collect_pages(app):
     files = set()
     for file_list in getattr(app.env, 'nbsphinx_files', {}).values():
         files.update(file_list)
-    for file in app.status_iterator(files, 'copying linked files... ',
-                                    sphinx.util.console.brown, len(files)):
+    try:
+        status_iterator = sphinx.util.status_iterator
+    except AttributeError:
+        status_iterator = app.status_iterator  # For Sphinx < 1.6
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks!

The version check (with the added dependency) seems a bit heavy to me.
What about just trying the new one and if it doesn't work, try the old one?

Something like this (untested!):

```python
try:
    status_iterator = sphinx.util.status_iterator
except AttributeError:
    status_iterator = app.status_iterator  # For Sphinx < 1.6
# use status_iterator(...) here
```

jschueller:
Yes, that's even better. I just tested it successfully.

nbsphinx.py
```diff
@@ -1174,8 +1174,13 @@ def html_collect_pages(app):
...
         files.update(file_list)
-    for file in app.status_iterator(files, 'copying linked files... ',
-                                    sphinx.util.console.brown, len(files)):
+    try:
+        status_iterator = sphinx.util.status_iterator
+    except AttributeError:
+        status_iterator = app.status_iterator  # For Sphinx < 1.6
+    for file in status_iterator(files, 'copying linked files... ',
+                                sphinx.util.console.brown,
+                                len(files)):
```
mgeier: The line break here isn't necessary anymore.

jschueller: ok, fixed

mgeier:
Thanks @jschueller!

PR #129, opened 2017-10-06
Interpret links to reST sub-sections

tristan0x:
This pull-request provides a fix to allow links from ipynb to reST sub-sections.
Documentation has been updated accordingly, with an example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ build/
 dist/
 nbsphinx.egg-info/
 .ipynb_checkpoints/
+.python-version
--- a/doc/markdown-cells.ipynb
+++ b/doc/markdown-cells.ipynb
@@ -246,7 +246,13 @@
     "[A reStructuredText file](a-normal-rst-file.rst)\n",
     "```\n",
     "\n",
-    "Links to sub-sections are not (yet?) possible."
+    "Links to sub-sections are also possible. Example: [Sphinx Directives](a-normal-rst-file.html#sphinx-directives-for-info-warning-boxes)",
+    "\n",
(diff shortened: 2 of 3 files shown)


mgeier:
Thanks for the PR!

This doesn't work for LaTeX output though, any ideas how to handle that?

clenk:
Any chance this could get added but with a note in the documentation that it doesn't work in LaTeX output, like the note for links to orphan notebooks (http://nbsphinx.readthedocs.io/en/0.2.14/markdown-cells.html#Links-to-Other-Notebooks)?

I thought maybe since you can link to sections in .ipynb notebooks the way to get LaTeX working would be to do something similar to [what's done for ".ipynb#" links](https://github.com/tristan0x/nbsphinx/blob/d00a1eabadf764ba9692cd5fea8bedaa9ebe6771/nbsphinx.py#L1013) but that led to `(if the link has no caption the label must precede a section header)` warnings I haven't been able to fix.

mgeier:
In general it's possible to have HTML-only features as long as they are marked as such in the docs (as @clenk mentioned).

However, it would be nicer if this could be done in a more general way, also allowing the links to work in LaTeX.
What about creating a `Transform` similar to `CreateSectionLabels` (or generalizing it to arbitrary file extensions if possible) and running it for all RST files with [app.add_transform()](http://www.sphinx-doc.org/en/stable/extdev/appapi.html#sphinx.application.Sphinx.add_transform)?

mgeier:
Thanks @tristan0x and @clenk, #135 was merged which includes and extends this PR.

Issue #130, opened 2017-10-06
Disable warnings for links to non-existing local files


clenk:
**tldr; Please support links to sub-sections in non-notebook files.**

I'm using nbsphinx with autodoc and autosummary and would like to have a link from one of the notebooks to one of the pages that will be generated by autodoc. In a markdown cell in the notebook I use a link like `[something](../api/something.html#heading)`. In the final output the link works fine, but during the build process sphinx complains `WARNING file not found: 'api/something.html#heading'`.

After reading [this page in your docs](http://nbsphinx.readthedocs.io/en/0.2.14/markdown-cells.html#Links-to-*.rst-Files-(and-Other-Sphinx-Source-Files)), I added '.html' to `source_suffix` in my sphinx configuration. It doesn't throw a warning if I remove the anchor part of the link, but I need to be able to link to a specific section on the page.

The link works fine in the final output, but I want to use a lot of these links and I'm afraid sphinx dumping a ton of these warnings will make it harder to notice more important errors.

clenk:
The fix in #129 works for me. I hope the LaTeX issue can be resolved and the PR merged in soon.

mgeier:
It wouldn't be a problem to add a configuration option for suppressing those warnings.
But I've just seen that Sphinx recently introduced their own system for suppressing warnings:
https://github.com/sphinx-doc/sphinx/pull/2346
http://www.sphinx-doc.org/en/stable/config.html#confval-suppress_warnings

What about just adding `type` and `subtype` to all `nbsphinx` warnings?

BTW, you should not add `.html` to `source_suffix`, that doesn't make sense.

clenk:
Yes, adding `type` and `subtype` to `nbsphinx` warnings seems like a better way to go.

Adding `.html` felt weird but I was just trying stuff out to see what would work.

mgeier:
@clenk Would you like to make a PR for that?

clenk:
Unfortunately It's unlikely I will have time to, and #135 solved my problem.

mgeier:
@clenk No problem.

I think this feature has been added in Sphinx 1.4 and the documentation says "Now, this option should be considered *experimental*", so we should probably wait a bit anyway before implementing this.

I'm leaving this issue open for now ...

mgeier:
I think this can be closed.

If someone still has a problem with those warnings, please leave a comment here!

PR #133, opened 2017-10-12
Allow links to Python entities

clenk:
This PR lets you create links to things in the Python sphinx domain like you would in ReST. So for example:

    [py:module:mymodule](mymodule)

This is especially helpful when paired with `autodoc`. Works for html and LaTeX.
Related to #130.

### DIFF
--- a/doc/markdown-cells.ipynb
+++ b/doc/markdown-cells.ipynb
@@ -266,6 +266,21 @@
     "The linked files are automatically copied to the HTML output directory.\n",
     "For LaTeX output, no link is created."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Links to Python Domain\n",
+    "\n",
+    "Links to Python modules, classes, functions, etc. (anything in the [Python domain](http://www.sphinx-doc.org/en/stable/domains.html#the-python-domain)) can be made like this:\n",
+    "\n",
+    "```\n",
+    "[py:meth:my_module.MyClass.my_method](my_method)\n",
+    "```\n",
+    "\n",
+    "This can be useful when paired with the `autodoc` Sphinx extension.\n",
+   ]
   }
  ],
  "metadata": {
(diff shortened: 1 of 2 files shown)


mgeier:
Aren't you confounding URL and link text?
If anything, shouldn't it be `[mymodule](py:module:mymodule)`?

How is your example supposed to look like if it were used in an RST file?
Are you talking about ``:py:mod:`mymodule` ``?

I'm reluctant to add "magic" URIs, is there some other system that does something like that?

Wouldn't it be better to create a MarkDown extension that allows using arbitrary Sphinx roles and directives?

clenk:
Thank you for responding. Yes, you're right - that's a typo; should be `[mymodule](py:mod:mymodule)`.

Sorry, I'm still learning how Sphinx, autodoc, and nbsphinx work. I'm not sure where in the pipeline a MarkDown extension would be added. Would it be like an additional filter in `nbpshinx`'s `markdown2rst()` function?

I also found `recommonmark`, which lets you [embed rST codeblocks in markdown](https://recommonmark.readthedocs.io/en/latest/auto_structify.html#embed-restructuredtext). It works on `.md` files, whereas I'm dealing with markdown in `.ipynb` files, but maybe a similar approach could be taken?

mgeier:
Ideally we would use a Markdown (CommonMark) library that allows syntax extensions. AFAIK, such a thing doesn't exist yet for Python (I was actually working on one but I currently don't have time to continue that).
This would allow using reST roles and directives without change in syntax, because I think the syntax doesn't collide with Markdown (except perhaps in a few rare corner cases).

Since there isn't an extensible CommonMark library, the `recommonmark` people came up with the ugly code blocks work-around, which I hate.

Sphinx roles could probably also be implemented reasonably well using an AST transformation as you mentioned above.
You could look for inline code spans and check if they are preceded or followed by `:xxx:`. I don't know the detailed syntax rules, but it might somehow work to some extent.

For support for arbitrary Sphinx directives I think we'll have to wait for an extensible CommonMark library.

mgeier:
Superseded by #135.

PR #135, opened 2017-10-16
Subsection links

clenk:
- Generalizes `CreateSectionLabels` to work on all Sphinx-parsed files
- Allows links to subsections in arbitrary file types
- Allows links to object signatures in Sphinx domains (see [desc_signature](http://www.sphinx-doc.org/en/stable/extdev/nodes.html#sphinx.addnodes.desc_signature))
- Works in both HTML and LaTeX output

Based on #129.


nbsphinx.py
```diff
@@ -1006,14 +1008,14 @@ def apply(self):
             else:
                 target = ''
 
+            subsection_matches = re.match(self._subsection_re, uri)
```
mgeier: Why not use `self._subsection_re.match(uri)`?

mgeier:
Thanks @clenk, this looks nicely generic!

However, the subsection link to RST doesn't work for me and I get the warning:

    .../nbsphinx/doc/markdown-cells.ipynb:316: WARNING: undefined label: a-normal-rst-file.html#sphinx-directives-for-info-warning-boxes (if the link has no caption the label must precede a section header)

The class `CreateSectionLabels` still mentiones "notebook" in some comments and variable names, it would be good to generalize that, too.

Can you please explain to me what this "domain-specific object signatures" stuff is and how it works?

clenk:
I'll get this cleaned up soon when I get a chance, but nice to know I'm on the right track. I think that link didn't work because it's to a `.html` file but should be to a `.rst` file. I'll make sure.

I probably used the wrong terminology with "domain-specific object signatures", but it refers to module signatures, class signatures, and function signatures from [Sphinx domains](http://www.sphinx-doc.org/en/stable/domains.html). For example, a Python function signature. It should work with any other domain though, like C++ or JavaScript. This obviates the need for #133. The existing `CreateSectionLabels` code adds labels for sections, but these code objects (classes, functions, etc.) aren't structured in a section node. So the code for "domain-specific object signatures" looks for the node type that they _are_ structured in, which is [desc_signature](http://www.sphinx-doc.org/en/stable/extdev/nodes.html#sphinx.addnodes.desc_signature). Then it adds the necessary label information to `env.domaindata` like the existing `CreateSectionLabels` code does. I'll see about adding an example to the documentation.

doc/a-normal-rst-file.rst
```diff
@@ -65,3 +65,13 @@ This is how an info box looks like:
     It may include nested formatting, even another info/warning box:
 
     .. nbwarning:: **Warning:** You should probably not use nested boxes!
+
+Domain Objects
+--------------
+
+.. py:function:: example_python_function(foo)
+
+   This is just for testing domain object links. See
```
mgeier: Same indentation as above, please.

doc/a-normal-rst-file.rst
```diff
@@ -65,3 +65,13 @@ This is how an info box looks like:
...
+   `this section <markdown-cells.html#Links-to-Domain-Objects>`_.
```
mgeier: This is not the right way to link to notebooks (doesn't work in LaTeX output), see 2 sections above.

doc/markdown-cells.ipynb
```diff
@@ -246,7 +246,13 @@
     "[A reStructuredText file](a-normal-rst-file.rst)\n",
     "```\n",
     "\n",
-    "Links to sub-sections are not (yet?) possible."
+    "Links to sub-sections are also possible. Example: [Sphinx Directives](a-normal-rst-file.rst#Sphinx-Directives-for-Info/Warning-Boxes)",
```
mgeier: Strangely, the `\n` at the end of the line is missing ... did you edit this in Jupyter?

Also, a dot is missing after the link.

doc/a-normal-rst-file.rst
clenk: Links won't work right if they are inside domain objects - I guess they aren't parsed the same way. I'll move this description/link outside the domain directive.

mgeier:
Thanks!

PR #136, opened 2017-10-22
Add slash to the beginnings of all generated labels

mgeier:
I think forcing absolute paths in labels is more consistent.
Before, the paths had to be kind of absolute already, but with this PR they are "formally" absolute by starting with a slash.

This doesn't change the usage from within notebooks, but it does break existing links from RST files to notebooks.

Does anyone have a problem with that?


jschueller:
It seems adding a backslash is not compatible with previous versions of nbsphinx, how do we ensure the same rsts works with both new and old versions ?

mgeier:
@jschueller TBH I wasn't sure if anyone is actually using this. Thanks for the feedback.

Do you use links from RST files to notebooks?
If yes, can you please show me some examples?

I could add a second set of labels using the old way (without leading slash) to allow for some transition period. But at some point, I'd like to get rid of the non-slash labels. It was my mistake that I didn't use the slash from the beginning.

jschueller:
We make extensive use of links to our example in the notebook form, here's an example (around the end of the document):
https://raw.githubusercontent.com/openturns/openturns/master/python/doc/theory/reliability_sensitivity/directional_simulation.rst

mgeier:
OK, cool.

You can try to replace `:ref:` with `:doc:`. This is the default Sphinx mechanism for referencing other "documents". And this should work regardless of this PR.
See http://www.sphinx-doc.org/en/stable/markup/inline.html#role-doc.

I've added the `:ref:` thing just to be able to link to sections within documents, which doesn't work by default in Sphinx.

Would it be OK for you to change `:ref:` to `:doc:`?
Or would you like me to add the old labels for compatibility? Forever of for some transition period?

jschueller:
Using ```:doc:``` does not seem to work it tells me "unknown document", altough I did specify its absolute path.
```Warning, treated as error:
/home/schueller/projects/openturns/schueller4/build/python/src/sphinx_build/theory/data_analysis/anderson_darling_test.rst:72: WARNING: unknown document: /examples/data_analysis/andersondarling_test.ipynb```

**Comment by jschueller** [comment]
I tried relative, and omiting the .ipynb suffix without luck

**Comment by mgeier** [comment]
I didn't know how to build your docs, so I tried it with the `nbsphinx` docs.
It indeed didn't work for me with the `.ipynb` suffix, but both of these worked:

```rst
:doc:`/subdir/a-notebook-in-a-subdir`
:doc:`subdir/a-notebook-in-a-subdir`
```

Since you are writing the link in `theory/data_analysis`, I guess you'd have to use one of those:

```rst
:doc:`../../examples/data_analysis/andersondarling_test`
:doc:`/examples/data_analysis/andersondarling_test`
```

BTW that's exactly the reason why I wanted to change this. The old way (using ``:ref:`examples/data_analysis/andersondarling_test.ipynb` ``) is not consistent with how the `:doc:` role works. The same is true for image paths.

jschueller:
Indeed, it works now, my relative path was wrong. The rendering is the same, I'll change all my links, thanks.

mgeier:
OK, good to know.
You could also just keep using the `master` branch of `nbsphinx` and just add the leading slashes to your `:ref:` links.

jschueller:
Yes but developers with non git version of nbsphinx wont be able to generate the doc.

mgeier:
OK, that's true.

PR #141, opened 2017-11-04
Hard coded the value of uuid to ensure that the package becomes repro…

am107:
For the packet to be reproducible, I needed to make sure that the same uuid is used each time the package is built. 
"With free software, anyone can inspect the source code for malicious flaws. But Debian provide binary packages to its users. The idea of “deterministic” or “reproducible” builds is to empower anyone to verify that no flaws have been introduced during the build process by reproducing byte-for-byte identical binary packages from a given source." - https://reproducible-builds.org/

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -150,11 +150,10 @@ RST_TEMPLATE = """
 
 {{ output.data['text/html'] | indent | indent }}
 {%- elif datatype == 'application/javascript' %}
-{% set div_id = uuid4() %}
 
     .. raw:: html
 
-        <div id="{{ div_id }}"></div>
+        <div id="{{ '17e41401-056b-4c7f-8bf7-a5ed668e35ea' }}"></div>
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for the suggestion, but I don't understand ...

The function `uuid4()` is *not* called at "build time". Only the file `setup.py` is executed when installing (or creating a wheel).

The UUID values have to be different for each JavaScript output cell, in order to "connect" each JavaScript chunk with the right `<div>` element in the generated HTML page.

Finally, `nbsphinx` simply does the same as `nbconvert` does: https://github.com/jupyter/nbconvert/blob/4abde40750ee9adf666e5d6cb693d086566a390c/nbconvert/templates/html/basic.tpl#L213-L220.

am107:
Hey, thanks for the quick reply. Let me look into this.

mgeier:
This was fixed in #145.

PR #143, opened 2017-11-06
Hotfix for CreateSectionLabels with duplicates

### LINKED ISSUE #142: Version 0.2.15 breaks build
Hi -- just tried upgrading to 0.2.15, and a previously working sphinx build now breaks with the following error:

```Exception occurred:```
 ``` File "/home/joe-h/miniconda3/envs/bmll-docs/lib/python3.6/site-packages/nbsphinx.py", line 1099, in apply```
   ``` title = sig['ids'][0]```
```IndexError: list index out of range```

log
[sphinx-err-f73pf565.log](https://github.com/spatialaudio/nbsphinx/files/1446355/sphinx-err-f73pf565.log)
 file attached

clenk:
It seems that if there are domain objects with the same name in a single document, Sphinx only creates a link for the first one, so the 'ids' of the subsequent ones will be empty. We only need to create a label for the first one, as the others aren't linkable anyway.

Fixes #142.

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -1094,7 +1094,11 @@ class CreateSectionLabels(docutils.transforms.Transform):
 
         # Create labels for domain-specific object signatures
         for sig in self.document.traverse(sphinx.addnodes.desc_signature):
-            title = sig['ids'][0]
+            try:
+                title = sig['ids'][0]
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks a lot for the quick fix!

PR #145, opened 2017-11-07
Please make the build reproducible.

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that nbsphinx could not be built reproducibly.

Patch attached that uses different Javascript to get the sibling
``<div/>`` element instead of generating a (random) UUID and putting
that in the DOM.

 [0] https://reproducible-builds.org/

### DIFF
--- a/nbsphinx.py
+++ b/nbsphinx.py
@@ -150,13 +150,12 @@ RST_TEMPLATE = """
 
 {{ output.data['text/html'] | indent | indent }}
 {%- elif datatype == 'application/javascript' %}
-{% set div_id = uuid4() %}
 
     .. raw:: html
 
-        <div id="{{ div_id }}"></div>
+        <div></div>
         <script type="text/javascript">
-        var element = document.getElementById('{{ div_id }}');
+        var element = document.currentScript.previousSibling.previousSibling;
 {{ output.data['application/javascript'] | indent | indent }}
         </script>
 {%- elif datatype.startswith('application') and datatype.endswith('+json') %}


lamby:
This is being tracked in Debian at https://bugs.debian.org/881094

mgeier:
Thanks!

I think I know now what you mean ... you want to include the generated HTML docs in the Debian package, and that's where the random UUID that's currently in there doesn't allow reproducible builds.
Right?

`currentScript` doesn't seem to be available for IE: https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript
Is anyone out there still using IE?

If yes, we could probably use the work-around suggested in https://stackoverflow.com/a/4172217/

Something like

```javascript
var element = document.createElement("div");
// work-around for currentScript, see https://stackoverflow.com/a/4172217/
var scripts = document.getElementsByTagName("script");
var thisScript = scripts[scripts.length - 1];
thisScript.parentNode.insertBefore(element, thisScript);
```

This leaks the temporary variables into the following user-provided JavaScript code, so we should probably avoid that:

```javascript
var element = document.createElement("div");
(function () {
    // work-around for currentScript, see https://stackoverflow.com/a/4172217/
    var scripts = document.getElementsByTagName("script");
    var thisScript = scripts[scripts.length - 1];
    thisScript.parentNode.insertBefore(element, thisScript);
})();
```

I know this is getting quite ugly ... what do you think about it?

And just out of curiosity: why do you need to use `previousSibling` twice?

lamby:
> you want to include the generated HTML docs in the Debian package, and that's where the random UUID that's currently in there doesn't allow reproducible builds.

Correct!

> And just out of curiosity: why do you need to use previousSibling twice?

The first ``previousSibling`` matches the whitespace between the tags. :)

> [..] I know this is getting quite ugly ... what do you think about it?

Well, I've seen worse. I think it's okay as it's automatically-generated code and we can add a lot of comments. Alternatively, we could just not support IE... \*g*

mgeier:
OK, let's not care about IE.

If somebody complains, we can use the solution above.

lamby:
> let's not care about IE

♥

PR #147, opened 2017-11-10
Rewrite section anchors only for notebooks

### LINKED ISSUE #146: 2.0.16 changes id attributes in html generated from .rst
Hi,firstly thanks for the 2.0.16 fix!  On 2.0.16, the build runs but, in html generated from an .rst file, where I was previously seeing the following:
```
<div class="section" id="References">
<h1>References<a class="headerlink" href="#References" title="Permalink to this headline">¶</a></h1>
```
I now see
```
<div class="section" id="references">
<h1>References<a class="headerlink" href="#references" title="Permalink to this headline">¶</a></h1>
```
I.e. *the id has been capitalised*.  Installing 2.0.14 and 2.0.16 back and forth does toggle this behaviour so it's definitely `nbsphinx` that's doing it, even though this is not in an .ipynb file.  Is this something to do with the 2.0.15 version note "Links to subsections are now possible in all source files"?
   This might seem trivial but several typical test processes make use of these ids.  In our case it broke some integration tests.
    A good resolution for us would be either (a) to reverse it and guarantee that nbsphinx will not make changes in html generated from .rst files, or (b) to know very explicitly what nbsphinx is doing to id attributes (and have any changes to this flagged explicitly in release notes) so that we can feel secure using the ids when writing tests.
  Many thanks in advance.

mgeier:
This is an attempt to fix #146.

@clenk Does this make sense? Or do you know a better solution?

I don't really like checking for the string `'.ipynb'`, I'd rather check if the current source file was parsed with the `NotebookParser`. Is there a way to do that?


clenk:
Sorry my PR introduced so many bugs! This seems to make sense to me. I'm not familiar enough with Sphinx to know if there's a way to check what parsers have been used on the current file.

lakhman:
I inadvertently came across this, I had this library in my extensions, but wasn't using it, and it transforms all my id's to be capitalized, Once I removed it, my issue was solved.

----

To see which parser parsed your document, you can set an attribute option In your `NotebookParser` and then read it in the transformer:

```python
# NotebookParser:630: Set a attribute we can test against in the transform
document['parsed_nbsphinx'] = True

# CreateSectionLabels Transformer
nbsphinx_parsed = 'parsed_nbsphinx' in self.document
```

also, you can contain your notebook and cells inside a custom node, and then traverse to find that node, and modify all the sections below it.

```python
# Instead of this
for section in self.document.traverse(docutils.nodes.section):

# You would do something like this:
notebook_nodes = self.document.traverse(my_custom_notebook_node)
for notebook in notebook_nodes:
    for section in notebook.traverse(docutils.nodes.section):
        # Only modify sections that come from notebooks
```

Then in your `Nbinput` and `Nboutput` wrap the parsed data in a `def my_custom_notebook_node(nodes.container)`.

Now all the notebooks data are contained within `my_custom_notebook_node` which you can traverse for and manipulate however.

You might like `docutils.nodes.make_id` if you want id's normalized like sphinx.

mgeier:
Thanks @clenk and @lakhman!

@lakhman I've kept the `'.ipynb'` check for a quick fix, I'll look into your suggestions later.

PR #150, opened 2017-12-13
Add nbsphinx_prolog/nbsphinx_epilog

mgeier:
This can be used as a solution for #5, but also for many other things.


hagenw:
I think it is a good idea to have the possibility to allow for user content similar to the LaTeX prolog and epilog. I tested the documentation locally and all links where working.

I would say it is a temporary solution for #5, but a little bit complicated for a normal user. In the long run it would be nice to have just a config option to enable or disable the links at the top of the page. I guess, in addition you would need an option to set the binder url. Do you also need one for specifying the github url or can this be automatically extracted?

mgeier:
> I would say it is a temporary solution for #5, but a little bit complicated for a normal user.

I've added a few simple examples to help "normal" users: 48ab715f63c84cd7c5861e24ec9930c6c157406b.
I hope that makes it less complicated.

> In the long run it would be nice to have just a config option to enable or disable the links at the top of the page.

Users can put there whatever they want, I don't want to limit them with my choice of words and formatting. I guess the appropriate formatting also depends on the used HTML theme.
I also don't want to limit this to Github and Binder, there might be many other services/URLs that people might want to use.

A minor technical problem is that the path of the source directory within the repository (in our case `doc/`) is unknown to Sphinx, if I'm not mistaken.

> Do you also need one for specifying the github url or can this be automatically extracted?

I think readthedocs.org tries to extract that information, but I see no reason why I should burden myself with that. This is something the user can provide easily.

BTW, the "Edit on Github" link on readthedocs.org is supposed to point to the same Github page as the link I'm generating in the `nbsphinx_prolog` of the documentation. However, this still doesn't currently work for branches, not even for normal RST files, see e.g. http://nbsphinx.readthedocs.io/en/rtd-theme/a-normal-rst-file.html.

hagenw:
The example should indeed be a very good hint how to add a binder link on your own page.
Maintaining prolog and epilog should then be enough.

PR #158, opened 2018-01-09
DOC: Mention the "raises-exception" cell tag

### LINKED ISSUE #125: Ignore errors in a single cell
Is there currently a way to ignore errors produced by a single cell, or can it only be turned on globally or by notebook? If this isn't possible, it would be a nice feature to have - now that cells have their own metadata, it could be as simple as a check for:
```
"nbsphinx": {
  "allow_errors": true
},
```
at the cell-level metadata.

mgeier:
This should work starting with `nbconvert` version 5.3, which is not yet released.

See https://github.com/jupyter/nbconvert/pull/684.

* [x] depending on the outcome of https://github.com/jupyter/nbconvert/issues/730, the behavior of *not* raising an exception in a cell marked with `raises-exception` should be documented.

Closes #125.

### DIFF
--- a/doc/allow-errors-per-cell.ipynb
+++ b/doc/allow-errors-per-cell.ipynb
@@ -51,7 +51,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The following code cell is executed although the previous cell raised an exception."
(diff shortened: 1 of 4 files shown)


mgeier:
This is only available in `nbconvert` >= 5.4.

Issue #155, opened 2018-01-06
Centered equations not rendering properly


rsokl:
When trying to reproduce the centered equation from your documentation (using the latest version of `nbsphinx` from pypi):

***

![eqn0](https://user-images.githubusercontent.com/29104956/34644157-2ef39134-f2ff-11e7-8f26-dbe7cc3f40f1.PNG)

*** 
I find that the centered equation does not render properly:

***

![eqn](https://user-images.githubusercontent.com/29104956/34644148-09ea9c0c-f2ff-11e7-9f7d-3da29cb24dfe.PNG)

***

I used the exact content of your notebook cell:
```
## Equations

Equations can be formatted really nicely, either inline, like $\text{e}^{i\pi} = -1$, or on a separate line, like

\begin{equation}
\int_{-\infty}^\infty f(x) \delta(x - x_0) dx = f(x_0)
\end{equation}

*Note:* Avoid leading and trailing spaces around math expressions, otherwise errors like the following will occur when Sphinx is running:

    ERROR: Unknown interpreted text role "raw-latex".

See also the [pandoc docs](http://pandoc.org/MANUAL.html#math):

> Anything between two `$` characters will be treated as TeX math. The opening `$` must have a non-space character immediately to its right, while the closing `$` must have a non-space character immediately to its left, and must not be followed immediately by a digit.
```

mgeier:
What Sphinx version are you using?

This looks like a bug from versions < 1.4.2, see #44 and https://github.com/sphinx-doc/sphinx/pull/2473 and https://github.com/sphinx-doc/sphinx/pull/2476.

LLrsokl:
I'm running Sphinx v1.6.4

mgeier:
Hmmm, that's strange.

I don't know what exactly is going wrong here, but probably having a look at the intermediary RST file gives us a clue. Can you please check out the `save-rst` branch and look in the generated file in the HTML output directory (I guess it'll be called `markdown.rst`) and have a look how the relevant `math` directive looks?

BTW, I've given the wrong reference #44 above. For Markdown cells, the "nowrap" thing was introduced in #43 and updated in #110.

rsokl:
Sorry, I'm not quite following. I checked out save-rst, navigated to the docs directory, and ran `python -m sphinx . _build -j4`.

There are not intermediate .rst in the _build dir. The html for the centered equation is:

```html
[...]

PR #160, opened 2018-01-12
Use universal_newlines=True when communicating with pandoc

mgeier:
This is work in progress and may not be used at all.

See #155.


mgeier:
This is not needed because the problem was solved with #161.

Issue #174, opened 2018-03-29
Commit "Rewrite local links for RST" emits "WARNING: file not found"


jschueller:
Latest release 0.3.2 breaks the world for us as the last commit https://github.com/spatialaudio/nbsphinx/commit/7e791655205beebccce65b920d4cf798c8f6d614 breaks a link to a plot directive script (here empirical_cdf-1.py):

```
Warning, treated as error
/tmp/openturns/python/src/sphinx_build/theory/data_analysis/empirical_cdf.rst.rst:file not found: 'theory/data_analysis/empirical_cdf-1.py'
```

Here is the rst file, it contains a plot directive and as well a link to a notebook:
https://github.com/openturns/openturns/blob/master/python/doc/theory/data_analysis/empirical_cdf.rst

Here is the rendered version, before the 0.3.2 update:
http://openturns.github.io/openturns/master/theory/data_analysis/empirical_cdf.html

mgeier:
So you are saying everything works perfectly but there is a warning?

jschueller:
No, the renderend version was built with 0.3.1, before the 0.3.2 update. We treat sphinx warnings as errors. If I believe the message emited, the pyplot script may be missing in the html.

mgeier:
And what happens if you allow warnings?

jschueller:
If I allow warnings, the build succeeds.
The problem is that it emits a huge number of these, and that prevents us from catching the other useful warnings.

I prepared a very small example with the use of nbsphinx with a plot directive: https://github.com/jschueller/foo, see the travis log at: https://travis-ci.org/jschueller/foo.

We see that it emits 4 warnings on a single plot directive. We have lots of these...

jschueller:
Note that after some checking the link to the script is not broken as I suggested, issue title changed.

mgeier:
Thanks for confirming that the warning is the only problem, now we can try to solve this.

Ideas:

* just remove the warning
* create a new config option to disable this specific warning
* modify all warnings to be able to use `supress_warnings`, already discussed in #130 

Any more ideas?

Any comments?

jschueller:
I'd prefer to actually fix the warning in nbsphinx. Do you have a rough idea of what could cause it in that commit I pointed out ?

mgeier:
I know exactly what causes it:

[...]

PR #177, opened 2018-04-13
Add an option to disable link checks

jschueller:
tries to adress #130, see also discussion in #174


mgeier:
Thanks for this PR!
I've created an alternative solution for #174, see #178.

PR #178, opened 2018-04-17
Copy locally linked files only for Jupyter notebooks

mgeier:
... while still re-writing local links to Sphinx source files in all
Sphinx source files (not only notebooks).

See also #174.

### DIFF
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -75,19 +75,6 @@ For example, see this subsection_.
     .. _subsection: subdir/a-notebook-in-a-subdir.ipynb#A-Sub-Section
 
 
-Links to Local Files (HTML only)
---------------------------------
-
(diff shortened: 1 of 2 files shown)


mgeier:
A bit more explanation:

I split up `ProcessLocalLinks` into two parts: `RewriteLocalLinks` and `CopyLinkedFiles`.
This way, I can apply the former to all Sphinx source files, while the latter is only applied to Jupyter notebooks.

jschueller:
Is a release planned after this gets merged ?

mgeier:
Thanks @jschueller for testing!

> Is a release planned after this gets merged ?

Yes, definitely. I also want to merge #179 and then I want to wait for a few days for people to test it in the `master` branch.
If you see more than a week of inactivity, please remind me ...

jschueller:
Ok, thanks.

PR #185, opened 2018-04-27
Add options for inputs, outputs and prompts to be excluded

### LINKED ISSUE #131: Use of nbconvert TemplateExporter to exclude input/output and/or the related in/out prompts
In order to use this **for every notebook at once**  I'm still trying some things out (see [#15)](https://github.com/spatialaudio/nbsphinx/issues/15) for my earlier comments)

As suggested there, I'v added some nbsphinx options and used the  global setting in conf.py 

```
def setup(app):
  ...
    #hide or display input, output,....
    app.add_config_value('nbsphinx_exclude_output', False, rebuild='env')
    app.add_config_value('nbsphinx_exclude_input', False, rebuild='env')
    app.add_config_value('nbsphinx_exclude_input_prompt', False, rebuild='env')
    app.add_config_value('nbsphinx_exclude_output_prompt', False, rebuild='env')
    app.add_config_value('nbsphinx_exclude_markdown', False, rebuild='env')
    app.add_config_value('nbsphinx_exclude_code_cell', False, rebuild='env')
```
which is used then in the config setting of  **Exporter(nbconvert.RSTExporter)**

```
class Exporter(nbconvert.RSTExporter):
...
    def __init__(self, execute='auto', kernel_name='', execute_arguments=[],
                 allow_errors=False, timeout=30, codecell_lexer='none',
                 exclude_output = False, exclude_input = False, 
                 exclude_input_prompt = False, exclude_output_prompt = False, 
                 exclude_markdown = False, exclude_code_cell = False):
       ...
        self._exclude_output = exclude_output
        self._exclude_input = exclude_input
        self._exclude_input_prompt = exclude_input_prompt
        self._exclude_output_prompt = exclude_output_prompt
        self._exclude_markdown = exclude_markdown
        self._exclude_code_cell = exclude_code_cell 
        
        super(Exporter, self).__init__(
            template_file='nbsphinx-rst.tpl', extra_loaders=[loader],
            config=traitlets.config.Config(
                {'HighlightMagicsPreprocessor': {'enabled': True},
                'TemplateExporter':{
                "exclude_output": exclude_output,
                "exclude_input": exclude_input,
                "exclude_input_prompt": exclude_input_prompt,
                "exclude_output_prompt": exclude_output_prompt,
                "exclude_markdown": exclude_markdown,
                "exclude_code_cell": exclude_code_cell}}),
            filters={
                'convert_pandoc': convert_pandoc,
                'markdown2rst': markdown2rst,
                'get_empty_lines': _get_empty_lines,
                'extract_toctree': _extract_toctree,
                'get_output_type': _get_output_type,
                'json_dumps': json.dumps,
            })
```

This works fine but not for every output prompt. 
I tried to solve it but apparently in **class NbOutput(rst.Directive)** I don't seem to be able to get hold of the "exclude_output_prompt" setting.
With a global variable I can tackle this but still trying to figure out how to do it correctly without using this 'dirty hack':

```
class NbOutput(rst.Directive):
 ...
    def run(self):
           ....
            global nbsphinx_args_output_prompt
            if nbsphinx_args_output_prompt:
                text = ''
            else:
                text = 'Out [{}]:'.format(execution_count)
```

Any Idea how I can pass **env.config.nbsphinx_exclude_output_prompt** to the **run** method of the class **NbOutput**?

DavidPowell:
Resolves #131. Code was contributed by @keluc.

Currently no changes have been made to the docs to reflect the additional options.

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -46,6 +46,7 @@ import traitlets
 
 _ipynbversion = 4
 
+
 # See nbconvert/exporters/html.py:
 DISPLAY_DATA_PRIORITY_HTML = (
     'application/javascript',
@@ -520,7 +521,11 @@ class Exporter(nbconvert.RSTExporter):
     """
 
     def __init__(self, execute='auto', kernel_name='', execute_arguments=[],
-                 allow_errors=False, timeout=30, codecell_lexer='none'):
+                 allow_errors=False, timeout=30, codecell_lexer='none',
+                 exclude_output = False, exclude_input = False, 
+                 exclude_input_prompt = False, exclude_output_prompt = False, 
+                 exclude_markdown = False, exclude_code_cell = False):
+                 
         """Initialize the Exporter."""
         self._execute = execute
         self._kernel_name = kernel_name
@@ -529,10 +534,24 @@ class Exporter(nbconvert.RSTExporter):
         self._timeout = timeout
         self._codecell_lexer = codecell_lexer
         loader = jinja2.DictLoader({'nbsphinx-rst.tpl': RST_TEMPLATE})
+        self._exclude_output = exclude_output
+        self._exclude_input = exclude_input
+        self._exclude_input_prompt = exclude_input_prompt
+        self._exclude_output_prompt = exclude_output_prompt
+        self._exclude_markdown = exclude_markdown
+        self._exclude_code_cell = exclude_code_cell 
+        
         super(Exporter, self).__init__(
             template_file='nbsphinx-rst.tpl', extra_loaders=[loader],
             config=traitlets.config.Config(
-                {'HighlightMagicsPreprocessor': {'enabled': True}}),
+                {'HighlightMagicsPreprocessor': {'enabled': True},
+                'TemplateExporter':{
+                "exclude_output": exclude_output,
+                "exclude_input": exclude_input,
+                "exclude_input_prompt": exclude_input_prompt,
+                "exclude_output_prompt": exclude_output_prompt,
+                "exclude_markdown": exclude_markdown,
+                "exclude_code_cell": exclude_code_cell}}),
             filters={
                 'convert_pandoc': convert_pandoc,
                 'markdown2rst': markdown2rst,
@@ -647,6 +666,13 @@ class NotebookParser(rst.Parser):
             allow_errors=env.config.nbsphinx_allow_errors,
             timeout=env.config.nbsphinx_timeout,
             codecell_lexer=env.config.nbsphinx_codecell_lexer,
+            
+            exclude_output = env.config.nbsphinx_exclude_output, 
+            exclude_input = env.config.nbsphinx_exclude_input, 
+            exclude_input_prompt = env.config.nbsphinx_exclude_input_prompt, 
+            exclude_output_prompt = env.config.nbsphinx_exclude_output_prompt, 
+            exclude_markdown = env.config.nbsphinx_exclude_markdown, 
+            exclude_code_cell = env.config.nbsphinx_exclude_code_cell
         )
 
         try:
@@ -778,7 +804,10 @@ class NbOutput(rst.Directive):
 
         # Optional output prompt
         if execution_count:
-            text = 'Out[{}]:'.format(execution_count)
+            if self.state.document.settings.env.config.nbsphinx_exclude_output_prompt:
+                text = ''
+            else:
+                text = 'Out[{}]:'.format(execution_count)
             container += CodeNode.create(text, classes=['prompt'])
             latex_prompt = text + ' '
         else:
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR!

Is `exclude_code_cell` really ever needed?
And even if that's the case, wouldn't a combination of `exclude_input` and `exclude_output` do the same thing?

Is `exclude_markdown` really needed?
What would be a use case for that?

DavidPowell:
For my particular case, I only use the `nbsphinx_exclude_input` and `nbsphinx_exclude_output_prompt` options. I could also envisage that `nbshinx_exclude_input_prompt` might be useful.

In all other cases (hiding entire code or markdown cells or hiding input) if I ever wanted to do this, it would be on a per-cell basis by creating tags, and using the new tag based element filtering feature of nbconvert.

Unless anyone can point to a use-case that requires other options, I think it best to go with just the 3 options mentioned above.

Should I just add another commit to remove unwanted options, I would you prefer that I rebased and squashed to remove them from the history?

rddaz2013:
That means the "hide_cell" option  is not a function of nbsphinx - nbconvert/sphinx is doing the job?

I had implemented it via the templates. 
https://github.com/spatialaudio/nbsphinx/compare/master...rddaz2013:master

DavidPowell:
@rddaz2013 I take it that your preferred option is to hide individual input cells, based on metadata created by the Jupyter nbextension `hide_input`. The current implementation gives an option to show or hide all cells, which is good enough for my purposes. Does your use case require hiding only specific inputs cells?

In my view, hiding individual inputs and outputs would be better done with nbconvert's new feature, [tag based element filtering](http://nbconvert.readthedocs.io/en/latest/changelog.html#tag-based-element-filtering). This would avoid the need for any custom templates, doesn't require any extensions and the same approach can be used in other contexts, e.g. using nbconvert to directly generate html.

mgeier:
@DavidPowell I agree, can you please just add the options you need? If somebody else needs more, we can add them later.

> would you prefer that I rebased and squashed to remove them from the history

Normally I would prefer a rebase+squash, but in this case I think it would be good to keep @keluc's commits, because he first took the initiative.
But I guess you could still re-base without losing the original authorship.

DavidPowell:
Unwanted options have been removed. As I didn't get any feedback from @rddaz2013 on what he means by hiding cells, nothing has been done on that front.

mgeier:
`nbsphinx_exclude_input_prompt` doesn't seem to work.

`nbsphinx_exclude_output_prompt` removes the output prompt but an awkward space remains (in LaTeX it seems to be the width of one character, in HTML it is the whole prompt width).
Is this intentional?

tshead2:
From what I can see here, `exclude_input_prompt` isn't used by the new code.  I'm chiming in because I write lots of library documentation using notebooks, and the input prompts are meaningless distractions in that case.  See http://toyplot.readthedocs.io/en/stable/tutorial.html for an example of a notebook without prompts on RTD (generated with my own home-grown code, which I'd like to eliminate in favor of nbsphinx).

keluc:
I've taken a further look, because in the standalone use of  nbconvert it all seems to work (exclude_input_prompt etc.), but in this context this doesn't always work. 
- I already had an issue with the exclude_output_prompt which wasn't always excluded (not further investigated). I handled this by amending the run method in the class NbOutput, [see my comments in the beginning of this discussion](https://github.com/spatialaudio/nbsphinx/issues/131) . I think we have the same problem in NbInput.run method. We can apply the same approach.
- With respect to the awkward space, it seems to me that it also can be handled in the same piece of code of both run methods, by only adding to the container variable if we want to show the prompts. 
- With respect to the latex issue, in the same run method, there is a line of code 
`latex_prompt = text + ' '` which may cause this problem (not checked)

To me, the reason why we can't use the options of nbconvert to it's full potential has to do with this approach (NbInput.run and NbOutput.run). The following resolves both issues (latex issue still to test)

```
class NbInput(rst.Directive):
 ...
    def run(self):
    ...
        # Input prompt
        if self.state.document.settings.env.config.nbsphinx_exclude_input_prompt:
            text = ''
            latex_prompt = ''
        else:        
            text = 'In [{}]:'.format(execution_count if execution_count else ' ')
            container += CodeNode.create(text, classes=['prompt'])
            latex_prompt = text + ' '
...
```

and 

```
class NbOutput(rst.Directive):
....
    def run(self):
    ....
        # Optional output prompt
        if execution_count:
            if self.state.document.settings.env.config.nbsphinx_exclude_output_prompt:
                text = ''
                latex_prompt = ''
            else:
                text = 'Out[{}]:'.format(execution_count)
                container += CodeNode.create(text, classes=['prompt'])
                latex_prompt = text + ' '
        else:
            # Empty container for HTML:
            container += rst.nodes.container(classes=['prompt', 'empty'])
            latex_prompt = ''
...
```

mgeier:
@tshead2 Thanks for your input! It's really helpful that you show an actual example of what you want to achieve. I don't really see a problem in showing the prompts, though. You are telling your readers that you are using notebooks (and you even urge them to play around with them), so I don't see why it should be helpful to try to hide the fact that your code snippets are actually Jupyter notebook cells.

Having said that, you can actually hide the prompts in the HTML output by using:

    nbsphinx_prompt_width = 0

Since that only does some CSS manipulations, it doesn't have any impact on the LaTeX output.

tshead2:
@mgeier - Happy to butt-in!  When it comes to running notebooks, I appreciate the prompts for keeping track of the order in which I've (re)run cells.  For documentation, where the order is always linear, they may be confusing for non-notebook users, provide no useful information for notebook experts, consume space, and add visual noise.  Anyway, that's good news about `nbsphinx_prompt_width` ... I don't see it in the documentation, where should I be looking for that?

While I'm at it, I'll also put in a good word for *occasionally* hiding cell inputs.  I sometimes need to generate a plot to illustrate an idea, and I want the reader's focus to be on the plot, not the code used to create it.

Cheers,
Tim

mgeier:
@tshead2 

> I don't see it in the documentation, where should I be looking for that?

The most reliable source is the source.

All options are (and future options will be) visible in the `setup()` function near the bottom of the file `src/nbshinx.py`:

https://github.com/spatialaudio/nbsphinx/blob/20621a39d237ba5791ac75f1d68c78de5b01892a/src/nbsphinx.py#L1438-L1448

All of these options (unless I missed some) are also mentioned in the file `doc/conf.py`:

https://github.com/spatialaudio/nbsphinx/blob/20621a39d237ba5791ac75f1d68c78de5b01892a/doc/conf.py#L22-L82

Most of the options are also mentioned in the documentation notebooks, but for some of them I was too lazy to mention them.

If you want to help out, please make a PR (or multiple) with improvements!

OverLordGoldDragon:
Needed feature, any plans to merge?

PR #197, opened 2018-07-17
Use add_source_suffix() if available

mgeier:
As suggested by @tk0miya in
https://github.com/sphinx-doc/sphinx/issues/5164#issuecomment-405092954

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -596,7 +596,7 @@ class NotebookParser(rst.Parser):
 
     """
 
-    supported = ()
+    supported = 'jupyter_notebook',
 
     def get_transforms(self):
         """List of transforms for documents parsed by this parser."""
@@ -1448,7 +1448,12 @@ def _add_notebook_parser(app):
 
 def setup(app):
     """Initialize Sphinx extension."""
-    _add_notebook_parser(app)
+    try:
+        # Available since Sphinx 1.8:
+        app.add_source_suffix('.ipynb', 'jupyter_notebook')
+        app.add_source_parser(NotebookParser)
+    except AttributeError:
+        _add_notebook_parser(app)
 
     app.add_config_value('nbsphinx_execute', 'auto', rebuild='env')
     app.add_config_value('nbsphinx_kernel_name', '', rebuild='env')


src/nbsphinx.py
```diff
@@ -1437,7 +1437,13 @@ def _add_notebook_parser(app):
 
 def setup(app):
     """Initialize Sphinx extension."""
-    _add_notebook_parser(app)
+    try:
+        # Available since Sphinx 1.8:
+        app.add_source_suffix('.ipynb', 'jupyter_notebook')
+        # Available since Sphinx 1.4:
+        app.add_source_parser(NotebookParser)
```
tk0miya: At present (including Sphinx-1.7), `app.add_source_parser()` takes two arguments; `suffix` and `parser`. This is new styled one will be introduced at 1.8.
In this PR, you try to call `app.add_source_suffix()` API first. As a result, that confirms current Sphinx is newer than 1.8. So this calling does not harm anybody.
But I'm worry about the comment above confuses somebody.

src/nbsphinx.py
```diff
@@ -596,7 +596,7 @@ class NotebookParser(rst.Parser):
 
     """
 
-    supported = ()
+    supported = 'jupyter_notebook',
```
tk0miya: Is there any formal name for `.ipynb` file?

src/nbsphinx.py
mgeier: Thanks @tk0miya, you are right, this might be confusing.

I've removed the confusing comment in ac837480ce6da5a42032316a710fd9c2d24dac3f.

src/nbsphinx.py
mgeier: I don't know, I was just taking your suggestion.

How did you choose the "format names" of the other existing formats?

I don't know of an official "type name" for Jupyter notebooks, the only thing that comes to mind is using the MIME type. According to https://jupyter.readthedocs.io/en/latest/reference/mimetype.html, this is

    application/x-ipynb+json

I was thinking about using this, but the other existing types (most notably `'restructuredtext'` and `'markdown'`) are clearly not MIME types, so it didn't seem reasonable to me to use the MIME type for Jupyter notebooks.

tk0miya: TBH, I've never give a name to existing formats. So I don't know what is good naming for `supported` attribute.
Anyway, Sphinx don't mind the name is proper or not. So I think this is okay.

mgeier:
Thanks @tk0miya!

PR #201, opened 2018-09-05
strip only newlines instead of all whitespace

gdementen:
I think it was always the intention to work like that and this bug is just a typo... I could be wrong obviously...

Attached is a small test case. It works fine after the patch.
[leading_blanks_testcase.zip](https://github.com/spatialaudio/nbsphinx/files/2352901/leading_blanks_testcase.zip)

Btw, nbconvert works correctly on this.

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -131,7 +131,7 @@ RST_TEMPLATE = """
 {%- if datatype == 'text/plain' -%}
 {{ insert_empty_lines(output.data[datatype]) }}
 
-{{ output.data[datatype].strip(\n) | indent }}
+{{ output.data[datatype].strip('\n') | indent }}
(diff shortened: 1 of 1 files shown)


gdementen:
the failure https://travis-ci.org/spatialaudio/nbsphinx/jobs/424793148#L890 does not seem to be related to my change (but rather to a dead link)

mgeier:
Thanks, this was indeed a typo!

The second example in your test case already worked because it uses a different code path.
The `datatype` for "standard output" and "standard error" is called `'ansi'`, because it can be colored with ANSI escape sequences:

https://github.com/spatialaudio/nbsphinx/blob/448f00b98c7382dcd1bf5b3e322851ac62012cd9/src/nbsphinx.py#L166-L185

And thanks for reminding me of this dead link.
I wasn't sure if the failure is temporary, but it has been broken for quite some time now ...

Issue #202, opened 2018-09-07
'NoneType' object has no attribute 'argv'  error with clean notebook


nlgranger:
On a clean notebook with `nbsphinx_execute = 'auto'`, running the notebook fails with the following error:

```
python -m sphinx -b html docs docs/_build                                                                                                                                                      
Running Sphinx v1.7.9
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: 0 added, 1 changed, 0 removed
reading sources... [100%] examples/minibatch_iterator                                                                                                                                                            
Notebook error:
AttributeError in examples/minibatch_iterator.ipynb:
'NoneType' object has no attribute 'argv'
(venv3) 
```

I'm trying to use nbsphinx on https://github.com/nlgranger/SeqTools (improved_doc branch) on a clean python 3 virtual env with the packages IPython, ipykernel, nbsphinx, Pillow and the project itself.

mgeier:
Is it possible that this is just an error in your notebook?
Can you please try to provide a minimal example notebook (with no additional dependencies) that shows this error?

nlgranger:
Sure, here is a self contained mini project, I'm trying to build the doc with `python -m sphinx docs docs/_build`: [test.zip](https://github.com/spatialaudio/nbsphinx/files/2363196/test.zip)

nlgranger:
If you mean `jupyter nbconvert --execute` it works without issues.

nlgranger:
I also have the error with a python 3.5 conda environment, can't make much sense out of it...

Juanlu001:
Can this be related to nbconvert 5.4 and https://github.com/jupyter/nbconvert/issues/878?

mgeier:
@Juanlu001 Yes it is.

mgeier:
@nlgranger Thanks for checking!

> If you mean `jupyter nbconvert --execute` it works without issues.

I meant `python3 -m nbconvert --execute`, but that does the same thing.

mgeier:
This has been fixed in https://github.com/jupyter/nbconvert/pull/924.

Now, we just have to wait for a new `nbconvert` release (most likely version 5.5).

PR #208, opened 2018-09-11
Don't pass an empty kernel name by default to ExecutePreprocessor

mpacer:
This uses the preferred pattern as described in https://github.com/jupyter/nbconvert/pull/886 for getting access to the default behaviour of deferring to the notebook's metadata by not passing in a `kernel_name` at all (rather than by passing in an empty kernel name).

However, https://github.com/jupyter/nbconvert/pull/886 should make the code today work.

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -573,10 +573,12 @@ class Exporter(nbconvert.RSTExporter):
             allow_errors = nbsphinx_metadata.get(
                 'allow_errors', self._allow_errors)
             timeout = nbsphinx_metadata.get('timeout', self._timeout)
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks @mpacer, but this is really ugly and un-Pythonic!

I've made a counter-suggestion in https://github.com/jupyter/nbconvert/pull/886#issuecomment-420734298.
That would make your code much simpler and easier to understand and I could keep my code unchanged.

mgeier:
This is not necessary anymore, since #202 (and https://github.com/jupyter/nbconvert/issues/878) was fixed by https://github.com/jupyter/nbconvert/pull/924.

Thanks anyway @mpacer!

PR #218, opened 2018-10-03
LaTeX: re-design code cells

mgeier:
Inspired by https://github.com/t-makaro/nb_pdf_template, I've tried to make the LaTeX version of the code cells a bit prettier. The colors are taken from JupyterLab.

As a positive side effect, this now uses the `sphinxVerbatim` including its line-break and page-break features!

A temporary example PDF file can be found there: https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-codecells/ (once this PR is merged this will be in https://readthedocs.org/projects/nbsphinx/downloads/pdf/latest/).

There are still a few rough edges, I'd love to hear some ideas for improvements!

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -117,6 +117,7 @@ RST_TEMPLATE = """
 {% macro insert_nboutput(datatype, output, cell) -%}
 .. nboutput::
 {%- if datatype == 'text/plain' %}{# nothing #}
+{%- elif datatype == 'ansi' %} ansi
 {%- else %} rst
 {%- endif %}
 {%- if output.output_type == 'execute_result' and cell.execution_count %}
@@ -176,9 +177,18 @@ RST_TEMPLATE = """
     .. raw:: latex
 
         %
-        \\begin{OriginalVerbatim}[commandchars=\\\\\\{\\}]
+        {
+        \\kern-\\sphinxverbatimsmallskipamount\\kern-\\baselineskip
+        \\kern+\\FrameHeightAdjust\\kern-\\fboxrule
+        \\vspace{\\nbsphinxcodecellspacing}
+        \\sphinxsetup{VerbatimBorderColor={named}{nbsphinx-code-border}}
+    {%- if output.name == 'stderr' %}
+        \\sphinxsetup{VerbatimColor={named}{nbsphinx-stderr}}
+    {%- endif %}
+        \\begin{sphinxVerbatim}[commandchars=\\\\\\{\\}]
 {{ output.data[datatype] | escape_latex | ansi2latex | indent | indent }}
-        \\end{OriginalVerbatim}
+        \\end{sphinxVerbatim}
+        }
         % The following \\relax is needed to avoid problems with adjacent ANSI
         % cells and some other stuff (e.g. bullet lists) following ANSI cells.
         % See https://github.com/sphinx-doc/sphinx/issues/3594
@@ -264,9 +274,12 @@ RST_TEMPLATE = """
 
 
 LATEX_PREAMBLE = r"""
-% Jupyter Notebook prompt colors
-\definecolor{nbsphinxin}{HTML}{303F9F}
-\definecolor{nbsphinxout}{HTML}{D84315}
+% Jupyter Notebook code cell colors
+\definecolor{nbsphinxin}{HTML}{307FC1}
+\definecolor{nbsphinxout}{HTML}{BF5B3D}
+\definecolor{nbsphinx-code-bg}{HTML}{F5F5F5}
+\definecolor{nbsphinx-code-border}{HTML}{E0E0E0}
+\definecolor{nbsphinx-stderr}{HTML}{FFDDDD}
 % ANSI colors for output streams and traceback highlighting
 \definecolor{ansi-black}{HTML}{3E424D}
 \definecolor{ansi-black-intense}{HTML}{282C36}
@@ -295,6 +308,27 @@ LATEX_PREAMBLE = r"""
 \newenvironment{notice}{\begin{sphinxadmonition}}{\end{sphinxadmonition}}%
 }{}
 \makeatother
+
+% Define an environment for non-plain-text code cell outputs (e.g. images)
+\makeatletter
+\newenvironment{nbsphinxfancyoutput}{%
+\def\nbsphinxfcolorbox{\spx@fcolorbox{nbsphinx-code-border}{white}}%
+\def\FrameCommand{\nbsphinxfcolorbox\nbsphinxfancyaddprompt\@empty}%
+\def\FirstFrameCommand{\nbsphinxfcolorbox\nbsphinxfancyaddprompt\sphinxVerbatim@Continues}%
+\def\MidFrameCommand{\nbsphinxfcolorbox\sphinxVerbatim@Continued\sphinxVerbatim@Continues}%
+\def\LastFrameCommand{\nbsphinxfcolorbox\sphinxVerbatim@Continued\@empty}%
+\MakeFramed{\advance\hsize-\width\@totalleftmargin\z@\linewidth\hsize\@setminipage}%
+}{\par\unskip\@minipagefalse\endMakeFramed}
+\makeatother
+\newbox\nbsphinxpromptbox
+\def\nbsphinxfancyaddprompt{\ifvoid\nbsphinxpromptbox\else
+    \kern\fboxrule\kern\fboxsep
+    \copy\nbsphinxpromptbox
+    \kern-\ht\nbsphinxpromptbox\kern-\dp\nbsphinxpromptbox
+    \kern-\fboxsep\kern-\fboxrule\nointerlineskip
+    \fi}
+\newlength\nbsphinxcodecellspacing
+\setlength{\nbsphinxcodecellspacing}{0pt}
 """
 
 
@@ -755,7 +789,7 @@ def _create_code_nodes(directive):
         # 'class' can be 'stderr'
         inner_classes.append(directive.options.get('class', ''))
         prompt_template = config.nbsphinx_output_prompt
-        if directive.arguments and directive.arguments[0] == 'rst':
+        if directive.arguments and directive.arguments[0] in ['rst', 'ansi']:
             fancy_output = True
     else:
         assert False
@@ -775,7 +809,12 @@ def _create_code_nodes(directive):
         inner_node = docutils.nodes.container(classes=inner_classes)
         sphinx.util.nodes.nested_parse_with_titles(
             directive.state, directive.content, inner_node)
-        outer_node += FancyOutputNode('', inner_node, prompt=prompt)
+        if directive.arguments[0] == 'rst':
+            outer_node += FancyOutputNode('', inner_node, prompt=prompt)
+        elif directive.arguments[0] == 'ansi':
+            outer_node += inner_node
+        else:
+            assert False
     else:
         text = '\n'.join(directive.content.data)
         inner_node = docutils.nodes.literal_block(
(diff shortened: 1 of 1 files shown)


src/nbsphinx.py
```diff
@@ -177,9 +178,16 @@
     .. raw:: latex
 
         %
-        \\begin{OriginalVerbatim}[commandchars=\\\\\\{\\}]
+        {
+        \\vspace{-0.5\\baselineskip}
+        \\sphinxDeclareColorOption{VerbatimBorderColor}{ {named}{nbsphinx-code-border}}
```
jfbu: you probably want rather here

```
\\sphinxsetup{VerbatimBorderColor={named}{nbsphinx-code-border}}
```

src/nbsphinx.py
```diff
@@ -177,9 +178,16 @@
...
+    {%- if output.name == 'stderr' %}
+        \\sphinxDeclareColorOption{VerbatimColor}{ {named}{nbsphinx-stderr}}
```
jfbu: same remark as previous comment

src/nbsphinx.py
```diff
@@ -295,6 +306,18 @@
...
+\makeatletter
+\newenvironment{nbsphinxfancyoutput}{%
+\def\nbsphinxfcolorbox{\spx@fcolorbox{nbsphinx-code-border}{white}}%
+\def\FrameCommand{\nbsphinxfcolorbox\sphinxVerbatim@Before\sphinxVerbatim@After}%
+\def\FirstFrameCommand{\nbsphinxfcolorbox\sphinxVerbatim@Before\sphinxVerbatim@Continues}%
+\def\MidFrameCommand{\nbsphinxfcolorbox\sphinxVerbatim@Continued\sphinxVerbatim@Continues}%
+\def\LastFrameCommand{\nbsphinxfcolorbox\sphinxVerbatim@Continued\sphinxVerbatim@After}%
+\MakeFramed{\advance\hsize-\width\@totalleftmargin\z@\linewidth\hsize\@setminipage}%
+}{\par\unskip\@minipagefalse\endMakeFramed}
+\makeatother
```
jfbu: this looks sane enough...

src/nbsphinx.py
```diff
@@ -1342,73 +1370,66 @@ def visit_codearea_latex(self, node):
...
     out = []
     lines = ''.join(self.popbody()).split('\n')
     assert lines[0] == ''
     out.append(lines[0])
+    out.append('{')  # Start a scope for colors
+    prompt = node['prompt']
+    if 'nbinput' in node.parent['classes']:
+        promptcolor = 'nbsphinxin'
+        out.append(r'\sphinxDeclareColorOption{VerbatimColor}'
+                   '{{named}{nbsphinx-code-bg}}')
```
jfbu: I don't know where this is happens, but it is likely you want better here

```
\sphinxsetup{VerbatimColor={named}{nbsphinx-code-bg}}
```

src/nbsphinx.py
```diff
@@ -1342,73 +1370,66 @@ def visit_codearea_latex(self, node):
...
+    else:
+        # Reduce the space above output cells
+        out.append(r'\vspace{-0.5\baselineskip}')
+        promptcolor = 'nbsphinxout'
+    out.append(r'\sphinxDeclareColorOption{VerbatimBorderColor}'
+               '{{named}{nbsphinx-code-border}}')
```
jfbu: same remark as above. Will this be scoped by some LaTeX environment or braces `{...}`?

mgeier:
Thanks a lot for the review @jfbu! I hope you enjoyed watching the lord of the rings!

src/nbsphinx.py
mgeier: Thanks, I've changed that.

src/nbsphinx.py
mgeier: OK, good, I was a bit intimidated by all this fancy TeX code ...

src/nbsphinx.py
mgeier: Yes, in braces.

A few lines above is an

    out.append('{')

... and further down is an

    out.append('}')

I didn't know if scoping it like that works, but I just tried it and it seems to work fine.

src/nbsphinx.py
jfbu: what is the use case? you mention images, do you mean possibly more than one? because the whole point of MakeFramed is to offer some breakable framed box. If you don't need a breakable framed box, you may makedo with simply using directly `\fcolorbox` (using a minipage inside it with width `\linewidth`). However As  `\fcolorbox` includes in the width the added one from the frame, one would need some extra code (see below). The `\spx@fcolorbox` can not be used as is directly because due to its first and last code lines with `\@totalleftmargin` which are voodoo magic tailored for usage with `\MakeFramed`. I should have abstracted this into some other macro and included in `sphinx.sty` some `\sphinxfcolorbox` which could be used away from `\MakeFramed` context.

Anyway, perhaps you only need

```
\noindent{\fboxsep\sphinxverbatimsep\fboxrule\sphinxverbatimborder
    \kern-\fboxrule\kern-\fboxsep
    \fcolorbox{color-for-frame}{color-for-background}{%
    \begin{minipage}{\linewidth}% perhaps better \begin{minipage}[t]{\linewidth}%
    .... contents go here ...
    \end{minipage}}\kern-fboxsep\kern-\fboxrule}\par
```


The  `\sphinxVerbatim@Before`,  `\sphinxVerbatim@After` are empty, only inside `sphinxVerbatim` do they get configured to add a title, either before or after the framed contents. In your case as it stands you could as well use

```
\def\FrameCommand{\nbsphinxfcolorbox\@empty\@empty}%
etc... replace by \@empty instances of @Before, @After macros
```

(this is only for readability of your code, it would change nothing to efficiency because sphinx.sty has done `\let\sphinxVerbatim@Before\@empty` etc)

The `\sphinxVerbatim@Continues` and `\sphinxVerbatim@Continued` are for case with pagebreaks. You can renew locally in your environment the commands `\sphinxstylecodecontinues`, `\sphinxstylecodecontinued` (by default they use `\footnotesize`) and/or `\literalblockcontinuesname`, `\literalblockcontinuedname`. For example

```
\renewcommand\literalblockcontinuedname{(continuation of non-text code-cell from previous page)}
```

(just in case this sounds useful, but agaig I wonder if your non-text content has potential pagebreaks; if not the `\fcolorbox` as decorated above should be enough)

jfbu:
I enjoy it again each time I watch lord of the rings!

jfbu:
> I've tried your suggestion with |\fcolorbox| and it works nicely, but i would indeed like to have the page-breaking feature. Most of the time, the outputs will be single images, but it can also be things like math equation and even arbitrary Markdown content (which is then converted to arbitrary reST content).
> 


ah ok, then sure, you must go the `\MakeFramed` way. (in future Sphinx should provide tcolorbox usage, but one issue is that it has so many options it is out of question to document this as part of Sphinx, so users will be asked to get see by themselves tcolorbox documentation, which means only LaTeX savvy people can reasonably be expected to do that)

> There is one thing left that's bugging me: I've tried to position the output prompt of "fancy" blocks in |visit_fancyoutput_latex()|, using |\llap| and some manual |\vskip|. That's working, but I'm sure there is a better way, right?
> And there is one problem that occurs sometimes at page breaks: Sometimes the prompt number is still on the previous page, while the content of the output is entirely on the next page. You can see this in the latest PDF build <https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-codecells/>, e.g. at the end of page 17 and 19. It would be better if the whole frame including the prompt would move to the next page. Do you have an idea how to do that?
> 


I will look at this later.

Unfortunately I am stuck part of the day with some outdated Firefox (from Ubuntu 16.04, no admin rights) and since recently this means I can't use Github web interface at all.

I can read your commits but can't comment or review etc...

jfbu:
(make sure you click on the "show outdated" links; I accidentally reviewed your initial commit. sorry for mix up). In particular don't forget to replace the former `\sphinxVerbatim@Before` (now `\@empty`) by `\nbsphinxfancyaddprompt`.

jfbu:
oh sorry I think the prompt placement is not very good, possibly from the fact that I tested this with `sphinxVerbatim` (modified to use the proposed macro), but your direct use of `\MakeFramed` is lacking some parameter adjustments done in `sphinxVerbatim` environment. I need to check but again tonight lord of the rings (part 3)!...

mgeier:
Thanks @jfbu! I'm not at all getting a headache, I'm really enjoying this!

I've added your suggestions in 373aa02adf0f59593b308e019bb96ee3e6a0ee25.

I already thought about adding the prompts to the framed boxes (like titles are normally added), but I'm not even close to having the TeX skills to pull that off. But you can do it, that's so great!

The vertical placement is perfect with your suggestion and the problem at the page breaks doesn't occur anymore.
I think I'm fine with having the content of the box being treated like non-text. Sure, it would be perfect if normal text would be aligned at the base line (especially if the content is a single lowercase line), but I can live without it.

But there is a problem now (you can see it in [the current PDF output](https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-codecells/)): The horizontal alignment of the prompts is different from the one of the non-fancy code cells. Some are aligned w.r.t. the frame, the others are aligned w.r.t. the left edge of the text.
Is it feasible to use the same method of placing the prompts in the non-fancy cells (which use `sphinxVerbatim`)?
If not, how can this difference be compensated?

I also tried your suggestion for my previous version and added `\nobreak` right after `\llap{...}`, but it didn't change anything. The line breaks between prompts and box contents still happened.

src/nbsphinx.py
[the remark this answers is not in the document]
mgeier: Wow, now that is dark magic to me! I have some idea what most of the words mean, but I have no clue how this thing actually works. But it does work!

src/nbsphinx.py
```diff
@@ -1342,73 +1370,66 @@ def visit_codearea_latex(self, node):
...
+    out.append('}')  # End of scope for colors
     assert lines[-1] == ''
     out.append(lines[-1])
     self.body.append('\n'.join(out))
 
 
 def visit_fancyoutput_latex(self, node):
+    out = r"""
+\vspace{-0.3\baselineskip}
+\begin{nbsphinxfancyoutput}
```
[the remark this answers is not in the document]
mgeier: BTW, in the line above that, I've inserted this `\vspace{-0.4\baselineskip}`, which I found by trial and error.
It would be great to specify the exact space between input and output cells.
Do you happen to have an idea how to do that?

jfbu:
> But there is a problem now (you can see it in the current PDF output): The horizontal alignment of the prompts is different from the one of the non-fancy code cells. Some are aligned w.r.t. the frame, the others are aligned w.r.t. the left edge of the text.

oh sorry it is an oversight of my part. The code of `\nbsphinxfancyaddprompt` locates the prompt relative to the frame, so at an additional distance`\fboxsep+\fboxrule` to the left of contents. Use this:

```
\def\nbsphinxfancyaddprompt{\ifvoid\nbsphinxpromptbox\else
    \kern\fboxrule\kern\fboxsep
    \moveright\dimexpr\fboxsep+\fboxrule\relax\copy\nbsphinxpromptbox
    \kern-\ht\nbsphinxpromptbox\kern-\dp\nbsphinxpromptbox
    \kern-\fboxsep\kern-\fboxrule\nointerlineskip
    \fi}
```

and also don't use this (sorry I did not check closely, but now I tested that using `\Verb` (which fancyvrb provides) inside an argument does not work, so it makes no difference here to using  `\ttfamily`):

```
\savebox\nbsphinxpromptbox[0pt][r]{\textcolor{nbsphinxout}{\strut\Verb|%s |}}
```

but rather

```
\savebox\nbsphinxpromptbox[0pt][r]{\textcolor{nbsphinxout}{\strut\ttfamily %s\ }}
```

No fancy character `$` or `&` that LaTeX does not like can arise in the prompt right? else suffices to use `\detokenize{%s}` in place of `%s`.

About the `\ `: it provides a space in monospace font with suitable width whereas the ` ` would give a wider one due to something very strange at TeX level about `{\ttfamily [1:] }` versus `{\ttfamily [1:]\ }`. The former being wider than  the latter when `\nonfrenchspacing` is in place (which is default with English language). I had to debug this now, because I didn't expect it (besides I never see it as I am French hence my documents are in `\frenchspacing` mode automatically when using babel + french).

(regarding the prompt for code cells which you insert with the `\llap`, the space is made active inside the environment, and in the end it behaves like a `\ `, so the space at end of `\llap{\color{promptcolor}[1:] }` did not mysteriously give a double space in output.)

I am troubled by something. In your PDF I see no size difference, but the prompt for fancy cells is typeset using normal size, whereas I expected the prompt for code cell, which you insert via the `\llap` to use the font size for code cells, i.e. "small" by default when using pdflatex as tex engine (for xelatex, Sphinx uses other fonts by default and "small" is turned off as not being appropriate).

I don't currently understand why I see no size difference in your PDF output. (I should clone your project to check how it looks like in LaTeX file).

src/nbsphinx.py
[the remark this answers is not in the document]
jfbu: To be honest, I feel currently there is something I do not quite understand about where exactly the prompt ends up vertically. This code is used in interaction with framed.sty inner workings, and something is a bit weird in my testing. Perhaps later on I will get a better feel for it all.

src/nbsphinx.py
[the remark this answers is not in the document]
jfbu: This is a bit complicated area which involves framed parameter `\FrameHeightAdjust` as well as `\OuterFrameSep` (which defaults to `\topsep`). The Sphinx code for `sphinxVerbatim` uses a `\trivlist` to control vertical space but this is complicated. Today I am busy and can't investigate more (I need to see more closely the actual LaTeX code produced by nbsphinx) until perhaps the evening (European time).

src/nbsphinx.py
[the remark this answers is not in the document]
jfbu: arrgh, in fact I was lacking the famous `\strut` in my renewed testing, so now I understand again...

src/nbsphinx.py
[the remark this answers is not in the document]
jfbu: If I understand correctly the `nbsphinxfancyoutput` environment is always used following a `sphinxVerbatim`? 

If this the case you can control exactly the spacing between the bottom frame of sphinxVerbatim and the top one of fancycell by inserting a line in-between in LaTeX source:

```
\hrule height 0pt\relax
```

Then you can use `\vspace{spacing you want}` before the `\begin{nbsphinxfancyoutput}`.

I hope you don't want to control rather the distance from last baseline in sphinxVerbatim to first baseline in nbsphinxfancyoutput... because this is always same problem that in nbsphinxfancyoutput, the top frame is at distance of `\fboxsep` from the top of highest glyph on first line (if text line), thus `ppp` is not like `AAA`. But adding a `\strut` systematically at start of contents (which would control that max height) is problematic because contents maybe a list etc... (this is the very same problem which means locating the prompt box is not easy, because baseline of first text line depends on text line; and there might not be a first text line, it could be some display math etc..., so we simply position the prompt box in order to be perfectly at correct height if first text line had a `\strut`).

mgeier:
Thanks, that's great, now (with e03acf043abac8368d1c2104b391faf3aefd485f) the horizontal position is also perfect!

I've also added `\detokenize`, just to be on the safe side. I recently made the prompts configurable, so a user could really use anything.

> using `\Verb` (which fancyvrb provides) inside an argument does not work

Hmmm. It seems to work for me.
If I use `\ttfamily` instead, the fonts don't match! The one with `\ttfamily` is indeed slightly bigger.

I was just naively hoping that `\Verb` would somehow use the same settings as the `Verbatim` environment (and therefore also `sphinxVerbatim`). And it seemed to work ...

src/nbsphinx.py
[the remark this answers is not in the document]
mgeier: You are the best, that works great!

But there is still another thing: I would also like the same distance between code inputs and "normal" outputs (and there can also be `stdout` and `stderr` between them, which should have the same distances). I tried the same `\hrule` line there, but it didn't completely remove the gap.

And no, I don't want to control the distance between the baselines. I'd like to control the distance between the frames of the boxes.

jfbu:
> There is surely a way to get the font configured for fancyvrb in a way compatible to both cases, and even compatible with added user config, but probably this will need some @ letter thing. I will fetch it later from fancyvrb source code.


add this somewhere in your set-up where you define macros and environments

\makeatletter
\def\nbsphinxUseFVFont{%
  \FV@FontSize
  \FV@FontFamily
  \FV@FontSeries
  \FV@FontShape
  \selectfont
  \@noligs}
\makeatother

then in place of \ttfamily (or \small\ttfamily) use \nbsphinxUseFVFont. And \strut after it.

you can then test that it should work ok even after issueing \fvset{fontsize=\normalsize} for example. (I hope)
or \fvset{fontsize=auto}

jfbu:
> 
> add this somewhere in your set-up where you define macros and environments
> 
> \makeatletter
> \def\nbsphinxUseFVFont{%
> \FV@FontSize
> \FV@FontFamily
> \FV@FontSeries
> \FV@FontShape
> \selectfont
> \@noligs}
> \makeatother
> 

sorry suppress the un-needed \@noligs.

It can not work because this is all in second argument of \textcolor{}{}

hence catcode can not change, which is actually what \@noligs does.

even if you were to use {{\color{}\strut\nbsphinxUseFVFont %s }} (with extra
brace to be extra sure no color spill) this still wouldn't work because
of usage of \savebox from LaTeX2e which also fetches its argument.

One would have to employ TeX primitives directly 

\setbox\nbsphinxpromptbox\hbox to0pt{\hss{\color{<thecolorhere>}\strut\nbsphinxUseFVFont %s\ }}

That would be how to do it.

Attention I modified the \strut position. Indeed the \strut depends on current
font size. Now we want to use the same prompt as in SphinxVerbatim, but typically
the fancy cells will not use the "\small". So we want the normal \strut, not the small one.

But then once we have reached that stage we could again use your idea of \Verb

\setbox\nbsphinxpromptbox\hbox to0pt{\hss{\color{<thecolorhere>}\strut\Verb|%s|\ }}

but attention the control space will then not use the setting of monospace font
but the setting of normal roman font.

so it should be then

\setbox\nbsphinxpromptbox\hbox to0pt{\hss{\color{<thecolorhere>}\strut\Verb|%s |}}

note that here \Verb can change catcodes and it will make the space token act like \ 
We can't use \ now because \Verb is really changing catcodes.

src/nbsphinx.py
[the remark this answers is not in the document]
jfbu: Use

```
\kern-\dimexpr\sphinxverbatimsmallskipamount+\baselineskip-\FrameHeightAdjust\relax
\vspace{<whateveryouwant}
```

in-between to `sphinxVerbatim` (non-titled) environment following one another to control exactly the intermediate vertical space. I.e. if you use the negative `\kern` above and add no `\vspace` you should see the frames touching exactly.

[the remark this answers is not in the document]
mgeier: That works perfectly thank you so much!

src/nbsphinx.py
```diff
@@ -1438,7 +1439,8 @@ def visit_fancyoutput_latex(self, node):
     if prompt:
         prompt = nbconvert.filters.latex.escape_latex(prompt)
         out += r"""
-\savebox\nbsphinxpromptbox[0pt][r]{\textcolor{nbsphinxout}{\Verb|\strut{%s }|}}
+\savebox\nbsphinxpromptbox[0pt][r]{\color{nbsphinxout}%%
+\Verb|\strut{%s}\,|\hspace{\fboxrule}\hspace{\fboxsep}}
```
jfbu: alternatively you could remove the `\moveright` inside `\nbsphinxfancyaddprompt` (keep only `\copy\nbsphinxpromptbox`) and spare adding here these two `\hspace`'s. Matter of taste...

mgeier:
@jfbu I'm very satisfied with [the result](https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-codecells/), thank you again for your help!

I would say it is ready to merge now, but if there is anything else you want me to add/remove/change, please tell me.

jfbu:
@mgeier I don't see anything obvious to add here! I did not test but expect the code for framing fancy output cells will behave correctly in indented (lists) context.

mgeier:
@jfbu Thanks! Do you mean having fancy output cells as part of a list? This cannot happen, there is only one flat list of Markdown cells and code cells in a Jupyter notebook. But I tried having lists and block quotes as the content of a fancy output cell and this works nicely.

I've just squashed and merged this into `master`, but Github didn't realize that. I'll probably have to close this PR manually.

I've tried to use my newly learned TeX skills for improving the "separators" between notebooks in the documentation: #219.

This works quite nicely, but I think I'll need your help with controlling the page breaks.
Please see this PDF: https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-notebook-separators/

I've used some `\vfil` and some `\penalty` without really knowing what I'm doing. The point is that I want to avoid page breaks between the opening "separator" line and the following section. Same for the closing "separator" and the preceding section. This works more or less, but there are some problems in the output:

The "separator" on the top of page 28 should "stick" to the content above (i.e. some of the above content should be moved to page 28).

Page 29 has quite some empty space at the bottom, I guess this is a result of an inappropriate setting for `\penalty`. This should be filled with parts of the following page.

Page 37 has a lot of empty space that should be filled.

PR #219, opened 2018-10-10
DOC: Try to improve the "separators" between notebooks in LaTeX output

mgeier:
A PDF preview is available at https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-notebook-separators/

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -4,7 +4,7 @@
 # Use sphinx-quickstart to create your own conf.py file!
 # After that, you have to edit a few things.  See below.
 
-# Select nbsphinx and, if needed, add a math extension (mathjax or pngmath):
+# Select nbsphinx and, if needed, add a math extension (mathjax or imgmath):
 extensions = [
     'nbsphinx',
     'sphinx.ext.mathjax',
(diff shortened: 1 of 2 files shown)


mgeier:
@jfbu Yes please, that would be great!

As you can see from my last few comments, I've tried a few more things. I think I've made some improvements, but there is still an unwanted page break on the top of pages 28 and 30 in [the test PDF](https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-notebook-separators/).

And I'm sure all this could be done in a much better way!

> Can you tell me exactly how a section can end, is it always either after a sphinxVerbatim or after a fancycell?

No, it can be anything. A notebook can end with a Markdown cell, but it could even end with a "raw LaTeX" cell, in which case it could be really any LaTeX code.

jfbu:
About `\set@minipage` what is purpose, footnotes?

mgeier:
TBH, I don't completely understand what is does and why it is called like this, but I was using it to "swallow up" the additional spacing between my separation line and the following `\section`. I'm sure there is a better way to do it!

But please notice that although notebooks typically start with a section title, they don't have to. Sometimes they start with a normal text paragraph. They might also start with a code cell.

jfbu:
> I can try to implement the somewhat hacky idea mentioned there of making the "End of" part of the last frame, if notebook ends that way... but is it worth the effort ?

I found much better way. The point is that we have an a priori idea of the vertical space taken up by the "opening" and "closing" decorations, so my pessimism was unwarranted, see updated #220.

PR #220, opened 2018-10-11
DOC: alternative try to improve the "separators" between notebooks in LaTeX output

jfbu:
I kept the first three commits of #219.

Some constructs in LaTeX make it impossible (afaik) to prevent a pagebreak: once a negative penalty is in the vertical list, one can not get rid of it. Lists from LaTeX2e and "framed" environments from framed.sty encourage breaks before and after. The only ways to prevent a page-break after a frame are:

- hack into `\fb@afterframe` to prevent insertion of `\penalty-30`,
- or make the stuff one wants absolutely on same page a part of the framing itself.

A `sphinxVerbatim` additionally encapsulates the `framed` in a `trivlist`. The `trivlist` adds it own negative penalty (of value -51).

A somewhat intimidating idea would be, if it is known notebook ends with  a code cell to use not `sphinxVerbatim` but `sphinxLastVerbatim` where the latter would be wrapper hacking into both the `trivlist` and the `framed.sty` code to prevent insertion of negative penalties before the closing rule and "End of sentence...".

Thus, the prevsent PR is better for starting a notebook than for ending one. For "starting" the code tries to emulate the code for `\section` which encourages a pagebreak before and discourages after it. For "ending the notebook" we have mainly the problems listed above.

The main issue in LaTeX is that (afaik) once a structure as decided to make a pagebreak favourable it is very difficult (and I believe impossible) to undo that with 100% certainty.

(the only way is to gather the whole material in a big vbox and vsplit it later... the kind of things framed.sty or tcolorbox do, and nesting is very very problematic then)

side note: I did not manage to test my PR on nbsphinx doc, I get some 

```
File "/.../lib/python3.6/site-packages/jinja2/lexer.py", line 739, in tokeniter
    name, filename)jinja2.exceptions.TemplateSyntaxError: unexpected char '\\' at 103
```

when trying

```
$ python3 setup.py build_sphinx -b latex
```

As a result perhaps the relative vertical location of `The following section...` and `End of...` with respect to rule is perhaps not the good one.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -4,7 +4,7 @@
 # Use sphinx-quickstart to create your own conf.py file!
 # After that, you have to edit a few things.  See below.
 
-# Select nbsphinx and, if needed, add a math extension (mathjax or pngmath):
+# Select nbsphinx and, if needed, add a math extension (mathjax or imgmath):
 extensions = [
     'nbsphinx',
     'sphinx.ext.mathjax',
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -64,20 +64,30 @@ nbsphinx_prolog = r"""
 .. raw:: latex
 
     \vfil\penalty-1\vfilneg
-    \vspace{\baselineskip}
-    \textcolor{gray}{The following section was generated from
-    \texttt{\strut{}{{ docname }}}\\[-0.5\baselineskip]
-    \noindent\rule{\textwidth}{0.4pt}}
-    \vspace{-2\baselineskip}
+    \par
+    \smallskip
+    \noindent\textcolor{gray}{\scriptsize
+    The following section was generated from
+    \sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}.}
+    \par\nointerlineskip\kern-1ex
+    \noindent\textcolor{gray}{\rule{\textwidth}{0.4pt}}
+    \par\unskip\kern-1ex
+    \makeatletter\@setminipage\makeatother
 """
 
 # This is processed by Jinja2 and inserted after each notebook
 nbsphinx_epilog = r"""
 .. raw:: latex
 
-    \textcolor{gray}{\noindent\rule{\textwidth}{0.4pt}\\
-    \hbox{}\hfill End of
-    \texttt{\strut{}{{ env.doc2path(env.docname, base='doc') }}}}
+    \makeatletter\@minipagefalse\makeatother
+    \par\nointerlineskip
+    \noindent\textcolor{gray}{\rule{\textwidth}{0.4pt}}
+    \par\nointerlineskip\kern-1ex
+    \noindent\textcolor{gray}{\scriptsize\hfill
+    End of \sphinxcode{\sphinxupquote{\strut
+    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}}.}
+    \par
+    \smallskip
     \vfil\penalty-1\vfilneg
 """
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -63,32 +63,39 @@ nbsphinx_prolog = r"""
 
 .. raw:: latex
 
-    \vfil\penalty-1\vfilneg
     \par
-    \smallskip
-    \noindent\textcolor{gray}{\scriptsize
-    The following section was generated from
-    \sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}.}
-    \par\nointerlineskip\kern-1ex
-    \noindent\textcolor{gray}{\rule{\textwidth}{0.4pt}}
-    \par\unskip\kern-1ex
-    \makeatletter\@setminipage\makeatother
+    \makeatletter
+      \addpenalty\@secpenalty
+      \@tempskipa 2.3ex \@plus .2ex\relax
+      \addvspace\@tempskipa
+    {\color{gray}\scriptsize
+    \vbox{\parskip\z@skip\hsize\linewidth
+    \noindent The following section was generated from
+    \sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}.\par
+    \kern -1ex % this 1ex depends on \scriptsize value
+    \noindent\rule{\hsize}{0.4pt}%
+    \par
+    % add here \kern for negative or positive extra vertical space if needed
+    }}\@nobreaktrue\everypar{\@nobreakfalse}\nobreak
+    \makeatother
 """
 
 # This is processed by Jinja2 and inserted after each notebook
 nbsphinx_epilog = r"""
 .. raw:: latex
 
-    \makeatletter\@minipagefalse\makeatother
-    \par\nointerlineskip
-    \noindent\textcolor{gray}{\rule{\textwidth}{0.4pt}}
-    \par\nointerlineskip\kern-1ex
-    \noindent\textcolor{gray}{\scriptsize\hfill
-    End of \sphinxcode{\sphinxupquote{\strut
-    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}}.}
     \par
-    \smallskip
-    \vfil\penalty-1\vfilneg
+    \makeatletter
+    {\color{gray}%
+    \scriptsize
+    \nobreak\vtop{%
+    \kern -1ex
+    \parskip\z@skip\hsize\linewidth\parfillskip\z@skip
+    \noindent\rule{\hsize}{0.4pt}\par
+    \noindent\hfill End of \sphinxcode{\sphinxupquote{\strut
+    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}}.\par
+    }}\makeatother
+    \bigskip % or whatever is needed
 """
 
 # Input prompt for code cells. "%s" is replaced by the execution count.
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -69,15 +69,13 @@ nbsphinx_prolog = r"""
       \addpenalty\@secpenalty
       \@tempskipa 2.3ex \@plus .2ex\relax
       \addvspace\@tempskipa
-    {\color{gray}\scriptsize
-    \vbox{\parskip\z@skip\hsize\linewidth
+    {\color{gray}\parskip\z@skip\scriptsize
     \noindent The following section was generated from
-    \sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}.\par
-    \kern -1ex
-    \noindent\rule{\hsize}{0.4pt}%
+    \sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}\dotfill
     \par
-    }}\@nobreaktrue\everypar{\@nobreakfalse\everypar{}}\nobreak
+    }\@nobreaktrue\everypar{\@nobreakfalse\everypar{}}\nobreak
     \makeatother
+    \vskip-\parskip\nobreak
 """
 
 # This is processed by Jinja2 and inserted after each notebook
@@ -86,13 +84,10 @@ nbsphinx_epilog = r"""
 
     \par
     \makeatletter
-    {\color{gray}\scriptsize
-    \nobreak\vtop{\kern -1ex
-    \parskip\z@skip\hsize\linewidth\parfillskip\z@skip
-    \noindent\rule{\hsize}{0.4pt}\par
-    \noindent\hfill End of \sphinxcode{\sphinxupquote{\strut
-    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}}.\par
-    }}\makeatother
+    {\color{gray}\parskip\z@skip\scriptsize
+    \nobreak\noindent\dotfill\sphinxcode{\sphinxupquote{\strut
+    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}} ends here.\par
+    }\makeatother
     \kern-\baselineskip
     \penalty-9999
     \kern\baselineskip
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -64,6 +64,7 @@ nbsphinx_prolog = r"""
 .. raw:: latex
 
     \par
+    \needspace{3\baselineskip}
     \makeatletter
       \addpenalty\@secpenalty
       \@tempskipa 2.3ex \@plus .2ex\relax
@@ -72,11 +73,10 @@ nbsphinx_prolog = r"""
     \vbox{\parskip\z@skip\hsize\linewidth
     \noindent The following section was generated from
     \sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}.\par
-    \kern -1ex % this 1ex depends on \scriptsize value
+    \kern -1ex
     \noindent\rule{\hsize}{0.4pt}%
     \par
-    % add here \kern for negative or positive extra vertical space if needed
-    }}\@nobreaktrue\everypar{\@nobreakfalse}\nobreak
+    }}\@nobreaktrue\everypar{\@nobreakfalse\everypar{}}\nobreak
     \makeatother
 """
 
@@ -86,16 +86,17 @@ nbsphinx_epilog = r"""
 
     \par
     \makeatletter
-    {\color{gray}%
-    \scriptsize
-    \nobreak\vtop{%
-    \kern -1ex
+    {\color{gray}\scriptsize
+    \nobreak\vtop{\kern -1ex
     \parskip\z@skip\hsize\linewidth\parfillskip\z@skip
     \noindent\rule{\hsize}{0.4pt}\par
     \noindent\hfill End of \sphinxcode{\sphinxupquote{\strut
     {{ env.doc2path(env.docname, base='doc') | escape_latex }}}}.\par
     }}\makeatother
-    \bigskip % or whatever is needed
+    \kern-\baselineskip
+    \penalty-9999
+    \kern\baselineskip
+    \bigskip
 """
 
 # Input prompt for code cells. "%s" is replaced by the execution count.
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -739,16 +739,14 @@ class NotebookParser(rst.Parser):
         if resources.get('nbsphinx_orphan', False):
             rst.Parser.parse(self, ':orphan:', document)
         if env.config.nbsphinx_prolog:
-            rst.Parser.parse(
-                self,
-                jinja2.Template(env.config.nbsphinx_prolog).render(env=env),
-                document)
+            prolog = exporter.environment.from_string(
+                env.config.nbsphinx_prolog).render(env=env)
+            rst.Parser.parse(self, prolog, document)
         rst.Parser.parse(self, rststring, document)
         if env.config.nbsphinx_epilog:
-            rst.Parser.parse(
-                self,
-                jinja2.Template(env.config.nbsphinx_epilog).render(env=env),
-                document)
+            epilog = exporter.environment.from_string(
+                env.config.nbsphinx_epilog).render(env=env)
+            rst.Parser.parse(self, epilog, document)
 
 
 class NotebookError(sphinx.errors.SphinxError):


jfbu:
ah wait, there is a way:

- use the `\needspace` macro (already used by Sphinx) to simplify code for new notebook,

- for closing a notebook, it is possible to let TeX think the "End of..." occupies no vertical space, by typesetting it, move back up, add an attractive break point there, move down by the same distance we went up and add a penalty here. The problem though is that the "End of..." line will perhaps appear a bit on top of page footer (but that depends also `\footskip` parameter value.) or rather that page will go a bit further down than other pages.

I will give it a try, and I need to understand and fix the Jinja parsing thing.

mgeier:
Thanks a lot, that's so great!

I've created a PDF on RTD: https://media.readthedocs.org/pdf/nbsphinx/latex-notebook-separators/nbsphinx.pdf.

There you see that at the top of page 30 there is still an unwanted page break. But probably the special case of nested notebooks is not that important anyway ...

I'm still not quite perfectly satisfied with the distances: IMHO the distance before the notebook is a bit too large and the distance after it is a bit too small.

Also, I think I would like the text "The following ..." to be very close to the line. But I just got another idea: What about not having a separate full-width line but instead starting the line just after the text is over? And vice versa at the end? This would also make for two fewer line breaks to worry about ...
And what about using a gray `\dotfill` instead of a gray line?
This would also avoid the extra dot at the end of the file name, which I didn't like in the first place.

> apart from svg inclusion which caused problem at my locale

Sorry about that, this simply doesn't work yet. I'll have to tackle that at some point, see #3. It's quite an old issue as you can see by its number ...

doc/conf.py
```diff
@@ -63,22 +63,40 @@
...
-    \hbox{}\hfill End of
-    \texttt{\strut{}{{ env.doc2path(env.docname, base='doc') }}}}
-    \vfil\penalty-1\vfilneg
+    \par
+    \makeatletter
+    {\color{gray}\scriptsize
+    \nobreak\vtop{\kern -1ex
+    \parskip\z@skip\hsize\linewidth\parfillskip\z@skip
+    \noindent\rule{\hsize}{0.4pt}\par
+    \noindent\hfill End of \sphinxcode{\sphinxupquote{\strut
```
mgeier: I never liked the capitalized "E" and the extra dot at the end here.

I just had an idea how to avoid that: "\<docname\> ends here.", what do you think about that?

jfbu: sounds great

jfbu:
> There you see that at the top of page 30 there is still an unwanted page break. But probably the special case of nested notebooks is not that important anyway ...

I see no immediate fix, but will think about this more later (I am short on time for today)

> I'm still not quite perfectly satisfied with the distances: IMHO the distance before the notebook is a bit too large and the distance after it is a bit too small.

how  can both be true at the same time :) ? I wanted to add comments to code where one could modify spacing but that caused issue with Jinja lexer.

> Also, I think I would like the text "The following ..." to be very close to the line. But I just got another idea: What about not having a separate full-width line but instead starting the line just after the text is over? And vice versa at the end? This would also make for two fewer line breaks to worry about ...

Yes, you can do that with some `\hrulefill` I guess, but don't worry about the line breaks in current code before the two lines are in one vertical box and can't ever get separated.



>    apart from svg inclusion which caused problem at my locale

> Sorry about that, this simply doesn't work yet. I'll have to tackle that at some point, see #3. It's quite an old issue as you can see by its number ...

but here it's entirely my fault I have inkscape as Mac app but not in the PATH, I was too lazy to build it, and the one from Mac app does not like being called from command line, it needs full paths for image files.

As I said I am short on time today, feel free to push to this branch to implement the dashed lines or whatever (or tell me exactly what it is you want so I can add it later (tomorrow))

jfbu:
> Yes, you can do that with some \hrulefill I guess, but don't worry about the line breaks in current code before the two lines are in one vertical box and can't ever get separated.

I meant *because* the two lines are in one vertical box...

mgeier:
Thanks for your time, and don't worry, I'm not in a hurry!

> how can both be true at the same time

Sorry, I wasn't clear enough. I meant: IMHO the distance between the opening line and the beginning of the notebook is a bit too large and the distance between the end of the notebook and the closing line is a bit too small.

> but that caused issue with Jinja lexer

Ah, I see, `{%` has a special meaning for Jinja. I thought the only problematic situation is `{{` ... now I understand why the Jinja syntax can be switched (but I still think it isn't worth it in this case).

I just found a work-around, but it's not pretty:

    {{ '{' }}% my comment

This translates to

    {% my comment

A similar option would be

    {{ '{% my comment' }}

But I think the best option is to just not write comments, as you did.

> or tell me exactly what it is you want so I can add it later

I would prefer if you could add it. I don't want to mess up the TeX stuff ...

I tried changing it locally, and I think this looks quite nice:

```
\noindent The following section was generated from \sphinxcode{...} \dotfill
```
```
\noindent\dotfill{} \sphinxcode{...} ends here.
```

The color and font size can stay the same as before (unless you have a better suggestion).

doc/conf.py
```diff
@@ -63,22 +63,35 @@
...
-    \noindent\rule{\textwidth}{0.4pt}}
-    \vspace{-2\baselineskip}
+    \par
+    \needspace{3\baselineskip}
+    \makeatletter
+      \addpenalty\@secpenalty
+      \@tempskipa 2.3ex \@plus .2ex\relax
+      \addvspace\@tempskipa
+    {\color{gray}\parskip\z@skip\scriptsize
+    \noindent The following section was generated from
```
jfbu: This holds on one line, so I removed the vertical box encapsulation, but then the parskip has to be set to zero.

doc/conf.py
```diff
@@ -63,22 +63,35 @@
...
+    \sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}\dotfill
+    \par
+    }\@nobreaktrue\everypar{\@nobreakfalse\everypar{}}\nobreak
+    \makeatother
+    \vskip-\parskip\nobreak
```
jfbu: The distance from the "start of notebook" to next text baseline is `\parskip + \baselineskip`. I removed the `\parskip` here. The extra `\vskip` must be preceded by `\nobreak` (at end of line before `\makeatother`). The `\baselineskip` will be `12.6pt` with your settings (linespread=1.05) if next contents is plain text, and `18.9pt` if it is a section title as the latter uses `\Large`.

doc/conf.py
```diff
@@ -63,22 +63,35 @@
...
-    \textcolor{gray}{\noindent\rule{\textwidth}{0.4pt}\\
-    \hbox{}\hfill End of
-    \texttt{\strut{}{{ env.doc2path(env.docname, base='doc') }}}}
-    \vfil\penalty-1\vfilneg
+    \par
+    \makeatletter
+    {\color{gray}\parskip\z@skip\scriptsize
+    \nobreak\noindent\dotfill\sphinxcode{\sphinxupquote{\strut
+    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}} ends here.\par
+    }\makeatother
```
jfbu: Here also the above is on one line (I hope..., else if very long name causing two lines, a pagebreak is possible), so here too no vertical box, and again parskip must be set to zero to avoid too much space.

doc/conf.py
```diff
@@ -63,22 +63,35 @@
...
 .. raw:: latex
 
-    \textcolor{gray}{\noindent\rule{\textwidth}{0.4pt}\\
-    \hbox{}\hfill End of
-    \texttt{\strut{}{{ env.doc2path(env.docname, base='doc') }}}}
-    \vfil\penalty-1\vfilneg
+    \par
+    \makeatletter
+    {\color{gray}\parskip\z@skip\scriptsize
+    \nobreak\noindent\dotfill\sphinxcode{\sphinxupquote{\strut
```
jfbu: the line will be located at baselineskip (for scriptsize) distance from previous text baseline (if there is text above).

You can for example add a line above that one `\baselineskip5ex`, then the vertical separation will be 5ex (this is big), independently of whether last text line above has descenders (pqy...).

For a short distance I tested `\baselineskip2ex`, but then there is risk of overlapping with descenders, and if they are present TeX automatically lower the "ends here" line: I prevented this by also adding `\lineskiplimit -1ex`. But you would need this only for very low setting such as `\baselineskip 2ex` (which is only a bit more than the `\strut` height).

It seems regular baselineskip for scriptsize is fine enough. So no need to add any specific value assignment for `\baselineskip`.

doc/conf.py
```diff
@@ -63,22 +63,35 @@
...
+    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}} ends here.\par
+    }\makeatother
+    \kern-\baselineskip
+    \penalty-9999
+    \kern\baselineskip
+    \bigskip
```
jfbu: In case of a notebook included in another notebook, this "bigskip" will be between the two "ends here." lines. But is this really worrisome? (additionally, pagebreak can occur there, but again is that so problematic)

mgeier:
That's great, I love it!

IMHO the spacing is perfect now. And I like the dots .............

> Is it feasible to have some `nbsphinx.sty` file with these and possibly other LaTeX macros, and add `\usepackage{nbsphinx}` to the `usepackages` latex_elements key, and add `nbsphinx.sty` to `latex_additional_files`?

This would certainly be possible, but I'm not sure yet if it is really necessary.
In the beginning I was a bit nervous about finding all the separate files like the Jinja RST template, LaTeX files and CSS files, so I just included everything into the source file. This has also the advantage that everything is in one file and users can just copy the file `nbsphinx.py` around wherever they want and it only has to be found by Python, everything else works out-of-the-box from there. But this isn't an important argument, because it's just as easy to install it from PyPI.

There is something strange happening in the [current example PDF](https://readthedocs.org/projects/nbsphinx/downloads/pdf/latex-notebook-separators/): At the bottom of page 26, there is a rather large empty space. But even worse, the closing dotted line clashes with the footnote, even though there is a lot of space ...

jfbu:
> There is something strange happening in the current example PDF: At the bottom of page 26, there is a rather large empty space. But even worse, the closing dotted line clashes with the footnote, even though there is a lot of space ...

Yes I had seen it. The material at top of next page can only break at start of second paragraph `The following notebooks ...`, because breaking after section title is prohibited and breaking after first line of a paragraph is inhibited by penalty (I forgot with one clubpenalty or widowpenalty) and the a lone line at top of page is inhibited by penalty  (I forgot with one widowpenalty or clubpenalty). Perhaps I went overboard with the `-9999` penalty to encourage a pagebreak at end of a notebook (will experiment a bit).

Also, it is a fact that LaTeX inserts footnotes at the point the page contents end (might have to do with the `\raggedbottom` issued by sphinx.sty, need to check that). This all conspires to give that result...

jfbu:
I forgot to explain about the clash of the dotted line. That is due to the fact that the code in the PR makes the dotted line "invisible" to TeX: first we move back up vertically after having typeset the dotted line, then we say "hey TeX, you may break the page here if you like, I am hiding the dotted line which would make you hesitate if the page was full", then we move down vertically in case the pagebreak did not happen.

jfbu:
sorry for close-open, my mistake, wrong button...

doc/conf.py
mgeier: It's not a big deal, but I think the `\bigskip` is not necessary here.
What about moving it to the start of the "opening line"?

In fact, this looks better when changing from a notebook to a normal RST page (e.g. from Section 10 to 11 on page 35) and back (from section 11 to 12 on page 37).

What do you think?

mgeier: OK, I don't think this is a problem. IMHO the space before the very first notebook looks appropriate with a `\bigskip` in place.

mgeier:
Thanks a lot! This indeed looks like voodoo to me, but it seems to work nicely. I also tried it with another project and it also worked well.

I would say it's perfect now!

doc/conf.py
```diff
@@ -88,10 +88,15 @@
...
-    \kern-\baselineskip
-    \penalty-9999
-    \kern\baselineskip
+    }\dimen@\baselineskip
+    \vskip\z@\@minus\dimen@
+    \penalty-100
+    \vskip\z@\@minus-\dimen@
+    \vskip-\dimen@
+    \penalty9999
+    \vskip\dimen@
```
jfbu: This line and the two above probably do nothing good or bad (there are there because I sort of copied without too much thinking the `\needspace` code in reverse; rather than ensuring some space at bottom of page we try to prevent space at bottom of page.

doc/conf.py
```diff
@@ -88,10 +88,15 @@
...
+    \vskip\z@skip
+    \makeatother
```
jfbu: Unfortunately, although this code is very good at avoiding collisions with footnotes, it does allow more page break than

```
\kern-\baselineskip
\penalty-9999
\kern\baselineskip
```

used in earlier commit.

mgeier:
Thanks! I have no clue what the change is about, but if you say it's better, I believe you.
I didn't see any problems in the example PDF nor in another PDF I tired.

I was thinking about your `\startnotebook{...}` and `\stopnotebook{...}` idea and I think it would make a lot of sense.
For now, I would put the definition in the preamble-string in `nbsphinx.py`, the question of a separate `nbsphinx.sty` can be tackled at some later point in the future.

This would make for simpler usage:

```python
nbsphinx_prolog = r"""
{% set docname = env.doc2path(env.docname, base='doc') %}

.. only:: html

    html stuff

.. raw:: latex

    \startnotebook{\strut
    {{ docname | escape_latex }}}
"""

nbsphinx_epilog = r"""
.. raw:: latex

    \stopnotebook{\strut
    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}
"""
```

But it would also have the advantage that the actual TeX magic can be changed in the future and users will get the updated stuff automatically.

What about potential name clashes, do you think `\startnotebook` is safe?

src/nbsphinx.py
```diff
@@ -329,6 +329,65 @@
...
+    \ifdim\dimen@<\baselineskip
+      % little space left
+      \nointerlineskip
+      \def\nbsphinx@tmp{% will get executed after the closing line
+        % hide the height of the closing line
+        \kern-\baselineskip
+        % try to push down the footnote insert by counteracting \raggedbottom
+        % but afaik only a hack into the shipout routine could improve further
+        % (as is done by "bottom" option of footmisc package)
+        \nobreak\vskip\z@\@plus .001fil
```
jfbu: previous commit had `0.0001fil` but I used here `0.001fil` to try pushing footnote insert more

```
$ latexdef raggedbottom

\raggedbottom:
macro:->\def \@textbottom {\vskip \z@ \@plus .0001fil}\let \@texttop \relax
```

In extreme cases, the dotted line will be at same location as footnote separator line in my testing, but in 99% of cases, there will still be some small separation.

jfbu:
> Thanks! I have no clue what the change is about, but if you say it's better, I believe you.
> I didn't see any problems in the example PDF nor in another PDF I tired.

I stress test by manually editing LaTeX file with some additional `\vspace` to push down contents on the page. The code with all the `\vskip` allowed pagebreaks after code cells which did not happen with the earlier commits. Thus in my latest commit (now moved to LaTeX preamble) I again changed the method to both ensure no pagebreak and avoid to the best possible collision with footnotes.

mgeier:
Cool, now not only the result looks great, but also the usage is simple!

But this got me another idea: What if users want to express their creativity?

What about moving the "user-stylable" stuff to the user configuration?

I tried this and it actually seems to work. I moved the `\color` and `\scriptsize` closer to the actual text and moved everything out of the macro definition. The remaining thing is

    ... \noindent #1\par ...

With that, I can get the previous result with

```
nbsphinx_prolog = r"""
{% set docname = env.doc2path(env.docname, base='doc') %}

.. only:: html

    ...

.. raw:: latex

    \nbsphinxstartnotebook{\color{gray}\scriptsize
    The following section was generated from \sphinxcode{\sphinxupquote{\strut
    {{ docname | escape_latex }}}}\dotfill}
"""

nbsphinx_epilog = r"""
.. raw:: latex

    \nbsphinxstopnotebook{\color{gray}\scriptsize
    \dotfill\sphinxcode{\sphinxupquote{\strut
    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}} ends here.}
"""
```

This is LaTeX code which mere mortals like me can still understand. One could probably change `\color` to `\textcolor`, I don't know which is more typical.

This would allow users to change colors and text sizes, and to e.g. change the `\dotfill` to an `\hrulefill`, but it would also allow them to do crazy stuff like this:

```
nbsphinx_prolog = r"""
.. raw:: latex

    \nbsphinxstartnotebook{\color{blue}\footnotesize
    \begin{sphinxShadowBox}
    \hfill Beginning of \hfill
    {{ env.doc2path(env.docname, base='doc') | escape_latex }},
    \hfill hooray! \hfill \mbox{}
    \end{sphinxShadowBox}}
"""

nbsphinx_epilog = r"""
.. raw:: latex

    \nbsphinxstopnotebook{\vbox{\color{red}\footnotesize
    \rule{\textwidth}{0.5pt}
    \centerline{\strut End of
    {{ env.doc2path(env.docname, base='doc') | escape_latex }}, sad!}
    \par\nointerlineskip\rule{\textwidth}{0.5pt}}}
"""
```

What do you think about that?

jfbu:
I will think about it but there are some challenges here:

- the code works because it has an a priori idea of the size of the decoration, both for opening and for closing line,

- the code works as is only for material holding on one line, else one must use some `\vbox` or `minipage`.

- sphinxShadowBox you use in your example is a breakable box. I think but I am not sure one can use it in a minipage but I have forgotten now.

mgeier:
I was worried about the first point, too. But then I tried it with an extreme font size (I think it was `\Huge`) and it still seemed to work fine (as long as the content still fits into one line).

The second point is not a problem. I can write in the documentation that a single line should be used. And if not, multiple lines should be wrapped into a `\vbox{...}`. I've actually used a `\vbox{...}` in one of my crazy examples above.
Or alternatively, we could just always use `\vbox{...}`. Does it have and disadvantages if we have only a single line?

The third point is not a problem. The `sphinxShadowBox` was just a silly example. I don't think anybody would want to use that.

jfbu:
> I was worried about the first point, too. But then I tried it with an extreme font size (I think it was \Huge) and it still seemed to work fine (as long as the content still fits into one line).

Not surprising because the vertical spaces such as  `\baselineskip` or the `ex` unit are fontsize dependent. The problem is in case the contents occupy more than one line: the opening might not reserve enough space at bottom of page via  `\needspace` and the closing might not insert a good pagebreak at correct location.

Of course the decorations must be unbreakable themselves. Perhaps I can try to think about having a macro doing the encapsulation in a vbox or minipage (attention that LaTeX minipage like all other LaTeX things does `\leavevmode` which ultimately causes `\parskip` insertion which `\vbox` does not trigger), with some optional parameter indicating space to be reserved.

```
\nbsphinxstartnotebook[2cm]{fancy contents with sphinxShadowBox for example}
```

where `2cm` instructs that the thing wants at least this

(in actual code I use `3\baselineskip` which means it should be ok for decoration + section title + one line at least; as the baselineskip is the normal one, but the opening line has a very small one due to `\scriptsize`)

mgeier:
Hmmm, I see ...

IMHO, specifying the needed height is a bit annoying for a user.

Maybe I'm too naive, but can't you just put all the stuff into a savebox, use its height in `\needspace` and then place the box on the page?

mgeier:
Sorry for being annoying, but I had yet another idea (assuming the `\needspace` problem can be solved):

The existence of the `\nbsphinxstartnotebook` and `\nbsphinxstopnotebook` already simplified things enormously for the user. But I think we can go one step further!
What about applying the TeX magic automatically whenever `nbsphinx_prolog` or `nbsphinx_epilog` are used?

We could just insert raw LaTeX nodes into the document, we wouldn't need the special macros in the preamble anymore.

Here's a sketch for this, see `src/nbsphinx.py`, about line 800:

```
        if env.config.nbsphinx_prolog:
            latex = r"""
\makeatletter
\par
% more stuff
\noindent
\makeatother
"""
            document += docutils.nodes.raw(latex, latex, format='latex')
            prolog = exporter.environment.from_string(
                env.config.nbsphinx_prolog).render(env=env)
            rst.Parser.parse(self, prolog, document)
            latex = r"""
\makeatletter
\par
% more stuff
\makeatother
"""
            document += docutils.nodes.raw(latex, latex, format='latex')
        rst.Parser.parse(self, rststring, document)
        if env.config.nbsphinx_epilog:
            latex = r"""
\makeatletter
\par
% much more stuff!
\noindent
\makeatother
"""
            document += docutils.nodes.raw(latex, latex, format='latex')
            epilog = exporter.environment.from_string(
                env.config.nbsphinx_epilog).render(env=env)
            rst.Parser.parse(self, epilog, document)
            latex = r"""
\makeatletter
\par
% more stuff
\makeatother
"""
            document += docutils.nodes.raw(latex, latex, format='latex')
```

This would be the suggested example setting:

```
nbsphinx_prolog = r"""
.. raw:: latex

    \textcolor{gray}{\scriptsize
    The following section was generated from \sphinxcode{\sphinxupquote{\strut
    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}}\dotfill}

"""

nbsphinx_epilog = r"""
.. raw:: latex

    \textcolor{gray}{\scriptsize
    \dotfill\sphinxcode{\sphinxupquote{\strut
    {{ env.doc2path(env.docname, base='doc') | escape_latex }}}} ends here.}

"""
```

Ideally, users wouldn't have to worry about the page breaks at all, it should just work.

jfbu:
> IMHO, specifying the needed height is a bit annoying for a user.

The specified height would have been only an optional argument. If only one line, then no need to use it.

> Maybe I'm too naive, but can't you just put all the stuff into a savebox, use its height in \needspace and then place the box on the page?

Not naive. This can be done. You know I am the engineer avoiding to give too much  freeway to delirious Steve Jobs.

But you already rolled out the next step to the shareholders, so if I understand correctly the idea is to avoid having to tell the user to make use of `\nbsphinxstartnotebook` because it will be added automatically?

By the way, I learned something about `\textcolor` thanks to you. I always thought it was a horizontal box construct (like `\colorbox`) because I rarely use it and did not check its doc, but in fact not all. It is indeed  horizontal command (it uses `\leavevmode`) but it can get applied to contents which will flow across multiple lines (keeping the `\par` outside its argument) It is not `\long` so can't be used for multiple paragraphs (the same limitation for `\texttt` etc....) or with a `\par` in its argument.

I will try to integrate it all (a bit busy currently).

mgeier:
> if I understand correctly the idea is to avoid having to tell the user to make use of `\nbsphinxstartnotebook` because it will be added automatically?

Yes. And it would even be added if the user doesn't use a `.. raw:: latex` directive. It would also be inserted around reST code. I'm not sure how useful this really is, though.

I was just thinking about writing the documentation for this feature, where I would have needed to urge users to use `\nbsphinxstartnotebook`. Otherwise there might be unwanted page breaks. And I would have needed to tell them that they should only use a single line. And if not, they should use `\vbox{...}`.
And then I thought instead of giving complicated instructions to users, wouldn't it be better if that works automatically?

For my own uses, I don't really care if I have to use `\nbsphinxstartnotebook` or not.

If you think it's a silly idea, please tell me. I'm basically brainstorming.

src/nbsphinx.py
```diff
@@ -331,20 +331,20 @@
 \setlength{\nbsphinxcodecellspacing}{0pt}
 
 % Define support macros for attaching opening and closing lines to notebooks
+\newsavebox\nbsphinxbox
 \makeatletter
 \newcommand{\nbsphinxstartnotebook}[1]{%
     \par
     \bigskip
```
jfbu: this `\bigskip` should arguably be moved over away from `\nbsphinxstartnotebook` to let user insert it via Jinja template if so desired.

mgeier:
Thanks, that works nicely!

src/nbsphinx.py
mgeier: I think it makes sense to add some space here, because we are taking away some vertical space between the "separation line" and the start of the notebook (especially if it starts with a section title).

Ideally, we would add the same space here that we remove below, but I guess that's not possible, right?

The `\bigskip` was just a compromise that looked OK to me, but if you think it's better to remove it from here, please do so!

jfbu: I agree some extra vertical space is needed (as the code inserts next only  the vertical space typical of a `\section` heading) but I was thinking it perhaps better if left to user choice, to use or not the `\bigskip`, hence my proposal was to move it out the macro `\nbsphinxstartnotebook` itself, and leave it to user insertion via the rst_prolog.

jfbu:
> Thanks, that works nicely!

I did some testing for `stopnotebook` but I don't succeed currently to absolutely definitely make a pagebreak impossible, especially when following immediately a code cell. But I tested it with rather extreme example

capture d ecran 2018-10-17 a 12 30 33

src/nbsphinx.py
mgeier: OK, that's fine by me.

We can use the `\bigskip` (or something similar) in the `conf.py`.

mgeier:
If there are some extreme cases where a page break happens, I can live with that. AFAICT, the "normal" cases work perfectly now.

I think we can use your current solution for now, and if you (or somebody else) come up with further tweaks, we can add them to the macro definitions at a later point.

jfbu:
> I did some testing for stopnotebook

Sorry I wasn't clear: I meant I tested various modified  `\nbsphinxstopnotebook` macros using a measurement step and checking page totals, allowing the "closing line" to be multiple paragraphs in an unbreakable box, the problem is that when applied to our current "one-line" thing, it is not as good into preventing pagebreak as is current code in the branch.

This is why I don't commit yet a `stop` allowing multi-paragraph as in screenshot in my previous comment. On the other hand the `start` was stress-tested and seems to behave well.

src/nbsphinx.py
```diff
@@ -355,37 +354,22 @@
...
-        \penalty-9999
-        % if TeX decides to move the closing line to next page (necessarily
-        % together with some material above it) the above vskip and negative
-        % kern must get compensated; if on the other hand the pagebreak
-        % happens here, they get discarded and don't offset the next page.
-        \vskip\z@\@plus -.001fil
-        \kern\baselineskip
-        }%
+      \unvbox\nbsphinxbox
+      \kern-2\baselineskip
```
jfbu: I did some tests using `\kern\dimen@` here but it really is too drastic in allowing a long "stop notebook" to spill over into bottom of page, sometimes overwriting footnotes. Using `2\baselineskip` looked like better compromise, but  I did not fully test.

jfbu:
> The code will try to keep about 2 text lines from this at all cost on same page, but is allowed to move some material to next page.

not exactly. Only the top line of closing contents is (I hope) guaranteed never ever to allow a pagebreak above it. And if the whole thing has at most two lines (especially if in smaller font size) they should always also stay on page. But if its longer than that, some material might move to next page. Best I can do with my current level of understanding of TeX (beyond that a mastery of the specific LaTeX shipout routine is probably needed, but that's horrendously complicated).

src/nbsphinx.py
```diff
@@ -329,6 +329,49 @@
...
+    \nobreak % it updates page totals
+    \dimen@\pagegoal
+    \advance\dimen@-\pagetotal \advance\dimen@-\pagedepth
+    \advance\dimen@-\ht\nbsphinxbox \advance\dimen@-\dp\nbsphinxbox
+    \ifdim\dimen@<\z@
+      % little space left
+      \unvbox\nbsphinxbox
+      \kern-2\baselineskip
+      \nobreak\vskip\z@\@plus .01fil
+      \vskip\z@\@plus -.01fil
```
jfbu: Sorry, I forgot a `\penalty` here else this does not make much sense. Problem is that without it, footnotes may get overprinted by the closing material, and with it I have some cases (when the closenotebook happens after a `\section` + one line of text) where the `\penalty` moves to next page and causes it to break near top if it zero or negative. I had seen all this in my testing yesterday but pushed a bit early. (there are many configurations to test, and perhaps I focused too much on some rare ones)

Please wait a bit.

I thought about `\enlargethispage` which is provided by LaTeX but got no positive result at my first try.

src/nbsphinx.py
```diff
@@ -363,10 +363,11 @@
     \ifdim\dimen@<\z@
       % little space left
       \unvbox\nbsphinxbox
-      \kern-2\baselineskip
-      \nobreak\vskip\z@\@plus .01fil
-      \vskip\z@\@plus -.01fil
-      \kern2\baselineskip
+      \kern-.8\baselineskip
+      \nobreak\vskip\z@\@plus1fil
+      \penalty100
```
jfbu: We need a penalty here else TeX can not break at the next glue (which follows glue). But that penalty might move to next page if TeX has decided to move previous stuff already to next page. So it must be somewhat positive else it may cause a pagebreak of next page. Taking value 100 worked satisfactorily.

This however inserts potential breakpoint when a notebook closing is immediately followed by another one, but not much choice here. Thus in thoses cases it is possible the innermost closing line is at bottom of page, and the outermost one ends up at top of next page.

src/nbsphinx.py
```diff
@@ -363,10 +363,11 @@
...
+      \vskip\z@\@plus-1fil
+      \kern.8\baselineskip
```
jfbu: This is a compromise, the bigger we take this the more a small closing line (using say `\scriptsize` hence with small baseline -- but the `\baselineskip` here is the one at normal size), might appear inside the footnote space. But if we take it too small, the result will be less good with closing lines using say `\Large`.

src/nbsphinx.py
mgeier: I can live with that!

I like using notebooks within notebooks, but normally the outer notebook has little text and very little to no code. It mostly acts as sub-table-of-contents in HTML and is barely visible in LaTeX output. Therefore, I would in my projects use Jinja to disable the prolog/epilog for those outer notebooks.

I think the example in the docs (https://nbsphinx.readthedocs.io/en/0.3.5/subdir/toctree.html) is a quite un-typical use case because the outer notebook has much more content than the inner one.

A more typical example is there: https://splines.readthedocs.io/en/latest/catmull-rom.html. In that case I will disable the prolog/epilog for certain notebooks (as soon as the next `nbsphinx` release is out).

src/nbsphinx.py
mgeier: TBH, I don't know how this works. But can't you somehow use (a fraction of?) the size of the `\nbsphinxbox` here instead of `.8\baselineskip`? Then it would scale with the content of the closing line?

doc/conf.py
```diff
@@ -63,22 +63,21 @@
 
 .. raw:: latex
 
-    \vfil\penalty-1\vfilneg
-    \vspace{\baselineskip}
+    \par\bigskip
```
mgeier: Is the `\par` necessary here?

IMHO this is not intuitive for a "normal" LaTeX user.

doc/conf.py
```diff
@@ -63,22 +63,21 @@
...
 # This is processed by Jinja2 and inserted after each notebook
 nbsphinx_epilog = r"""
 .. raw:: latex
 
-    \textcolor{gray}{\noindent\rule{\textwidth}{0.4pt}\\
-    \hbox{}\hfill End of
-    \texttt{\strut{}{{ env.doc2path(env.docname, base='doc') }}}}
-    \vfil\penalty-1\vfilneg
+    \nbsphinxstopnotebook{\scriptsize
+    \noindent\strut\textcolor{gray}{\dotfill\sphinxcode{\sphinxupquote{\relax
```
mgeier: What does the `\relax` do here?

Is it just to avoid the additional space?

If that's the case, you could just move the `\strut` there instead?

doc/conf.py
```diff
@@ -63,22 +63,21 @@
...
 
-    \vfil\penalty-1\vfilneg
-    \vspace{\baselineskip}
+    \par\bigskip
+    \nbsphinxstartnotebook{\scriptsize\strut
     \textcolor{gray}{The following section was generated from
-    \texttt{\strut{}{{ docname }}}\\[-0.5\baselineskip]
-    \noindent\rule{\textwidth}{0.4pt}}
-    \vspace{-2\baselineskip}
+       {\sphinxcode{\sphinxupquote{\strut{}{{ docname | escape_latex }}}}}\relax
```
mgeier: What does the `\relax` do here?

Ideally, I would like to avoid such low-level commands in this user-visible configuration suggestion.

mgeier:
> Do you want me to rearrange and squash commits?

Yes please, that would be great!

src/nbsphinx.py
jfbu: Fwiw, LaTeX has an `\enlargethispage`, but its mechanism is especially supported in the LaTeX output routine. Brief testing with it was not conclusive. We can't change at TeX macro level what has been previously contributed to the page. That's the core problem here. Environment like Framed encourage a pagebreak after them, and it is hard to counteract this. Besides, Framed (used by code-cell and fancy output cells) has no idea it will be followed by the end of the notebook. TeX never has any global vision of the document, it proceeds linearly and breaks the page as soon as it has enough material and suitable break points. Moreover LaTeX (despite common misunderstandings among user base of its "structural" aspects) never has any global view of the document structure, there is no out-of-the box way for a LaTeX environment to know it is last in a section for example. There is nothing even remotely analogous to a tree object of the document in LaTeX.

doc/conf.py
mgeier: OK, I see. Would something like `\vspace{1em}` be better than using `\bigskip`?

doc/conf.py
mgeier: Another question: what does the outer level of braces do in this line?
Is it really necessary?

doc/conf.py
jfbu:
```
$ latexdef bigskip

\bigskip:
macro:->\vspace \bigskipamount 

$ latexdef bigskipamount

\bigskipamount:
\skip19


\the\bigskipamount:
12.0pt plus 4.0pt minus 4.0pt
```

As you see `\bigskip` is a LaTeX command, which adds some stretch and shrink to the page, which is a good idea for separating high level structures. TeX needs glue in the page.

doc/conf.py
jfbu: No idea why this extra level of braces, it is completely unneeded.

doc/conf.py
jfbu: > LaTeX modifies TeX `\bigskip` in a way I don't like too much

This is the `\vspace` which you can use in a paragraph, it uses `\vadjust` to add space in-betwee lines.

```
$ texdef bigskip

\bigskip:
macro:->\vskip \bigskipamount 

```

The original TeX as above uses `\vskip` which forces vertical mode. Compare using `\vspace{8pt}` with `\vskip 8pt\relax` mid-paragraph. The later closes a paragraph and adds a new one, the former increases the spacing of some two lines of paragraph.

I largely prefer the `\vskip` version.

mgeier:
Thanks a lot for the explanations, I'm learning a lot more details about TeX and LaTeX than I ever thought I would!

I'm sorry to keep bothering you, but I'm starting to really dislike the fact that a user has to provide an extra vertical spacing. I think this should have a good spacing by default, and users can add any positive/negative vertical space if they really want to change the default.

I think I know now what a good default spacing would be: The vertical space before a "separator line" should be the same as the space between a normal paragraph and the following (top-level) `\section` title (no matter whether the notebook actually starts with a section or not).

I've made an example:

grafik

I think by default all spaces marked by arrows should have (about) the same vertical distance.

I think to get this, we would need an additional `\vskip\parskip` somewhere. But when I tried it, the space was still too small, so I guess some additional space should be inserted. But I don't know how much exactly and where exactly ...

jfbu:
I can not use the web interface due to some oddities with github today.

After login, when I move to visit https://github.com/spatialaudio/nbsphinx/pull/220
it looks as if I am not logged to my github account and I can't use the web interface.

side remark: I am dead tired of the software industry and its total lack of ethics
and its complete take over by the forces of advertisement and big data. I am 
almost certain the problem here is with Firefox "disable tracking" turned on,
combined with private browsing mode. But I will not test this.

I will stop all activities on github if these conditions persist.

as per your problem

- I made a mistake the correct vertical spacing in startnotebook
should have been \@tempskipa 3.5ex \@plus 1ex \@minus .2ex\relax

- however there will still be missing space because when a section title
is typeset it inserts baselineskip glue, corresponding to \Large fontsize
(in the current context) but here the all thing is typeset in a vbox, which is
then unvbox-ed. No baselineskip glue is added under such circumstances
when the hboxes from the unboxed vbox are added to the main vertical list.

Should vertical distance be measured to top or to baseline
of text ? The latter option is more or less impossible because the user 
is free to use \scriptsize or \Huge in the added material and \nbsphinxstartnotebook
can't know about it.

If it is measured to top we can do computation

extra vertical space= baselineskip at Largesize - height strutbox at Large size
or height of typical letter like if heading was titled "Foo".

with your font config this means about 5.59pt in my testing
 (this value is font dependent because the skip above is expressed in ex
  units which are font dependent, but no choice for us because \section
  uses such units, although package titlesec makes it all even more involved)

again this is extra space which would make the space look the same
if "startnotebook" actually used \Large font size for its one-liner.

it is not baseline to baseline distance, but baseline to top of inserted box
(which in \scriptsize is not very tall)

Pushing a commit soon. (if command line still works)

jfbu:
Le 22 oct. 2018 à 10:23, jfbu <jfbu@free.fr> a écrit :

> Pushing a commit soon. (if command line still works)

done, but I see nothing on the we interface. Attaching the patch,
in case this mail is actually received by GitHub even though
is is not emitted by  a Windows computer

jfbu:
Le 22 oct. 2018 à 10:23, jfbu <jfbu@free.fr> a écrit :

> Should vertical distance be measured to top or to baseline
> of text ? The latter option is more or less impossible because the user 
> is free to use \scriptsize or \Huge in the added material and \nbsphinxstartnotebook
> can't know about it.

It can by defining the box as a \vtop and get its height

ah, well maybe another commit then

(the box is unvbox-ed to get correct spacing **after it**, else, it would
appear as big box and next text line (section heading typically)
will have no useful reference point; but we can define it as \vtop
so as to get height of first line in it, then unvbox it)

jfbu:
> I will stop all activities on github if these conditions persist.

turns out problems had hit github data storage... and I was not the target :)

... to explain my rant, (I'll stay brief or at least try), I could not pay my electricity bill without disactivating the Firefox protection against tracking. With it the Google captchas would not show. And I must go via Google captchas as I navigate in private browsing mode and the (public state) company site does not like that. Anyway, after understanding that the problem this time was not with AdBlock Plus being irritating for the web interface of my electric power provider, but the disabling tracking feature of Firefox, I turned it off and paid my 40 euros bill. To discover after that there were **27** cookies on my computer. **27** cookies for having paid one 40 euros bill. I am fed up with the tracking industry and I am not even sure my electricity provider company knows about these cookies apart 2 or 3 possibly.

That was explanation for the rant. I simply don't want the big data industry invading all aspects of my live. And sadly I live in a country which is very complacent about it, in contrast to Germany where is a bit more of awareness after the Stasi years.

I understand you may disagree my rant polluting this thread, and will not say anything more.

Issue #12, opened 2016-01-17
LaTeX: Images are too big


mgeier:
I already [posted this on the Jupyter mailing list](https://groups.google.com/forum/#!topic/jupyter/TTMpJkem7jE), I'm just repeating it here:

The images (which look fine in the HTML output) are hideously large in the LaTeX output, see e.g.
https://media.readthedocs.org/pdf/nbsphinx/0.2.2/nbsphinx.pdf

This also happens with the "normal" nbconvert.

How can I get a reasonable image size?
Ideally this should work out-of-the box ...

I don't know if that's addressed already somewhere, I only found some mentions of image size in https://github.com/jupyter/nbconvert/pull/178 and https://gist.github.com/kroger/3856821.

Any further pointers would be very helpful!

actsasgeek:
Has anyone resolved this? I'm noticing the same thing and I can't seem to do anything to fix it.

mgeier:
It hasn't been resolved in `nbsphinx`, but there were some changes to `nbconvert` (see the link above, and there are probably further changes) which could probably be ported to `nbsphinx`.
@actsasgeek Do you want to give it a shot?

mgeier:
I've created #222 to solve this, but there are still a few issues ...

Issue #52, opened 2016-06-17
Support images with raw html syntax


fperez:
There is some background discussion in [this Jupyter notebook issue](https://github.com/jupyter/notebook/issues/1443), but the gist of the issue is that nbsphinx currently doesn't recognize images declared e.g. as

``` html
<img src="images/dashboard_files_tab.png" width="791px"/>
```

only ones that use the plain markdown syntax:

``` markdown
![alt](images/dashboard_files_tab.png)
```

This obviously precludes passing (useful) things like custom size hints.  I don't know if this is easy to fix in nbsphinx itself or if there's an underlying tool (say pandoc) that does the heavy lifting with the files... But I figured I'd mention it, it would help us use whichever syntax is more sensible in a given setting.

For now, for Jupyter we'll work around it in our docs by limiting ourselves to the `![..](..)` syntax.

mgeier:
Currently, pandoc is used to convert markdown cells, but in the not-so-near future this will probably be changed, see #36.

`nbsphinx` already parses raw HTML content for alert/warning `<div>`s (see #47 and [the docs](http://nbsphinx.readthedocs.io/en/latest/markdown-cells.html#Info/Warning-Boxes)), something similar could probably be done for `<img>` tags.

How is this handled in `nbconvert` with LaTeX output?

takluyver:
> How is this handled in nbconvert with LaTeX output?

Also pandoc, so it may well have the same issue.

mgeier:
Alternatively, `<img>` tags could be converted to "real" image nodes with a pandoc filter like in #35, but I don't think this has any advantage over doing it with a Sphinx `Transform`.

However, yet another option would be to convert *all* raw HTML content (not only `<img>` tags) to [raw HTML docutils nodes](http://docutils.sourceforge.net/docs/ref/rst/roles.html#raw).

I don't know if that would work, and I don't know if that's actually a good idea, either.

This way, the images wouldn't show up in LaTeX output.

actsasgeek:
Was this resolved? I've taken to using the Jupyter notebook/iPython helpers but importing libraries just to show images in notebooks that are all Markdown is kind of kludgy.

mgeier:
@actsasgeek No, nothing has changed since then. Do you have any suggestions on how to proceed?

fperez:
Mmh, I just ran into this issue again today, so piping in here and pinging @mpacer to see if they have any ideas... 

[...]

PR #222, opened 2018-10-18
Get sizes from image files for better LaTeX output

mgeier:
See #12.

This mostly works, but there are a few remaining problems:

* [ ] Getting the size doesn't seem to work for PDF images

* [ ] `data:` URIs don't work

* [x] images are unnecessarily turned into links in HTML output (see https://github.com/sphinx-doc/sphinx/issues/4229 and [the `html_scaled_image_link` option](http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_scaled_image_link)). For now, I've used `html_scaled_image_link` in the docs.

* [ ] The maximum size is not constrained (`nbconvert` does that). `\sphinxincludegraphics` has this feature built-in, but it works only if no height/width is given (see `sphinx.sty`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ matrix:
   - env: ADDITIONAL_COMMAND="python -m sphinx doc/ doc/_build/ -b linkcheck"
 
   # a few older Sphinx releases using default Python version
-  - env: SPHINX="==1.3.2 docutils<0.13.1"
-  - env: SPHINX="==1.3.6 docutils<0.13.1"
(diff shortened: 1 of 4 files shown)


mgeier:
@rddaz2013 With my current solution (see 3f2a29ce4aea4d45be514ab37606c70e8f1044dc) the image size is taken from the file itself, no need to explicitly specify it.

The syntax you are mentioning would be a Markdown extension, you can try to ask the Jupyter people if such a thing is planned.

Alternatively, HTML syntax could be used, see #52.

Issue #240, opened 2018-11-15
Buggy white spaces and line breaks in markdown tables


falexwolf:
Hi, thank you for the great package!

The following simple table

test1 | test2
----- | ------
word | many words in a row, long_words_here_to_which_break_up_in_several_pieces

renders as this:

![image](https://user-images.githubusercontent.com/16916678/48581787-c635a680-e8f0-11e8-851b-fc836acf1e84.png)

Any help is appreciated!

mgeier:
Thanks for the report!

I cannot reproduce the problem, though.

Can you please provide some information about your system and how you installed `nbsphinx` and, more importantly, how you installed `pandoc` and which version of `pandoc` you are using?

rddaz2013:
Yes i have the same problem - but not in all tables and more in the latex-output.
I look into it.

Please @falexwolf show us the  rst or html code of the table.

mgeier:
@rddaz2013 

> can we fetch the rst output from nbsphinx before sphinx gets the table

Yes, you can use the `save-rst` branch for debugging, see https://github.com/spatialaudio/nbsphinx/tree/save-rst.

rddaz2013:
i think https://github.com/jgm/pandoc/issues/3276 got it right.

The width header of the header breaks the line...if the text in die cell is much bigger. 
I test the workaround with the --columns=500 into the pandoc comandline

PR #246, opened 2018-11-25
pandoc table fix

rddaz2013:
A fast and easy change for the issues #240

Please test it with the newest nbsphinx  (my own is a old branch) with the html and latexpdf output.

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -1093,7 +1093,8 @@ def pandoc(source, fmt, to, filter_func=None):
         # see issue #155
         cmd += ['--eol', 'lf']
     cmd1 = cmd + ['--from', fmt, '--to', 'json']
-    cmd2 = cmd + ['--from', 'json', '--to', to]
+    to = to + '+grid_tables'
+    cmd2 = cmd + ['--from', 'json', '--to', to,'--columns=500']
 
     p = subprocess.Popen(cmd1, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
     json_data, _ = p.communicate(encode(source))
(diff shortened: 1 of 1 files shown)


src/nbsphinx.py
```diff
@@ -1093,7 +1093,8 @@ def decode(data):
         # see issue #155
         cmd += ['--eol', 'lf']
     cmd1 = cmd + ['--from', fmt, '--to', 'json']
-    cmd2 = cmd + ['--from', 'json', '--to', to]
+    to = to + '+grid_tables'
```
mgeier: Is `+grid_tables` really necessary?

src/nbsphinx.py
```diff
@@ -1093,7 +1093,8 @@ def decode(data):
...
+    cmd2 = cmd + ['--from', 'json', '--to', to,'--columns=500']
```
mgeier: I'd prefer if the additional option would be added in a separate  line.
And you should probably also add a reference to the issue nuber, e.g. something like:

```python
cmd2 += ['--columns=500']  # Avoid breaks in tables, see issue #240
```

rddaz2013: I remember your note

src/nbsphinx.py
rddaz2013: I don't know really? because at the time the latex output of the tables look's stupid. Next step is the make the templates for latex table work - in the html output it works. The workflow is not good with table -> 

good table in ipython&html = bad tables in latex.
bad tables in ipython = nice tables in html but still bad tables in latex .-(

mgeier:
Thanks, I've merged your changes in 27a38ca97b87836dbefe2ee17bb8eaa72f38328e.

mgeier:
Thanks, I've merged your changes in 27a38ca97b87836dbefe2ee17bb8eaa72f38328e.

PR #268, opened 2018-12-18
Make wheel non-universal again

mgeier:


ChrisBarker-NOAA:
IT's a total kludge, but it would be nice if folks got a nice error message if they try to run this code under py2.

What I got was a syntax error, due to the use of "nonlocal" -- but that wouldn't be obvious to a casual user what the issue is...

mgeier:
@ChrisBarker-NOAA Do you have an idea how to implement such an error message?

I was assuming that Sphinx itself wouldn't even run on Python 2 (because they claim to support only Python 3.5+), but I just tried it and it can be installed and executed with Python 2 without any error message.

I could also remove the `nonlocal` and keep Python 2 compatibility, but I'm not sure if I want that. I have the feeling that most packages in the scientific Python area are ditching Python 2 nowadays.

ChrisBarker-NOAA:
well, you could do something like:

```
if sys.version_info < (3, 5)):
    raise RuntimeError("This package can only be used with Python 3.5 or greater")
```

LIke I said -- a kludge!

It seems there should be something in Python or the packaging ecosystem to make this more automatic...

ChrisBarker-NOAA:
Of course— d’uh!

Thus is kind of an intractable problem. 

Oh well.

mgeier:
I'm reverting this in #275, because apparently some people still use Python 2.

PR #270, opened 2018-12-18
WIP handle duplicate section titles

### LINKED ISSUE #262: Section numbering gets wrong when sections have the same titles
When there are two sections with the same title, the number for the first one is the same as the one for the second section.

Example markdown cell:
```markdown
# Title
## Normal
## Duplicated
## Duplicated
```

The output is:

```
1. Title
1.1 Normal
1.3 Duplicated
1.3 Duplicated
```

Note that there is no 1.2 section. Instead, there are two 1.3 sections.

Environment info:
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.8.2
- nbsphinx 0.3.5

This bug doesn't happen when you create a test .rst file, so it must be something wrong with nbsphinx, not sphinx.

mgeier:
This is work-in-progress with the goal to fix #262.

For now, this should fix the section numbers, but the anchor IDs of the duplicated titles are not quite right yet.

As always, I'm open to suggestions for improvements!

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -1363,9 +1363,14 @@ class CreateNotebookSectionAnchors(docutils.transforms.Transform):
     default_priority = 200  # Before CreateSectionLabels (250)
 
     def apply(self):
+        all_ids = set()
         for section in self.document.traverse(docutils.nodes.section):
             title = section.children[0].astext()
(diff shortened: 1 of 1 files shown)


mgeier:
I couldn't come up with a better solution, but I think it's better to merge this anyway ...

Issue #276, opened 2019-01-08
larger images break LaTeX PDF generation


madmott:
A reproducing example can be found at https://github.com/datarobot/nbsphinx-minimal-repro (pared down from my project's normal build process - there's some vestiges that amount to "this is here because it's needed in the full version", my apologies).

When I use nbsphinx==0.3.5, everything builds correctly.  When I use nbsphinx==0.4.0 or nbsphinx==0.4.1 I get the normal PDF up until the image, followed by about 5000 blank pages instead of the image, and then the build process errors.

Logs from the end of the build process:
```
Underfull \vbox (badness 10000) detected at line 257
! TeX capacity exceeded, sorry [parameter stack size=10000].
\__prop_split_aux:w ... #2#3\q_mark #4#5\q_stop ->
                                                  #4{\tl_set:Nn \l__xeCJK_fo...
l.257 \end{nbsphinxfancyoutput}
```

The specific problematic section in the generated .tex file seems to be
```
\begin{nbsphinxfancyoutput}

\noindent\sphinxincludegraphics[width=715\sphinxpxdimen,height=852\sphinxpxdimen]{{examples_advanced_model_insights_Advanced_Model_Insights_3_1}.png}

\end{nbsphinxfancyoutput}
```
When I modify the .tex file replace it with the following, the build works again.
```
\begin{nbsphinxfancyoutput}

\noindent\sphinxincludegraphics{{examples_advanced_model_insights_Advanced_Model_Insights_3_1}.png}

\end{nbsphinxfancyoutput}
```

This may be related to https://github.com/spatialaudio/nbsphinx/pull/222 and related work with images.

For now I'm simply downgrading as a workaround.

madmott:
This may actually be a duplicate of https://github.com/spatialaudio/nbsphinx/issues/231 - I'm not entirely sure.

mgeier:
Thanks for the report!

Yes, this is related to #222 and #231, but I didn't think the error would be that bad.

Sadly, I don't really know a solution to this.

The `\sphinxincludegraphics` macro actually is able to shrink big images to fit on the page, but this only works if `width` and `height` are not given.

Probably @jfbu has an idea?

mgeier:
@madmott A possible work-around would be to use PDF plots, did you try that yet?
See https://nbsphinx.readthedocs.io/en/0.4.1/code-cells.html#Plots.

jfbu:
[...]

PR #277, opened 2019-01-11
LaTeX: ensure maximal height of included graphics less than text height

jfbu:
Closes: #276

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -388,6 +388,33 @@ LATEX_PREAMBLE = r"""
       \unvbox\nbsphinxbox
     \fi
 }% end of \nbsphinxstopnotebook
+
+% Avoid fatal error with framed.sty if graphics too long to fit on one page
+\newdimen\nbsphinx@image@maxheight
+\AtBeginDocument{%
+  \nbsphinx@image@maxheight\textheight
+  % -20pt was found insufficient in testing
+  \advance\nbsphinx@image@maxheight -2.5\baselineskip
+}
+\renewcommand*{\sphinxincludegraphics}[2][]{%
+    \gdef\spx@includegraphics@options{#1}%
+    \setbox\spx@image@box\hbox{\includegraphics[#1,draft]{#2}}%
+    \in@false
+    \ifdim \wd\spx@image@box>\linewidth
+      \g@addto@macro\spx@includegraphics@options{,width=\linewidth}%
+      \in@true
+    \fi
+    % no rotation, no need to worry about depth
+    \ifdim \ht\spx@image@box>\nbsphinx@image@maxheight
+      \g@addto@macro\spx@includegraphics@options{,height=\nbsphinx@image@maxheight}%
+      \in@true
+    \fi
+    \ifin@
+      \g@addto@macro\spx@includegraphics@options{,keepaspectratio}%
+    \fi
+    \setbox\spx@image@box\box\voidb@x % clear memory
+    \expandafter\includegraphics\expandafter[\spx@includegraphics@options]{#2}%
+}% end of redefined \sphinxincludegraphics
 \makeatother
 """
 
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -322,6 +322,12 @@ LATEX_PREAMBLE = r"""
 % Define an environment for non-plain-text code cell outputs (e.g. images)
 \makeatletter
 \newenvironment{nbsphinxfancyoutput}{%
+    % Avoid fatal error with framed.sty if graphics too long to fit on one page
+    \let\sphinxincludegraphics\nbsphinxincludegraphics
+    \nbsphinx@image@maxheight\textheight
+    \advance\nbsphinx@image@maxheight -2\fboxsep   % default \fboxsep 3pt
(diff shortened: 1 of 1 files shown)


src/nbsphinx.py
```diff
@@ -388,6 +388,33 @@
       \unvbox\nbsphinxbox
     \fi
 }% end of \nbsphinxstopnotebook
+
+% Avoid fatal error with framed.sty if graphics too long to fit on one page
+\newdimen\nbsphinx@image@maxheight
+\AtBeginDocument{%
+  \nbsphinx@image@maxheight\textheight
+  % -20pt was found insufficient in testing
+  \advance\nbsphinx@image@maxheight -2.5\baselineskip
```
jfbu: Actually -20pt seems enough with minimal example given in issue #276, but when trying it out earlier with a pure LaTeX example it was not enough, and I opted for 2.5\baselineskip for safety

src/nbsphinx.py
```diff
@@ -388,6 +388,33 @@
...
+      \g@addto@macro\spx@includegraphics@options{,width=\linewidth}%
+      \in@true
+    \fi
+    % no rotation, no need to worry about depth
+    \ifdim \ht\spx@image@box>\nbsphinx@image@maxheight
+      \g@addto@macro\spx@includegraphics@options{,height=\nbsphinx@image@maxheight}%
+      \in@true
+    \fi
+    \ifin@
+      \g@addto@macro\spx@includegraphics@options{,keepaspectratio}%
```
jfbu: I hesitated about this, perhaps we should do it only when height was found too big. Sphinx does not do it it seems when it replaces the width by `\linewidth`.

By the way the code in sphinx.sty (1.8.3) has something strange it clears a scratch box only in one branch, it should do always. And it is quite probable I will backport code like the one here to Sphinx itself.

mgeier:
Thanks a lot!

TBH, I don't understand most of the details, but I think I can roughly follow the general idea ...

src/nbsphinx.py
mgeier: I don't understand the conditions when this line is applied, can you please elaborate?

What is the idea behind this line and what would happen if you would remove it?

What are the alternatives?

jfbu: Idea of fix backported to Sphinx at https://github.com/sphinx-doc/sphinx/commit/6e7e0692836e10eb0b32b6cfa5f7db0e63379302 for upcoming 1.8.4.

This PR is not affected by this, but in future you will not need anymore  `\nbsphinxincludegraphics` but only need to do some `\let\sphinxincludegraphics\sphinxsafeincludegraphics` inside the `nbsphinxfancyoutput` environment with suitable set-up for `\spx@image@maxheight`.

jfbu: if either the included graphic, taking into account the pre-existing `width=`, `height=` options is found too wide a `width=\linewidth` is added, if is found too tall a `height = ...` is added, and if either of these two happen, then this line is executed which ensure that the graphic will keep the same aspect ratio and have both its width and height limited by the new conditions.

jfbu: if you remove this line, the graphic will get distorted because for example the width will be kept but the height will be reduced to the maximal allowed one, or vice-versa.

jfbu: > will keep the same aspect ratio and have both its width and height limited by the new conditions.

I mean by that its intrinsic aspect ratio not the one corresponding to pre-existing width and height options.

mgeier: Thanks for the explanation, that sounds reasonable!

src/nbsphinx.py
```diff
@@ -388,6 +394,28 @@
...
```
jfbu: Hi, a memo if one day nbsphinx requires Sphinx 2.0. Then you won't need the custom LaTeX macro defined here, because Sphinx will use (once https://github.com/sphinx-doc/sphinx/pull/5992 is merged) always a "safe" image inclusion.

src/nbsphinx.py
```diff
@@ -322,6 +322,12 @@
 % Define an environment for non-plain-text code cell outputs (e.g. images)
 \makeatletter
 \newenvironment{nbsphinxfancyoutput}{%
+    % Avoid fatal error with framed.sty if graphics too long to fit on one page
+    \let\sphinxincludegraphics\nbsphinxincludegraphics
+    \nbsphinx@image@maxheight\textheight
+    \advance\nbsphinx@image@maxheight -2\fboxsep   % default \fboxsep 3pt
+    \advance\nbsphinx@image@maxheight -2\fboxrule  % default \fboxrule 0.4pt
+    \advance\nbsphinx@image@maxheight -\baselineskip
```
jfbu: with Sphinx 2.0 (which however will drop Python 2.x support) you can substitute everywhere above `\spx@image@maxheight` for `\nbsphinx@image@maxheight`, you can suppress the `\let` and drop the definition of `\nbsphinxincludegraphics` because `\sphinxincludegraphics` will be alias for its "safe" variant, always checking height of include graphics and reduce it if needed.

jfbu: but if you keep the code here as is, then it will work also with Sphinx 2.0. So this is a remark only for distant future when Sphinx 2.0 will be required by nbsphinx.

PR #302, opened 2019-05-03
Plotly

### LINKED ISSUE #128: Plotly graphs not showing in documentation
Graphs generated using plotly are not showing in the documentation. Matplotlib plots show up just fine.

manycoding:
Closes #128 
@mgeier Please review. I am not sure where to put this javascript in the code. Also, should we include it in all the pages?

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -1754,7 +1754,6 @@ def setup(app):
     except AttributeError:
         _add_notebook_parser(app)
 
-    # require.js for plotly rendering
     app.add_javascript(
         "https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js")
     
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -76,6 +76,12 @@ DISPLAY_DATA_PRIORITY_LATEX = (
 RST_TEMPLATE = """
 {% extends 'rst.tpl' %}
 
+{% block header %}
+.. raw:: html
+
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js"></script>
+
+{% endblock header %}
 
 {% macro insert_empty_lines(text) %}
 {%- set before, after = text | get_empty_lines %}
@@ -1754,8 +1760,8 @@ def setup(app):
     except AttributeError:
         _add_notebook_parser(app)
 
-    app.add_javascript(
-        "https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js")
+    # app.add_javascript(
+    #     "https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js")
     
     app.add_config_value('nbsphinx_execute', 'auto', rebuild='env')
     app.add_config_value('nbsphinx_kernel_name', '', rebuild='env')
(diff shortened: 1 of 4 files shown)


mgeier:
Thanks for this PR!

> Also, should we include it in all the pages?

It would probably be nicer to include it only in pages that are generated from notebooks.
It could probably work in a `raw:: html` directive in `RST_TEMPLATE`?

Is it really necessary to include a `plotly` plot in the docs?

What about testing for require.js in the section https://nbsphinx.readthedocs.io/en/0.4.2/code-cells.html#Arbitrary-JavaScript-Output-(HTML-only)?

We could show there that both jQuery and require.js are available.
While you are at it, you could remove the "Note" box, because the jQuery issue has been fixed in the meantime.

src/nbsphinx.py
```diff
@@ -1754,6 +1754,10 @@ def setup(app):
     except AttributeError:
         _add_notebook_parser(app)
 
+    # require.js for plotly rendering
```
mgeier: This is not just for plotly. No need to mention plotly.

manycoding:
@mgeier I put the script in the `head`, but it doesn't show up. What do I miss?


```
RST_TEMPLATE = """
{% extends 'rst.tpl' %}

{% block head %}
.. raw:: html

    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js"></script>

{% endblock head %}
```

mgeier:
> I put the script in the `head`, but it doesn't show up. What do I miss?

I guess the block is called `header`, see https://github.com/jupyter/nbconvert/blob/220596a04bba2e82f98549dcd3bd49dd235c0a28/nbconvert/templates/skeleton/null.tpl#L24-L25.

> I am not sure if I am understanding why. For me, as an end-user, plotly example belongs in Plots since as an end-user I don't care about JavaScript, I care that my graphs are saved.

Sure, and I would like to make it easy for Plotly users. Those plots should Just Work.

But I want to keep the dependencies (even if it's only the documentation dependencies) as small as possible, and I don't think we actually need a Plotly plot to test if require.js is available.

manycoding:
@mgeier `header` works, put the script is inserted into the body
Screenshot 2019-05-13 at 11 10 59

I agree upon the dependencies, I removed plotly and graph but how can I test if `require.js` present? If I put JavaScript code to check if the script is there, will the script be there during the execution?

mgeier:
> `header` works, put the script is inserted into the body

Yes, that's expected, because the Notebook only provides a part of the page, the actual HTML `<head>` element is controlled by Sphinx.

Would you prefer the script to be loaded in `<head>`?

If yes, I don't know how to do that with Sphinx (other than adding it to all files, not only notebooks).

> but how can I test if `require.js` is in the source?

Good question.

I looked around for a minimal example, but I couldn't find anything concise.

What if we don't change the documentation at all?

After your PR it should work fine, do we really need additional information in the docs?

There are many things which just work even though they are not mentioned in the docs ...

manycoding:
I am ok with both. Let's merge it and see how it goes.

manycoding:
@mgeier let's merge it? The tests seem to fail because one website was unavailable.

mgeier:
@manycoding Sorry for the delay, I've just merged it.

I would like to make the URL configurable at some point, but for now the fixed URL has to be enough.

mgeier:
I've created #327 to make the URL configurable (and optional).

PR #319, opened 2019-09-09
Auto-convert SVG images to PDF

mgeier:
This does not work yet.

For an error log see https://readthedocs.org/projects/nbsphinx/builds/9632595/


mgeier:
It turned out that the problem was with a single SVG file:

bad SVG

For some reason, when converted to PDF, this caused troubles.

I've replaced it with a PNG version, and now it seems to work.

PR #327, opened 2019-09-22
Add option to include require.js

mgeier:


mgeier:
@manycoding This reverts your PR #302, can you please check if this still works for you?

manycoding:
Works like a charm.

mgeier:
@manycoding Thanks for testing!

jorisvandenbossche:
@mgeier Would it be an option to disable this by default? 

I know you can now disable it by setting `nbsphinx_requirejs_path = ''` in conf.py, but including the requirejs breaks websites that use bootstrap, and at least for me, it was not directly easy to find out that it was nbsphinx / injected requirejs that was causing this.

I suppose there will be a way to let requirejs and bootstrap talk nicely together, but not by using it in the default way (including the bootstrap.js in a script tag). So eg websites using sphinx-bootstrap-theme, and both scikit-learn and pandas also recently started to use bootstrap in their development version (it's on the pandas docs that I notices javascript was broken recently, after the nbsphinx update)

mgeier:
> Would it be an option to disable this by default?

In general, sure.

Currently, it is enabled by default because Plotly plots needed it, see #128 and #302.

Also, it is loaded in the Classic Notebook, in JupyterLab and in `nbviewer`, so it seemed to me that people might rely on it being available by default.

> I suppose there will be a way to let requirejs and bootstrap talk nicely together

That would be nice.

> it's on the pandas docs that I notices javascript was broken recently, after the nbsphinx update

Can you please describe the problem?
It would probably be best to open a new issue for that!

There is a version of the `nbsphinx` docs using the bootstrap theme (https://nbsphinx.readthedocs.io/en/bootstrap-theme/), is the problem visible there, too?

mgeier:
I've found the Pandas issues:
https://github.com/pandas-dev/pandas-sphinx-theme/issues/25,
https://github.com/pandas-dev/pandas-sphinx-theme/issues/30.

jorisvandenbossche:
@mgeier Thanks for the answer! I don't directly see the problem happening at https://nbsphinx.readthedocs.io/en/bootstrap-theme/, would need to look more into it to check (but this week on a work week, will try to check next week)

mgeier:
I have the gut feeling that it is related to the bootstrap versions.
Bootstrap 3 works, 4 doesn't. Is that possible?

jorisvandenbossche:
Ah, possibly yes. In any case we are using Bootstrap 4.

PR #334, opened 2019-10-08
Widget documentation

mgeier:
This also makes sure that `require.js` is loaded before the widget JS.

To implement this easily, I've used the `config-inited` event which was introduced in Sphinx 1.8.


mgeier:
HTML preview: https://16-210404706-gh.circle-artifacts.com/0/html/code-cells.html#Interactive-Widgets-(HTML-only)

PDF preview: https://16-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#subsubsection.4.3.8 (but widgets don't work there!)

PR #338, opened 2019-10-12
CircleCI: Add overview page to select between HTML and PDF artifacts

mgeier:


mgeier:
That's the page: https://25-210404706-gh.circle-artifacts.com/0/artifacts.html

PR #344, opened 2019-11-02
Update template name changed in nbconvert development version

takluyver:
The template name is changing in nbconvert 6 - see https://github.com/jupyter/nbconvert/pull/1056 .

~~I don't know if you want to do a hard change like this, or detect the nbconvert version and switch the name as appropriate. But the change breaks nbconvert building its own docs.~~ - EDIT: it now gets the default template name from nbconvert.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
     install_requires=[
         'docutils',
         'jinja2',
-        'nbconvert!=5.4',
+        'nbconvert>=6',
         'traitlets',
         'nbformat',
         'sphinx>=1.8',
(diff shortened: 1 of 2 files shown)


takluyver:
And as soon as I opened the PR, I realised that was a bad way to do it. I've redone it to get the default template name from nbconvert, which should work with old and new nbconvert. If you're happy with that, feel free to squash the commits together.

mgeier:
Thanks for the nicely backwards-compatible fix!

takluyver:
Thanks - I've made a PR for nbconvert to use the development version of nbsphinx (https://github.com/jupyter/nbconvert/pull/1133). You might want to wait until nbconvert 6 is out, just in case the templates get changed again.

takluyver:
(In principle what I did should work even if the name changes again, but you never know what could happen)

PR #348, opened 2019-11-08
Check for ANSI colors in all plain text outputs

### LINKED ISSUE #330: [BUG]: Jupyter help magic unconverted unicode character
When using jupyter help magic (`?`), the output gets messed up, since jyputer internally wraps unicode characters around parts of the output to format it.
I think this is why it is also excluded when setting `nbsphinx_execute = "always"`.
The problem should be solvable with a `re.sub()`.

**Expected/Notebook render:**
expected_output

**Actual render HTML:**
wrong_output

**Actual render PDF:**
wrong_output_pdf

**Json representation of the cell:**

```json
{
   "cell_type": "code",
   "execution_count": 4,
   "metadata": {
    "tags": [
     "nbval-ignore-output"
    ]
   },
   "outputs": [
    {
     "data": {
      "text/plain": [
       "\u001b[1;31mDocstring:\u001b[0m\n",
       "%%bash script magic\n",
       "\n",
       "Run cells with bash in a subprocess.\n",
       "\n",
       "This is a shortcut for `%%script bash`\n",
       "\u001b[1;31mFile:\u001b[0m      c:\\anaconda3\\lib\\site-packages\\ipython\\core\\magics\\script.py\n"
      ]
     },
     "metadata": {},
     "output_type": "display_data"
    }
   ],
   "source": [
    "%%bash?"
   ]
  },
```

 - **OS:** Windows 10
 - **Version:** `nbsphinx==0.4.2`

mgeier:
I removed the name "fancy outputs" because now all outputs use the same mechanism.

Fixes #330.

Preview of the added documentation:

* HTML: https://47-210404706-gh.circle-artifacts.com/0/html/pre-executed.html#Client-specific-Outputs
* PDF: https://47-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#subsubsection.7.1.4

Sadly, there is a problem with LaTeX output:

Page breaks in long code cell outputs with plain text no longer seem to work.
Now there are two nested boxes. Only the outer (framed) box is broken between pages.
If the inner box (created by `sphinxVerbatim`) is longer than a page, I get an error:

    ! TeX capacity exceeded, sorry [input stack size=5000].

@jfbu Do you happen to have an idea how to fix this?

### DIFF
--- a/doc/pre-executed.ipynb
+++ b/doc/pre-executed.ipynb
@@ -148,6 +148,50 @@
    "source": [
     "1 / 0"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Client-specific Outputs\n",
+    "\n",
+    "When `nbsphinx` executes notebooks,\n",
+    "it uses the `nbconvert` module to do so.\n",
+    "Certain Jupyter clients might produce output\n",
+    "that differs from what `nbconvert` would produce.\n",
+    "To preserve those original outputs,\n",
+    "the notebook has to be executed and saved\n",
+    "before running Sphinx.\n",
+    "\n",
+    "For example,\n",
+    "the JupyterLab help system shows the help text as cell outputs,\n",
+    "while executing with `nbconvert` doesn't produce any output."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "\u001b[0;31mSignature:\u001b[0m \u001b[0msorted\u001b[0m\u001b[0;34m(\u001b[0m\u001b[0miterable\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0;34m/\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0;34m*\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0mkey\u001b[0m\u001b[0;34m=\u001b[0m\u001b[0;32mNone\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0mreverse\u001b[0m\u001b[0;34m=\u001b[0m\u001b[0;32mFalse\u001b[0m\u001b[0;34m)\u001b[0m\u001b[0;34m\u001b[0m\u001b[0;34m\u001b[0m\u001b[0m\n",
+       "\u001b[0;31mDocstring:\u001b[0m\n",
+       "Return a new list containing all items from the iterable in ascending order.\n",
+       "\n",
+       "A custom key function can be supplied to customize the sort order, and the\n",
+       "reverse flag can be set to request the result in descending order.\n",
+       "\u001b[0;31mType:\u001b[0m      builtin_function_or_method\n"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "sorted?"
+   ]
   }
  ],
  "metadata": {
@@ -166,7 +210,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.4"
+   "version": "3.7.5rc1"
   },
   "widgets": {
    "state": {},
@@ -174,5 +218,5 @@
   }
  },
  "nbformat": 4,
- "nbformat_minor": 1
+ "nbformat_minor": 4
 }
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -111,19 +111,13 @@ RST_TEMPLATE = """
 {%- if not cell.outputs %}
     :no-output:
 {%- endif %}
-{%- if cell.source.strip() %}
 
 {{ cell.source.strip('\n') | indent }}
-{%- endif %}
 {% endblock input %}
 
 
 {% macro insert_nboutput(datatype, output, cell) -%}
 .. nboutput::
-{%- if datatype == 'text/plain' %}{# nothing #}
-{%- elif datatype == 'ansi' %} ansi
-{%- else %} rst
-{%- endif %}
 {%- if output.output_type == 'execute_result' and cell.execution_count %}
     :execution-count: {{ cell.execution_count }}
 {%- endif %}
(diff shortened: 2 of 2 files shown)


jfbu:
@mgeier one can not nest one within another two usages of "framed" environment, but it is possible to put the inner one in a (non-breakable) minipage. If you add

```
\sphinxverbatimwithminipagetrue
```

to the "begin" part of your `nbsphinxoutput` environment, then the inner sphinxVerbatim will use such a minipage. This could avoid the TeX error you reported (untested). But the inner box becomes unbreakable.

The `nbsphinxoutput` role seems only to use some colours for background and the addition of the prompt? Perhaps you only need to modify the inner sphinxVerbatim way of framing? Perhaps `nbsphinxoutput` rather than issueing its own "framed" could locally redefine what sphinxVerbatim does? this way we avoid nesting two "framed" things. Is there anything inside the `nbsphinxoutput` apart from `sphinxVerbatim`?

For info, here is from LaTeX package [`tcolorbox`](https://ctan.org/pkg/tcolorbox) documentation, which is a much more modern (but adding big dependencies such as TikZ/pgf) package for breakable boxes:

> You can nest an unbreakable tcolorbox inside another tcolorbox, even inside a break- able one. But you cannot not nest a breakable box inside a breakable box. The /tcb/breakable key for a nested box is ignored automatically, i.e. inner boxes are always unbreakable.

in section 18.2 Limitations and Known Bugs.

jfbu:
Looking at the example of nbsphinx own doc I observe that a `sphinxVerbatim` from inside `nbsphinxoutput` is always used with ``\sphinxsetup{verbatimwithframe=false}``, and, except for the ``err`` output with ``VerbatimColor={named}{white}``. For the standard error stream there is however ``VerbatimColor={named}{nbsphinx-stderr}``.

Thus what remains from ``sphinxVerbatim`` usage of ``framed.sty`` commands is almost limited to the "continues on next page", "continued from previous page" thing. A strategy could thus be to transfer this mechanisme to outer `nbsphinxoutput`  environment "output continues on next page", "output continued from previous page", and to castrate the `sphinxVerbatim` to not use "framed". (the verbatim output by itself allows pagebreaks; the usage of "framed" is only to add background color, and a frame which may incorporate a caption and also implements the "continues/continued"  things)

This appears to be feasible but a bit involved. The simplest could possibly be if you could repalce on the fly "sphinxVerbatim" by "nbsphinxVerbatim", and define the "nbsphinxVerbatim" environment as a copied pasted "sphinxVerbatim" but with the pieces related to Framed removed (of course I can help there). It could also be possible to rather hack the environment to locally alter (via some tricks of LaTeX macros) the meaning of `\MakeFramed` etc to become no ops, this is more hacky, although some of it looks easy for example exploiting that there is this in original:

```
  \let\FrameCommand     \sphinxVerbatim@FrameCommand
  \let\FirstFrameCommand\sphinxVerbatim@FirstFrameCommand
  \let\MidFrameCommand  \sphinxVerbatim@MidFrameCommand
  \let\LastFrameCommand \sphinxVerbatim@LastFrameCommand
```

whose effect can esaily be reverted by locally redefining the right hand sides of the assignments.

In the case of standard error stream, will the box contain anything else? if not then background color ``nbsphinx-stderr`` usage can simply get moved to ``nbsphinxoutput`` way of using "framed" commands. Then the above strategy looks feasible.

jfbu:
Sorry about the ``\sphinxverbatimwithminipagetrue`` advice. Although it does work around some difficulties with the nesting of framed environment, it can not help (in fact it aggravates it) avoid a problem which breaks PDF build under certain circumstances https://github.com/sphinx-doc/sphinx/issues/6804

Best is really to try to avoid nesting "framed" environments... but as you see from Sphinx ticket I referred to, this is not always easy.

mgeier:
@jfbu Thanks a lot for your help!

I tried your suggestion and used `Verbatim` instead of `sphinxVerbatim`:  #350.
This basically works, but the verbatim lines are too far apart.
Do you know how to fix this?

If this doesn't work out, I'll have to use `sphinxVerbatim` again, but without the outer "framed" environment.
I haven't tried that yet, I don't know which of the two options is more promising.

jfbu:
I have a proof of concept of an adapted `sphinxVerbatim` which can be nested a [there](https://github.com/jfbu/sphinx/commit/5c7277e2e2183b0a4aa273b37cf157526d9cb432)

What this does in case of nesting (which here is always the case) is that it measures the length of the code-block and if it is not too close to the textheight it uses the normal `sphinxVerbatim` (but in a minipage, hence unbreakable), else it uses castrated `sphinxVerbatim` i.e. without framing. One then keeps things such as line numbers or highlighted lines but one loses the caption if there was one (this is provisory).

For better helping I would need to get a feeling for what can be there in `nbsphinxoutput`:  multiple `sphinxVerbatim`'s or only one? Is it a problem if individual `sphinxVerbatim` lose background color`? can they potentially come with some caption?

My inclination is that "continues/continued" hints should be moved over from `sphinxVerbatim` to `nbsphinxoutput` and indeed `sphinxVerbatim` therein should be a version not using `\MakeFramed`. The latter can be achieved in various ways.

jfbu:
For using [the WIP on sphinx-doc/sphinx#6804](https://github.com/jfbu/sphinx/commit/5c7277e2e2183b0a4aa273b37cf157526d9cb432) in the context here, the `nbsphinxoutput` environment must issue `\makeatletter\spx@inframedtrue\makeatother`

but I think it is better altogether indeed if nested `sphinxVerbatim` simply is one without usage of `\MakeFramed`. I think ultimately at sphinx-doc/sphinx I will have to define two environments one using "framed" the other not using it (one needs two, because the mechanism for the caption will have to be different; it may be possible to keep using a single env I have to think and consider backwards compatibility)

jfbu:
Although I think you are moving towards not nesting things (which if you can, is indeed the best), I point out that my [WIP branch](https://github.com/sphinx-doc/sphinx/compare/2.0...jfbu:6804_long_code_block_in_danger) on fixing sphinx-doc/sphinx#6804 is probably done. With this version of sphinxVerbatim you would not have encountered the PDF crash error here.

mgeier:
I think #351 is a better alternative, but I still want to answer your questions:

> For better helping I would need to get a feeling for what can be there in `nbsphinxoutput`:  multiple `sphinxVerbatim`'s or only one?

Only one.

> Is it a problem if individual `sphinxVerbatim` lose background color`?

Outputs containing `stderr` should have a special background color.
But it doesn't matter if the outer framed environment or the verbatim environment has that background color.

> can they potentially come with some caption?

No. Unless you count the prompt as a caption.

> My inclination is that "continues/continued" hints should be moved over from `sphinxVerbatim` to `nbsphinxoutput` and indeed `sphinxVerbatim` therein should be a version not using `\MakeFramed`.

Yes, that sounds reasonable.
But probably completely avoiding the nested environments is even better, like I've done in #351.

mgeier:
Sorry, I've just missed your latest comment when I posted my comment above ...

I will try your WIP branch.

mgeier:
OK, I tried your branch and when I add `\spx@inframedtrue` to the `nbsphinxoutput` environment, it works!
There is no crash and page breaks happen correctly.

Of course this would only work in future Sphinx versions.

Is there a reason to pursue this PR, or is it better to use #351 instead?

I guess based on your comment https://github.com/spatialaudio/nbsphinx/pull/351#issuecomment-552187755 I should use #351, right?

jfbu:
Yes, I agree #351 which avoids nesting things is definitly to be preferred. Although future Sphinx avoids fatal errors if long code-blocks get nested in another "framed" thing, it is still a bit bad that they lose background color generally configured for Verbatim and have no way of inserting the continuation hints across pages.

Thanks by the way for confirmation my WIP appears to work in a systematic nested context which is thus a good test, and for taking the time answering my questions above! (this helps me also for future!)

jfbu:
> My inclination is that "continues/continued" hints should be moved over from `sphinxVerbatim` to `nbsphinxoutput` and indeed `sphinxVerbatim` therein should be a version not using `\MakeFramed`. The latter can be achieved in various ways.

Hi @mgeier I feel bad about adding much entropy at times. I had not looked for a long time at your code and I wrote the above lines without much of a check. Actually the `nbsphinxoutput` here (or `nbsphinxfancyoutput` at #351) already does have the needed set-up for the continuation hints reusing the Sphinx macros. Thus all my (repeated) talk about moving these things was non-sense. For some reason I was under impression a plain `\MakeFramed` was used. Sorry for confused talk.

mgeier:
@jfbu Don't worry, I'm very grateful for all your help!

I'll close this PR now, because it has been superseded by #351.

PR #349, opened 2019-11-09
Fix css for 'sphinx_copybutton'

### LINKED ISSUE #333: The sphinx_copybutton extension doesn't properly work out-of-the-box
The extension adds "copy" buttons to all code areas, including Jupyter code cells, see https://github.com/choldgraf/sphinx-copybutton.

Some CSS to work around the issue is shown in https://github.com/spatialaudio/nbsphinx/issues/330#issuecomment-538772288, but it would certainly be nice to make this work out-of-the-box.

Currently, there are several problems:

* the copy buttons are also shown for the prompts, which is quite useless
* the distance from the top and right edge of the code cell is too large (and much larger than it is in "normal" code blocks)
* in Chrome, scrollbars appear on code cells with a single line

Firefox:

image

Chrome:

image

s-weigand:
Changed CSS so 'sphinx_copybutton' will look properly out of the box.
- Moved `padding: 0.4em;` from `.input_area, .output_area` to its child `pre`
- Added `padding: 0.4em;` to other `.input_area, .output_area` child elements, to keep the styling consistent
- Added `sphinx_copybutton` to the docs
- Tested output for `alabaster` and `sphinx_rtd_theme`

**Before:**
copybtn_before

**After:**
copybtn_after

**After sphinx_rtd_theme:**
copybtn_after_rtd

closes #333

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ dist/
 nbsphinx.egg-info/
 .ipynb_checkpoints/
 .python-version
+.vscode
+doc/_build
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,11 @@ matrix:
   - env: SPHINX="==2.2.0"
 
   # a few Python versions using latest Sphinx release
-  - python: "3.4"
-    env: BIBTEX="==0.4.2"
   - python: "3.5"
   - python: "3.6"
   - python: "3.7"
-  - python: "nightly"
-  #- python: "pypy3"
+  - python: "3.8"
+    # - python: "nightly"
 
   - name: Python 2.7 + latest Sphinx
     python: "2.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
       - python2 -m pip install sphinxcontrib-bibtex==0.4.2
       - python2 -m pip install sphinxcontrib-svg2pdfconverter
       - python2 -m pip install ipywidgets
-      - python2 -m pip install sphinx-copybutton
+      - python2 -m pip install 'sphinx-copybutton<0.2.6'
     script: &py2-script
       - python2 -m nbsphinx
       # We don't execute, because the example notebooks use Python 3
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ matrix:
   - python: "3.5"
   - python: "3.6"
   - python: "3.7"
-  - python: "3.8"
-    # - python: "nightly"
+  - python: "nightly"
+  # - python: "pypy3"
 
   - name: Python 2.7 + latest Sphinx
     python: "2.7"
@@ -52,6 +52,9 @@ matrix:
           - python2
     install: *py2-install
     script: *py2-script
+
+  allow_failures:
+    - python: "nightly"
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ matrix:
       - python2 -m pip install sphinxcontrib-bibtex==0.4.2
       - python2 -m pip install sphinxcontrib-svg2pdfconverter
       - python2 -m pip install ipywidgets
+      - python2 -m pip install sphinx-copybutton
     script: &py2-script
       - python2 -m nbsphinx
       # We don't execute, because the example notebooks use Python 3
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -7,7 +7,7 @@
 extensions = [
     'nbsphinx',
     'sphinx_copybutton',
-    'sphinx.ext.mathjax',  # for math equation
+    'sphinx.ext.mathjax',  # for math equations
     'sphinxcontrib.bibtex',  # for bibliographic references
     'sphinxcontrib.rsvgconverter',  # for SVG->PDF conversion in LaTeX output
 ]
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -6,7 +6,8 @@
 # Select nbsphinx and, if needed, other Sphinx extensions:
 extensions = [
     'nbsphinx',
-    'sphinx.ext.mathjax',  # for math equations
+    'sphinx_copybutton',
+    'sphinx.ext.mathjax',  # for math equation
     'sphinxcontrib.bibtex',  # for bibliographic references
     'sphinxcontrib.rsvgconverter',  # for SVG->PDF conversion in LaTeX output
 ]
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -621,6 +621,7 @@ div.nboutput.container div.output_area > img{
 }
 
 div.nboutput.container div.output_area > div[class*=highlight],
+div.nboutput.container div.output_area.rendered_html > img{
     padding: 0;
 }
 
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -178,7 +178,7 @@ RST_TEMPLATE = """
 
     .. raw:: html
 
-        <div class="js-render-output"></div>
+        <div class="output_javascript"></div>
         <script type="text/javascript">
         var element = document.currentScript.previousSibling.previousSibling;
 {{ output.data['application/javascript'] | indent | indent }}
@@ -615,7 +615,7 @@ div.nbinput.container div.input_area div[class*=highlight] > pre,
 div.nboutput.container div.output_area div[class*=highlight] > pre,
 div.nboutput.container div.output_area div[class*=highlight].math,
 div.nboutput.container div.output_area.rendered_html,
-div.nboutput.container div.output_area > div.js-render-output,
+div.nboutput.container div.output_area > div.output_javascript,
 div.nboutput.container div.output_area > img{
     padding: 0.4rem;
 }
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -611,7 +611,6 @@ div.rendered_html tbody tr:hover {
   background: rgba(66, 165, 245, 0.2);
 }
 
-/* set the padding for the pre so copybtn icon won't cause an overflow */
 div.nbinput.container > div.input_area > div[class*=highlight] > pre,
 div.nboutput.container > div.output_area > div[class*=highlight] > pre,
 div.nboutput.container > div.output_area > div[class*=highlight].math,
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -636,11 +636,6 @@ CSS_STRING_READTHEDOCS = """
     margin-bottom: 19px;  /* padding has already 5px */
 }
 
-/* fix copybtn icon positioning (needed for 'sphinx_copybutton') */
-a.copybtn > img{
-    vertical-align: top;
-}
-
 /* ... except between code cells! */
 .nblast.container + .nbinput.container {
     margin-top: -19px;
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -479,8 +479,8 @@ div.nboutput.container div.prompt pre {
 div.nbinput.container div.prompt,
 div.nboutput.container div.prompt {
     min-width: %(nbsphinx_prompt_width)s;
-    padding-top: 0.4em;
-    padding-right: 0.4em;
+    padding-top: 0.3rem;
+    padding-right: 0.3rem;
     text-align: right;
     flex: 0;
 }
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -587,10 +587,6 @@ div.nboutput.container div.output_area:not(.rendered_html) > img{
     padding: 0.4rem;
 }
 
-div.nboutput.container div.output_area > div[class*=highlight]{
-    padding: 0;
-}
-
 /* hide copybtn icon on prompts (needed for 'sphinx_copybutton') */
 .prompt a.copybtn {
     display: none;
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -625,7 +625,7 @@ div.nboutput.container div.output_area.rendered_html > img{
     padding: 0;
 }
 
-/* fix positioning of copybtn icon */
+/* fix positioning of copybtn icon from 'sphinx_copybutton' */
 a.copybtn {
     padding: 0;
     top: 0.1em;
@@ -634,7 +634,7 @@ a.copybtn > img{
     padding: 0;
 }
 
-/* hide copybtn icon on prompts */
+/* hide copybtn icon on prompts from 'sphinx_copybutton' */
 .prompt.highlight-none.notranslate a.copybtn {
     display: none;
 }
@@ -650,10 +650,14 @@ CSS_STRING_READTHEDOCS = """
     margin-bottom: 19px;  /* padding has already 5px */
 }
 
-/* fix copybtn overflow problem */
+/* fix copybtn from overflow problem 'sphinx_copybutton' */
 div.nbinput.container div.input_area > div[class^='highlight'],
 div.nboutput.container div.output_area > div[class^='highlight']{
     overflow-y: hidden;
+}
+/* fix copybtn icon positioning from 'sphinx_copybutton' */
+a.copybtn > img{
+    vertical-align: top;
 }
 
 /* ... except between code cells! */
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -584,7 +584,7 @@ div.nboutput.container div.output_area div[class*=highlight].math,
 div.nboutput.container div.output_area.rendered_html,
 div.nboutput.container div.output_area > div.output_javascript,
 div.nboutput.container div.output_area:not(.rendered_html) > img{
-    padding: 0.4rem;
+    padding: 0.3rem;
 }
 
 /* hide copybtn icon on prompts (needed for 'sphinx_copybutton') */
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -577,6 +577,25 @@ div.nboutput.container div.output_area.stderr {
 .ansi-bold { font-weight: bold; }
 .ansi-underline { text-decoration: underline; }
 
+
+div.nbinput.container div.input_area div[class*=highlight] > pre,
+div.nboutput.container div.output_area div[class*=highlight] > pre,
+div.nboutput.container div.output_area div[class*=highlight].math,
+div.nboutput.container div.output_area.rendered_html,
+div.nboutput.container div.output_area > div.output_javascript,
+div.nboutput.container div.output_area:not(.rendered_html) > img{
+    padding: 0.4rem;
+}
+
+div.nboutput.container div.output_area > div[class*=highlight]{
+    padding: 0;
+}
+
+/* hide copybtn icon on prompts from 'sphinx_copybutton' */
+.prompt a.copybtn {
+    display: none;
+}
+
 /* Some additional styling taken form the Jupyter notebook CSS */
 div.rendered_html table {
   border: none;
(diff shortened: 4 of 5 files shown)


src/nbsphinx.py
```diff
@@ -611,6 +610,28 @@
...
   background: rgba(66, 165, 245, 0.2);
 }
+
+/* set the padding for the pre so copybtn icon won't cause an overflow */
+div.nbinput.container div.input_area div[class*=highlight] pre,
+div.nboutput.container div.output_area div[class*=highlight] pre,
+div.nboutput.container div.output_area.rendered_html,
+div.nboutput.container div.output_area div.widget-subarea,
+div.nboutput.container div.output_area div.math,
+div.nboutput.container div.output_area > img{
```
mgeier: Shouldn't the last 3 lines all have `>` in them?

What if there are nested `div.widget-subarea` or `div.math` elements?

I don't really know if that can happen, but it wouldn't hurt, would it?

mgeier:
Thanks a lot for this PR, and sorry for the late reply!

As long as the extension is not properly supported on most themes (or at least on the RTD theme), I hesitate to enable it in the `nbsphinx` docs (because they are available for many themes).

What about just adding the CSS fixes?

s-weigand:
Some remarks after working on this:
1. Changing styles without changing styles and by hand doing side by side compares is a tedious job and I found [this service](https://percy.io/) helpfull for the job (5k screenshot compares/month free for open source and you could just upload the full pages from the docs).
2. IMHO it would be more handy to move the css to an actual `*.css` file for linting + autocomplete and just read it to a string in the package (same would be good for the RST and LaTeX templates).

Just some ideas 😄

s-weigand:
Hmmm looks like 'Python 2.7 + latest Sphinx' has some problems with `hashlib` on mac :dizzy_face:

src/nbsphinx.py
s-weigand: I changed it and went even further, by making all elements direct child elements, this should potentially mess up even less.

mgeier:
I've fixed the Travis-CI failure in #363.

src/nbsphinx.py
mgeier: The problem is that this is inconsistent with the rest of the CSS rules.

If you think this is necessary, you should change it for all CSS rules in a separate PR.

I think it is only really necessary for such general classes like `img` or `div.math` which could actually appear somewhere nested in user-specified HTML output.

Of course, theoretically, there could be e.g. a nested `div.output_area` in some user-provided HTML ... but I guess this is a topic for a separate PR.

mgeier:
> Changing styles without changing styles and by hand doing side by side compares is a tedious job and I found [this service](https://percy.io/) helpfull for the job (5k screenshot compares/month free for open source and you could just upload the full pages from the docs).

I'm only open to free-and-unlimited-for-open-source services.

I don't like if there is a hard limit. Though admittedly, 5k seems quite a lot.

Anyway, I have no clue how much work it would be to set something like this up.

> IMHO it would be more handy to move the css to an actual `*.css` file for linting + autocomplete and just read it to a string in the package (same would be good for the RST and LaTeX templates).

Yes, I agree.
The files don't even have to be read into a string, there should be officially supported ways to include those file in the Sphinx build process. I didn't really look into that, though.

I started with everything in one file, when everything was still small. Now I think it is a good idea to make separate files at some point, I just didn't really have the motivation to do so.

If you like, you can make a PR!

src/nbsphinx.py
```diff
@@ -611,6 +610,37 @@
 div.rendered_html tbody tr:hover {
   background: rgba(66, 165, 245, 0.2);
 }
+
+/* set the padding for the pre so copybtn icon won't cause an overflow */
```
mgeier: I don't think this comment is necessary, or is it?

src/nbsphinx.py
```diff
@@ -611,6 +610,37 @@
...
+div.nboutput.container > div.output_area > div,
+div.nboutput.container > div.output_area > img{
+    padding: 0.4em;
+}
+
+div.nboutput.container > div.output_area > div[class*=highlight]{
+    padding: 0;
+}
+
+div.nboutput.container > div.output_area > div.admonition{
```
mgeier: This can only happen in the case of unknown MIME type, right?

s-weigand: This is needed for [Info/Warning Boxes](https://nbsphinx.readthedocs.io/en/0.5.0/markdown-cells.html#Info/Warning-Boxes) to look the same as before (only tested for alabaster), since the style got changed by `div.nboutput.container > div.output_area > div,`, which again is needed that [javascript output](https://nbsphinx.readthedocs.io/en/0.5.0/code-cells.html#Arbitrary-JavaScript-Output-(HTML-only)) is properly indented. Admittedly the better solution would be to give the js output a class, which would allow a less general selector to be used.

src/nbsphinx.py
s-weigand: I admit, this won't be usefull for someone that looks in the CSS and doesn't know of the current CSS problems with copybtn, so I will remove it.

src/nbsphinx.py
mgeier: Then we should do that, shouldn't we?

It seems to be very simple, we can add a CSS class there:

https://github.com/spatialaudio/nbsphinx/blob/0014a286eaa15115ed5eaf2422df939912f0383c/src/nbsphinx.py#L181

`nbconvert` uses the class name `output_javascript`, so we could also use that.

src/nbsphinx.py
s-weigand: I changed the style back, so it is more homogeneous.

s-weigand:
Hmmm [repology](https://repology.org/badge/vertical-allrepos/python:nbsphinx.svg) is down atm, which is why the tests fail.
With the CSS as is atm, the style of [some warning boxes](https://nbsphinx.readthedocs.io/en/0.5.0/code-cells.html#Unsupported-Output-Types) changes, but I can't reproduce the warnings.
```shell
$ jupyter lab --version
1.2.4
```

mgeier:
Repology seems to be up again (https://github.com/repology/repology-webapp/issues/101)!

> With the CSS as is atm, the style of [some warning boxes](https://nbsphinx.readthedocs.io/en/0.5.0/code-cells.html#Unsupported-Output-Types) changes, but I can't reproduce the warnings.

That's not a problem.

JupyterLab and the Classic Notebook don't display anything in this case, but `nbsphinx` does.
IMHO it's nice to see that something it wrong. But it will happen quite rarely and the styling of these warnings is not important.

.travis.yml
```diff
@@ -30,6 +30,7 @@ matrix:
       - python2 -m pip install sphinxcontrib-bibtex==0.4.2
       - python2 -m pip install sphinxcontrib-svg2pdfconverter
       - python2 -m pip install ipywidgets
+      - python2 -m pip install 'sphinx-copybutton<0.2.6'
```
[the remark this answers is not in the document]
s-weigand: `sphinx-copybutton`droped python 2 with version 0.2.6, but didn't add the proper `python_requires` I guess, so python 2 tries to install it and fails due to python3 only syntax.
But you wanted to drop python2 with the gh actions PR anyway, due to the end of support in january.

.gitignore
```diff
@@ -5,3 +5,5 @@ dist/
 nbsphinx.egg-info/
 .ipynb_checkpoints/
 .python-version
+.vscode
```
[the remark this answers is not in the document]
s-weigand: This is just the config folder from vscode, I think there is no harm in having it in the `.gitignore`.
I wouldn't want to add it to my global config, since I also have repos with project specific `.vscode` configs. This also ensures that other users won't upload their config if they would i.e. make a `git add .`

src/nbsphinx.py
```diff
@@ -611,6 +610,35 @@
...
+div.nboutput.container div.output_area div[class*=highlight].math,
+div.nboutput.container div.output_area.rendered_html,
+div.nboutput.container div.output_area > div.output_javascript,
+div.nboutput.container div.output_area > img{
+    padding: 0.4rem;
+}
+
+div.nboutput.container div.output_area > div[class*=highlight],
+div.nboutput.container div.output_area.rendered_html > img{
+    padding: 0;
```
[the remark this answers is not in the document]
s-weigand: The `.rendered_html > img` is the jupyter logo in `[12]`  [above this](https://nbsphinx.readthedocs.io/en/0.5.0/code-cells.html#Math).

I think this `div.highlight` was for the last [ANSI color example](https://nbsphinx.readthedocs.io/en/0.5.0/code-cells.html#ANSI-Colors).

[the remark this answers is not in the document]
mgeier: Oh, I missed that!

What about avoiding the `img` part in the first place with something like ...

```css
div.nboutput.container div.output_area:not(.rendered_html) > img {
    padding: 0.4rem;
}
```

... in the rule above that?

For the first case, I don't see a difference in the ANSI output cells.

src/nbsphinx.py
```diff
@@ -611,6 +610,35 @@
...
+}
+
+/* fix positioning of copybtn icon */
```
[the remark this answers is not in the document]
s-weigand: Sure thing, thats a good idea.

src/nbsphinx.py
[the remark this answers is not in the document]
s-weigand: The `:not(.rendered_html)` did work fine 😄 
For the ANSI output, I can leave that rule out, I'm just trying to change things without changing the appearence and I saw that is was slightly different (I use Chome).
So just tell me if you want to keep it or not.

src/nbsphinx.py
```diff
@@ -611,6 +610,35 @@
...
+a.copybtn {
+    padding: 0;
```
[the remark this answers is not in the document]
s-weigand: This basically got an artifact from my hotfix, since back then when `. input_area, .output_area` had the `padding`, `a. copybtn` did inherit it (I think, too lazy to roll all back to check it out).Anyway, it looks fine without the rules, so I removed them.

src/nbsphinx.py
```diff
@@ -611,6 +610,35 @@
...
+    top: 0.1em;
```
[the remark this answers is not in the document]
s-weigand: see https://github.com/spatialaudio/nbsphinx/pull/349#discussion_r360508071

src/nbsphinx.py
```diff
@@ -611,6 +610,35 @@
...
+}
+a.copybtn > img{
+    padding: 0;
```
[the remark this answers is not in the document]
s-weigand: see https://github.com/spatialaudio/nbsphinx/pull/349#discussion_r360508071

src/nbsphinx.py
```diff
@@ -611,6 +610,35 @@
...
+}
+
+/* hide copybtn icon on prompts */
+.prompt.highlight-none.notranslate a.copybtn {
```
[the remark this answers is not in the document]
s-weigand: Yap you are right, I simplified it.

src/nbsphinx.py
```diff
@@ -611,6 +610,35 @@
 div.rendered_html tbody tr:hover {
   background: rgba(66, 165, 245, 0.2);
 }
+
```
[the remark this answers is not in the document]
s-weigand: Done ^^

src/nbsphinx.py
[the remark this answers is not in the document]
mgeier: Thanks! If the change is barely noticeable when switching, I don't really care. I prefer more concise and understandable CSS rules.

src/nbsphinx.py
```diff
@@ -622,6 +640,16 @@
     margin-bottom: 19px;  /* padding has already 5px */
 }
 
+/* fix copybtn overflow problem (needed for 'sphinx_copybutton') */
```
mgeier: What was the problem here?
I cannot reproduce ...

mgeier:
Thanks for the updates!

Please note my attempted CSS changes there: https://github.com/choldgraf/sphinx-copybutton/pull/45

src/nbsphinx.py
```diff
@@ -622,6 +640,16 @@
...
+div.nbinput.container div.input_area > div[class^='highlight'],
+div.nboutput.container div.output_area > div[class^='highlight']{
+    overflow-y: hidden;
+}
+/* fix copybtn icon positioning (needed for 'sphinx_copybutton') */
+a.copybtn > img{
+    vertical-align: top;
```
mgeier: Is this a bug fix?

If yes, you should propose it at https://github.com/choldgraf/sphinx-copybutton/pulls

src/nbsphinx.py
s-weigand: Due to some other changes this got obsolete.

src/nbsphinx.py
```diff
@@ -578,6 +577,25 @@
...
+div.nboutput.container div.output_area div[class*=highlight] > pre,
+div.nboutput.container div.output_area div[class*=highlight].math,
+div.nboutput.container div.output_area.rendered_html,
+div.nboutput.container div.output_area > div.output_javascript,
+div.nboutput.container div.output_area:not(.rendered_html) > img{
+    padding: 0.4rem;
+}
+
+div.nboutput.container div.output_area > div[class*=highlight]{
+    padding: 0;
```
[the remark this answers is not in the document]
s-weigand: This was for the ANSI stuff see https://github.com/spatialaudio/nbsphinx/pull/349#discussion_r360114969 . I removed it now.

src/nbsphinx.py
s-weigand: With the changes from https://github.com/choldgraf/sphinx-copybutton/pull/45 the change is barely noticeable, compared to the default. But this change was so that the distances of the image/icon to the top and right border are equal (RTD theme). If you want I can remove it and/or you might want to add it to your PR.

mgeier: OK, I've added it to my PR, see https://github.com/choldgraf/sphinx-copybutton/pull/45#issuecomment-569226728.

s-weigand: GZ for the merge of your PR, I removed the rule here ^^

doc/conf.py
```diff
@@ -6,7 +6,8 @@
 # Select nbsphinx and, if needed, other Sphinx extensions:
 extensions = [
     'nbsphinx',
-    'sphinx.ext.mathjax',  # for math equations
+    'sphinx_copybutton',
+    'sphinx.ext.mathjax',  # for math equation
```
mgeier: Is there a reason you changed equations -> equation?

src/nbsphinx.py
```diff
@@ -578,6 +577,21 @@
...
 .ansi-underline { text-decoration: underline; }
 
+
+div.nbinput.container div.input_area div[class*=highlight] > pre,
+div.nboutput.container div.output_area div[class*=highlight] > pre,
+div.nboutput.container div.output_area div[class*=highlight].math,
+div.nboutput.container div.output_area.rendered_html,
+div.nboutput.container div.output_area > div.output_javascript,
+div.nboutput.container div.output_area:not(.rendered_html) > img{
+    padding: 0.4rem;
```
mgeier: Just out of curiosity:
Is there a difference between `em` and `rem` in this very case?

doc/conf.py
s-weigand: There is no reason, I must have slipped on the keyboard. :confounded:
Fixed it!

src/nbsphinx.py
s-weigand: The difference is for the RTD Theme since they use:
```css
.rst-content pre.literal-block, .rst-content div[class^='highlight'] pre, .rst-content .linenodiv pre {
    font-family: SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",Courier,monospace;
    font-size: 12px;
    line-height: 1.4;
}
```
So for consistency's sake, for possible other themes as well, I decided to use `rem` instead of `em` for all of them.

mgeier: Thanks, that's interesting!

I've just tried the "material" theme and there is also a difference.

However, if you change it here, you'll have to change the three other occurences of `0.4em`, too. Otherwise, the prompts are at the wrong position.

And I think `0.4rem` look quite large, what about using `0.3rem` instead?

s-weigand: I think tweaking for individual themes should be done on a per theme base.
What are your thoughts about this?

mgeier: Sure, but we need to choose one default value. And I think `0.3rem` looks acceptable in all themes (that I tried). I don't think we need per-theme settings here (for now).

s-weigand: I  changed the padding to `0.3rem` and re implemented the overflow work-around.

mgeier:
Thanks!

There's still a problem with the placement of prompts, see https://github.com/spatialaudio/nbsphinx/pull/349#discussion_r362048154.

src/nbsphinx.py
```diff
@@ -640,6 +654,12 @@
 CSS_STRING_CLOUD = """
 /* CSS overrides for cloud theme */
 
+/* fix copybtn overflow problem (needed for 'sphinx_copybutton') */
+div.nbinput.container div.input_area > div[class^='highlight'],
+div.nboutput.container div.output_area > div[class^='highlight']{
+    overflow-y: hidden;
+}
```
mgeier: Would it hurt adding this for all themes?

This problem also appears (only on Chromium?) for the "material" theme, and probably for other themes as well.

s-weigand: I moved it up to general rules. 
I'm not too happy with that workaround, since it might potentially break the functionality of a theme where i.e. the max height is limited and a scrollbar is wanted. 
But on the other hand, if there isn't a use case where that is the case, it's just 'worrying about unlaid eggs'.

mgeier: IMHO, themes should not limit the max height (and I think no existing theme does).

But if they really, really want to, they can simply re-introduce the scrollbar with their CSS.

Anyway, none of this is set in stone, we can change it later if we feel like it. But for now, I think this is good.

mgeier:
Thanks for the update! Now the only thing that's missing is setting the prompt padding to `0.3rem`.

.travis.yml
```diff
@@ -14,13 +14,11 @@ matrix:
...
 
   # a few Python versions using latest Sphinx release
-  - python: "3.4"
-    env: BIBTEX="==0.4.2"
   - python: "3.5"
   - python: "3.6"
   - python: "3.7"
-  - python: "nightly"
-  #- python: "pypy3"
+  - python: "3.8"
```
mgeier: You don't need to add a job with Python 3.8, since that's the default (see line 3) which is used in line 7 already.

.travis.yml
```diff
@@ -14,13 +14,11 @@ matrix:
...
+    # - python: "nightly"
```
mgeier: I'd like to keep "nightly". I'm fine with ignoring it if it fails for an unrelated reason (as I guess it currently does).

.travis.yml
```diff
@@ -14,13 +14,11 @@ matrix:
   - env: SPHINX="==2.2.0"
 
   # a few Python versions using latest Sphinx release
-  - python: "3.4"
-    env: BIBTEX="==0.4.2"
```
mgeier: That's good, we don't need Python 3.4!

.travis.yml
```diff
@@ -14,13 +14,11 @@ matrix:
...
   - python: "3.5"
   - python: "3.6"
   - python: "3.7"
-  - python: "nightly"
-  #- python: "pypy3"
```
mgeier: I'd like to keep that comment as a reminder. I hope at some point it will work on PyPy.

.travis.yml
s-weigand: True that, I did overlook 'default Python + latest Sphinx'

.travis.yml
s-weigand: I added 'nightly' to `allow_failures`.

mgeier:
Thanks a lot for your work!

s-weigand:
Thanks for all the detailed reviews 😄 
Next I might look into pulling the template strings into separate files, with the proper extension.

mgeier:
Cool, I'm looking forward to this!

BTW, I've re-based all theme branches to see if the copy button works.
See https://nbsphinx.readthedocs.io/en/latest/usage.html#HTML-Themes for a list of themes.

I've found some issues in a few themes: https://github.com/spatialaudio/nbsphinx/issues/376

But all in all it looks great in most themes!

src/nbsphinx.py
NovaToro: I will work on it how do I fix cold factor

PR #350, opened 2019-11-09
Ansi alternative

mgeier:
Possible alternative to #348.

This is using `Verbatim` instead of `sphinxVerbatim`.

This is not yet complete.

I still have to set the background color on `stderr` cells.

The spacing inside the plain-text output cells is probably not quite correct.

But most importantly, the plain text lines are too far apart!

This can be e.g. be seen in the ANSI outputs: https://49-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#subsection.4.4

### DIFF
--- a/doc/pre-executed.ipynb
+++ b/doc/pre-executed.ipynb
@@ -148,6 +148,50 @@
    "source": [
     "1 / 0"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Client-specific Outputs\n",
+    "\n",
+    "When `nbsphinx` executes notebooks,\n",
+    "it uses the `nbconvert` module to do so.\n",
+    "Certain Jupyter clients might produce output\n",
+    "that differs from what `nbconvert` would produce.\n",
+    "To preserve those original outputs,\n",
+    "the notebook has to be executed and saved\n",
+    "before running Sphinx.\n",
+    "\n",
+    "For example,\n",
+    "the JupyterLab help system shows the help text as cell outputs,\n",
+    "while executing with `nbconvert` doesn't produce any output."
+   ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


jfbu:
To fix the extra whitespace:

```diff
diff --git a/src/nbsphinx.py b/src/nbsphinx.py
index 9695816..5c1453b 100644
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -139,8 +139,7 @@ RST_TEMPLATE = """
 
     .. raw:: latex
 
-        \\kern-\\sphinxverbatimsmallskipamount
-        \\kern+\\FrameHeightAdjust\\kern-\\fboxrule
+        \\lineskip0pt\\relax
         \\begin{Verbatim}[commandchars=\\\\\\{\\}]
 {{ output.data[datatype] | escape_latex | ansi2latex | indent | indent }}
         \\end{Verbatim}
```

But I think you are losing things by dropping sphinxVerbatim for Verbatim: long code lines will not wrap, the mechanism for highlighted lines is lost and (but does it apply here?) you lose the caption.

mgeier:
Thanks, I've added the suggested fix: bd41d5c2add1737dab053d3e7a84985861c41cf8

Here's a preview: https://52-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#subsection.4.4

> long code lines will not wrap

Oh, that's bad, I like the line wrapping ...

> the mechanism for highlighted lines is lost

This is not needed here.

> you lose the caption

This isn't needed either

> also, Sphinx adds a `\strut` in each codeline to help keep a consistent baseline distance

Indeed, that looks bad.

It looks like this isn't the right way to proceed ... therefore I've tried again: #351
This avoids the nested framed environments and AFAICT it works well!

mgeier:
Superseded by #351.

PR #351, opened 2019-11-10
Check for ANSI colors in all plain text outputs

mgeier:
This is yet another alternative to #348.

It makes somewhat less changes, in particular it keeps the separation between "fancy" and "non-fancy" output cells (with regards to LaTeX).

But like #348, it removes the separation between "ANSI" and "text/plain" outputs.

Here's a PDF preview: https://51-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#section.4

### DIFF
--- a/doc/pre-executed.ipynb
+++ b/doc/pre-executed.ipynb
@@ -148,6 +148,50 @@
    "source": [
     "1 / 0"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Client-specific Outputs\n",
+    "\n",
+    "When `nbsphinx` executes notebooks,\n",
+    "it uses the `nbconvert` module to do so.\n",
+    "Certain Jupyter clients might produce output\n",
+    "that differs from what `nbconvert` would produce.\n",
+    "To preserve those original outputs,\n",
+    "the notebook has to be executed and saved\n",
+    "before running Sphinx.\n",
+    "\n",
+    "For example,\n",
+    "the JupyterLab help system shows the help text as cell outputs,\n",
+    "while executing with `nbconvert` doesn't produce any output."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "\u001b[0;31mSignature:\u001b[0m \u001b[0msorted\u001b[0m\u001b[0;34m(\u001b[0m\u001b[0miterable\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0;34m/\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0;34m*\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0mkey\u001b[0m\u001b[0;34m=\u001b[0m\u001b[0;32mNone\u001b[0m\u001b[0;34m,\u001b[0m \u001b[0mreverse\u001b[0m\u001b[0;34m=\u001b[0m\u001b[0;32mFalse\u001b[0m\u001b[0;34m)\u001b[0m\u001b[0;34m\u001b[0m\u001b[0;34m\u001b[0m\u001b[0m\n",
+       "\u001b[0;31mDocstring:\u001b[0m\n",
+       "Return a new list containing all items from the iterable in ascending order.\n",
+       "\n",
+       "A custom key function can be supplied to customize the sort order, and the\n",
+       "reverse flag can be set to request the result in descending order.\n",
+       "\u001b[0;31mType:\u001b[0m      builtin_function_or_method\n"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "sorted?"
+   ]
   }
  ],
  "metadata": {
@@ -166,7 +210,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.4"
+   "version": "3.7.5rc1"
   },
   "widgets": {
    "state": {},
@@ -174,5 +218,5 @@
   }
  },
  "nbformat": 4,
- "nbformat_minor": 1
+ "nbformat_minor": 4
 }
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -111,19 +111,13 @@ RST_TEMPLATE = """
 {%- if not cell.outputs %}
     :no-output:
 {%- endif %}
-{%- if cell.source.strip() %}
 
 {{ cell.source.strip('\n') | indent }}
-{%- endif %}
 {% endblock input %}
 
 
 {% macro insert_nboutput(datatype, output, cell) -%}
 .. nboutput::
-{%- if datatype == 'text/plain' %}{# nothing #}
-{%- elif datatype == 'ansi' %} ansi
-{%- else %} rst
-{%- endif %}
 {%- if output.output_type == 'execute_result' and cell.execution_count %}
     :execution-count: {{ cell.execution_count }}
 {%- endif %}
@@ -133,10 +127,25 @@ RST_TEMPLATE = """
 {%- if output.name == 'stderr' %}
     :class: stderr
 {%- endif %}
-{%- if datatype == 'text/plain' -%}
-{{ insert_empty_lines(output.data[datatype]) }}
+{%- if datatype != 'text/plain' %}
+    :fancy:
+{%- endif %}
+{%- if datatype == 'text/plain' %}
 
-{{ output.data[datatype].strip('\n') | indent }}
+    .. rst-class:: highlight
+
+    .. raw:: html
+
+        <pre>
+{{ output.data[datatype] | ansi2html | indent | indent }}
+        </pre>
+
+    .. raw:: latex
+
+        \\begin{sphinxVerbatim}[commandchars=\\\\\\{\\}]
+{{ output.data[datatype] | escape_latex | ansi2latex | indent | indent }}
+        \\end{sphinxVerbatim}
+
 {%- elif datatype in ['image/svg+xml', 'image/png', 'image/jpeg', 'application/pdf'] %}
 
     .. image:: {{ output.metadata.filenames[datatype] | posix_path }}
@@ -179,38 +188,6 @@ RST_TEMPLATE = """
     .. raw:: html
 
         <script type="{{ datatype }}">{{ output.data[datatype] | json_dumps }}</script>
-{%- elif datatype == 'ansi' %}
-
-    .. rst-class:: highlight
-
-    .. raw:: html
-
-        <pre>
-{{ output.data[datatype] | ansi2html | indent | indent }}
-        </pre>
-
-    .. raw:: latex
-
-        %
-        {
-        \\kern-\\sphinxverbatimsmallskipamount\\kern-\\baselineskip
-        \\kern+\\FrameHeightAdjust\\kern-\\fboxrule
-        \\vspace{\\nbsphinxcodecellspacing}
-        \\sphinxsetup{VerbatimBorderColor={named}{nbsphinx-code-border}}
-    {%- if output.name == 'stderr' %}
-        \\sphinxsetup{VerbatimColor={named}{nbsphinx-stderr}}
-    {%- else %}
-        \\sphinxsetup{VerbatimColor={named}{white}}
-    {%- endif %}
-        \\fvset{hllines={, ,}}%
-        \\begin{sphinxVerbatim}[commandchars=\\\\\\{\\}]
-{{ output.data[datatype] | escape_latex | ansi2latex | indent | indent }}
-        \\end{sphinxVerbatim}
-        }
-        % The following \\relax is needed to avoid problems with adjacent ANSI
-        % cells and some other stuff (e.g. bullet lists) following ANSI cells.
-        % See https://github.com/sphinx-doc/sphinx/issues/3594
-        \\relax
 {% else %}
 
     .. nbwarning:: Data type cannot be displayed: {{ datatype }}
@@ -418,6 +395,20 @@ LATEX_PREAMBLE = r"""
     \setbox\spx@image@box\box\voidb@x % clear memory
     \expandafter\includegraphics\expandafter[\spx@includegraphics@options]{#2}%
 }% end of "\MakeFrame"-safe variant of \sphinxincludegraphics
+\makeatother
+
+\makeatletter
+\renewcommand*\sphinx@verbatim@nolig@list{\do\'\do\`}
+\begingroup
+\catcode`'=\active
+\let\nbsphinx@noligs\@noligs
+\g@addto@macro\nbsphinx@noligs{\let'\PYGZsq}
+\endgroup
+\makeatother
+\renewcommand*\sphinxbreaksbeforeactivelist{\do\<\do\"\do\'}
+\renewcommand*\sphinxbreaksafteractivelist{\do\.\do\,\do\:\do\;\do\?\do\!\do\/\do\>\do\-}
+\makeatletter
+\fvset{codes*=\sphinxbreaksattexescapedchars\do\^\^\let\@noligs\nbsphinx@noligs}
 \makeatother
 """
 
@@ -918,17 +909,16 @@ class NotebookError(sphinx.errors.SphinxError):
 
 
 class CodeAreaNode(docutils.nodes.Element):
-    """Input area or output area of a Jupyter notebook code cell."""
+    """Input area or plain-text output area of a Jupyter notebook code cell."""
 
 
 class FancyOutputNode(docutils.nodes.Element):
-    """A custom node for non-code output of code cells."""
+    """A custom node for non-plain-text output of code cells."""
 
 
 def _create_code_nodes(directive):
     """Create nodes for an input or output code cell."""
-    fancy_output = False
-    language = 'none'
+    directive.state.document['nbsphinx_include_css'] = True
     execution_count = directive.options.get('execution-count')
     config = directive.state.document.settings.env.config
     if isinstance(directive, NbInput):
@@ -936,8 +926,6 @@ def _create_code_nodes(directive):
         if 'no-output' in directive.options:
             outer_classes.append('nblast')
         inner_classes = ['input_area']
-        if directive.arguments:
-            language = directive.arguments[0]
         prompt_template = config.nbsphinx_input_prompt
         if not execution_count:
             execution_count = ' '
@@ -946,11 +934,8 @@ def _create_code_nodes(directive):
         if 'more-to-come' not in directive.options:
             outer_classes.append('nblast')
         inner_classes = ['output_area']
-        # 'class' can be 'stderr'
         inner_classes.append(directive.options.get('class', ''))
         prompt_template = config.nbsphinx_output_prompt
-        if directive.arguments and directive.arguments[0] in ['rst', 'ansi']:
-            fancy_output = True
     else:
         assert False
 
@@ -965,21 +950,24 @@ def _create_code_nodes(directive):
     # NB: Prompts are added manually in LaTeX output
     outer_node += sphinx.addnodes.only('', prompt_node, expr='html')
 
-    if fancy_output:
+    if isinstance(directive, NbInput):
+        text = '\n'.join(directive.content.data)
+        if directive.arguments:
+            language = directive.arguments[0]
+        else:
+            language = 'none'
+        inner_node = docutils.nodes.literal_block(
+            text, text, language=language, classes=inner_classes)
+    else:
         inner_node = docutils.nodes.container(classes=inner_classes)
         sphinx.util.nodes.nested_parse_with_titles(
             directive.state, directive.content, inner_node)
-        if directive.arguments[0] == 'rst':
-            outer_node += FancyOutputNode('', inner_node, prompt=prompt)
-        elif directive.arguments[0] == 'ansi':
-            outer_node += inner_node
-        else:
-            assert False
+
+    if 'fancy' in directive.options:
+        outer_node += FancyOutputNode('', inner_node, prompt=prompt)
     else:
-        text = '\n'.join(directive.content.data)
-        inner_node = docutils.nodes.literal_block(
-            text, text, language=language, classes=inner_classes)
-        codearea_node = CodeAreaNode('', inner_node, prompt=prompt)
+        codearea_node = CodeAreaNode(
+            '', inner_node, prompt=prompt, stderr='stderr' in inner_classes)
         # See http://stackoverflow.com/q/34050044/.
         for attr in 'empty-lines-before', 'empty-lines-after':
             value = directive.options.get(attr, 0)
@@ -1011,7 +999,6 @@ class NbInput(rst.Directive):
 
     def run(self):
         """This is called by the reST parser."""
-        self.state.document['nbsphinx_include_css'] = True
         return _create_code_nodes(self)
 
 
@@ -1019,20 +1006,17 @@ class NbOutput(rst.Directive):
     """A notebook output cell with optional prompt."""
 
     required_arguments = 0
-    optional_arguments = 1  # 'rst' or nothing (which means literal text)
     final_argument_whitespace = False
     option_spec = {
         'execution-count': rst.directives.positive_int,
         'more-to-come': rst.directives.flag,
-        'empty-lines-before': rst.directives.nonnegative_int,
-        'empty-lines-after': rst.directives.nonnegative_int,
+        'fancy': rst.directives.flag,
         'class': rst.directives.unchanged,
     }
     has_content = True
 
     def run(self):
         """This is called by the reST parser."""
-        self.state.document['nbsphinx_include_css'] = True
         return _create_code_nodes(self)
 
 
@@ -1250,12 +1234,12 @@ def _get_empty_lines(text):
 def _get_output_type(output):
     """Choose appropriate output data types for HTML and LaTeX."""
     if output.output_type == 'stream':
-        html_datatype = latex_datatype = 'ansi'
+        html_datatype = latex_datatype = 'text/plain'
         text = output.text
-        output.data = {'ansi': text[:-1] if text.endswith('\n') else text}
+        output.data = {'text/plain': text[:-1] if text.endswith('\n') else text}
     elif output.output_type == 'error':
-        html_datatype = latex_datatype = 'ansi'
-        output.data = {'ansi': '\n'.join(output.traceback)}
+        html_datatype = latex_datatype = 'text/plain'
+        output.data = {'text/plain': '\n'.join(output.traceback)}
     else:
         for datatype in DISPLAY_DATA_PRIORITY_HTML:
             if datatype in output.data:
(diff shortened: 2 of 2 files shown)


jfbu:
It seems you keep the LaTeX changes to a minimum, and it solves the problem on your side, then this looks good! For info I compiled nbsphinx.pdf in this branch using both Sphinx current 2.0 branch as well as my [WIP branch](https://github.com/sphinx-doc/sphinx/compare/2.0...jfbu:6804_long_code_block_in_danger) (which I will turn into a PR right away) and I detected no problem with the enhanced `sphinxVerbatim` the WIP branch promotes to fix sphinx-doc/sphinx#6804. However I examined PDF very quickly.

mgeier:
Thanks a lot @jfbu! I've also tried it with your branch and haven't found any problems.

I've found an unrelated problem, though: The single quotes are different in input and output cells:

image

See: https://51-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#subsection.7.4

It looks like Pygments replaces the quotes with `\PYGZsq{}`.

Do you happen to know how I can also use straight quotes in the output cell?

Do I have to use `\sphinxupquote{}` or is there a setting to apply this to the whole `sphinxVerbatim` environment (or globally to all environments)?

jfbu:
Hmm... if I copy over the

```latex
\begin{sphinxVerbatim}[commandchars=\\\{\}]
\llap{\color{red}[1]:\,\hspace{\fboxrule}\hspace{\fboxsep}}\PYG{l+s+s1}{\PYGZsq{}}\PYG{l+s+s1}{no problem}\PYG{l+s+s1}{\PYGZsq{}}
\end{sphinxVerbatim}
```

(changing color) into the output of a sphinx project I had handy (hence the background color) I have no problem in output. 

Capture d’écran 2019-11-10 à 14 36 37

I need to understand what's going wrong. But I was going to leave, so if you don't hear from me in next half-hour I will be offline until tonight European time.

Sphinx does take car to handle the `\PYGZsq` in its `\sphinxbreaksatspecials`.

s-weigand:
The HTML output in [my usecase](https://github.com/students-teach-students/python-tools-for-students) (compare with #330 ), also looks as it should now. 😃 
Thanks @mgeier 

**New HTML render:**
npsphinx_unicode_fix

jfbu:
I am sooo stupid. Of course the relevant part of nbsphinx.tex is

```
\begin{sphinxVerbatim}[commandchars=\\\{\}]
\llap{\color{nbsphinxout}[1]:\,\hspace{\fboxrule}\hspace{\fboxsep}}'no problem'
\end{sphinxVerbatim}
```

there the ``'`` is naked and not escaped to `\PYGZsq{}`.

jfbu:
a bit short on time but the story is that `sphinxVerbatim` does not make `'` an active character because it is supposed to have been tex-escaped by Pygmentize library. It does

```
\let\verbatim@nolig@list \sphinx@verbatim@nolig@list
```

and one has a very short  list ``\def\sphinx@verbatim@nolig@list {\do \`}``, because all others are TeX-escaped, execpt the comma but the comma is handled especially to allow linebreaks and Sphinx does not want `fancyvrb` to override this. Thus the very short `\verbatim@nolig@list` is what gets executed by `fancyvrb` via `\@noligs` and it thus misses making `'` an active character (which is needed for it to avoid being rendered curly because even monospace TeX fonts usually have it curly).

Thus

```latex
{\catcode`\'\active
\begin{sphinxVerbatim}[commandchars=\\\{\}]
'no problem'
\end{sphinxVerbatim}
}
```

fixes it.

jfbu:
Maybe slightly better way is 

```latex
{\fvset{codes*=\catcode`\'\active}
\begin{sphinxVerbatim}[commandchars=\\\{\}]
'no problem'
\end{sphinxVerbatim}
}
```

and probably you can issue only once in preamble

```
\fvset{codes*=\catcode`\'\active}
```

I don't expect bad things out of adding this to preamble. The effet is to add activation of `'` in (sphinx)Verbatim environments. It is done by default in original Verbatim, but not in sphinxVerbatim for reasons explained above of not being needed a priori.

jfbu:
> It looks like Pygments replaces the quotes with `\PYGZsq{}`.
> 

Yes

> Do you happen to know how I can also use straight quotes in the output cell?
> 

I had misunderstood your question initially but the answer is that the `'` character must be made "active". It is done by original `fancyvrb`'s  `Verbatim`  but not by `sphinxVerbatim` because the latter does not need it (cf `PYGZsq`) and must override some of `fancyvrb`'s  set-up and it did that in a way not handling the `'` character (as not being needed).

> Do I have to use `\sphinxupquote{}` or is there a setting to apply this to the whole `sphinxVerbatim` environment (or globally to all environments)?

The ``\fvset{codes*=\catcode`\'\active)`` is the way to add the config to all `sphinxVerbatim` environments.

Better to avoid making the `'` active globally because its active definition is the one for math mode and only verbatim like environments (which execute `\@noligs`)  confer to it the `\textquotesingle` meaning adapted for text mode.

mgeier:
That works perfectly, thanks @jfbu!
I've added your suggestion in edebf63078f6d4f0df6cf18177179d0358b73bd5.

Now the quotes are straight: https://53-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#subsection.7.4

@s-weigand Thanks for testing!

I will leave this PR open for a few days and if nothing else comes up, I will merge it (and close the other alternatives).
And then I think I'll make a new release.

jfbu:
If I get this right you are using `sphinxVerbatim` with contents not having been malaxed by Pygmentize? Thus for example we can very well have  `--`  in the LaTeX cell code? Basically, things will be as TeX-escaped as done by https://github.com/sphinx-doc/sphinx/blob/2.0/sphinx/util/texescape.py, right? Then we need to be a bit more pro-active to avoid mechanism of TeX ligatures transforming `--` or `---`.

Why has `sphinxVerbatim` no provision for that? because it assumed the Pygments mark-up is used (`\PYGZhy`). The fix is relatively simple. Add to LaTeX preamble:

```
\makeatletter
% Sphinx default only does \do\` because other characters are assumed Pygmentized
% the ' and ` will remain straight in PDF output
% the - will get prevented from forming TeX ligatures in PDF output
\def\sphinx@verbatim@nolig@list{\do\'do\`\do\-}
\makeatother
```

As the above handles the straight quote `'` the `\fvset` become superfluous and can be dropped. The above is to be inserted in preamble (after `sphinx` package loading).

A priori the `<` and `>`  can give TeX ligatures too but they need no handling as I expect them escaped to `r'\textless{}'` and ` r'\textgreater{}'`, if I understand correctly @mgeier context here. In future, btw, Sphinx will probably use rather `r'\sphinxtextless{}'` and ` r'\sphinxtextgreater{}'` to allow special customizations. There is currently code in `sphinx.sty` for "parsed-literals" which will then need to adjust to the new names.

See `\sphinxbreaksattexescapedchars` in `sphinx.sty`; it could make sense to use that macro here for those `sphinxVerbatim` environment which are used with non-Pygmentized contents: then soft wrap of long lines will work about the same as in normal `sphinxVerbatim` having only Pygmentized contents.

I am sorry that this comment indicates something to do and then goes to say perhaps even something else needs to be done, but I have no clear understanding of what exactly can end up inside these `sphinxVerbatim` blocks.

There is also `\sphinxbreaksviaactiveinparsedliteral` which could be added and it handles the `-` so the `-` would then need to be removed again from above hint at top of this comment.

The syntax would be something like 

```
\fvset{codes*=\catcode`\'\active\sphinxbreaksattexescapedchars\sphinxbreaksviaactiveinparsedliteral}
```

I have tested it very briefly. I *think* it can be done globally in preamble, i.e. that the impact on standard code cells using `sphinxVerbatim` with Pygmentized contents is nil, but this could require testing.

What you gain with the above is that long lines will wrap at those characters.

Beware however if you end up with math mode things in such cells? Is it at all possible? Sphinx code for parsed-literals has extra steps for those.

Can you have a `\url` inside such environments?

mgeier:
> If I get this right you are using `sphinxVerbatim` with contents not having been malaxed by Pygmentize?

Yes.
Because I want to apply ANSI control sequences, and I don't think that's possible with Pygments, right?

In fact, I could use a `docutils.nodes.literal_block` and afterwards apply the `ansi2latex` filter, but this would only work for LaTeX output. I didn't find a way to make this work with HTML output.

That's why I'm not using `docutils.nodes.literal_block` but doing it manually with `raw` directives.

> Thus for example we can very well have  `--`  in the LaTeX cell code?

Yes, totally.
Anything that can be created by Pythons `print()` function.
This is typically plain text, and it can contain control characters. Especially ANSI escape sequences.

> Basically, things will be as TeX-escaped as done by https://github.com/sphinx-doc/sphinx/blob/2.0/sphinx/util/texescape.py, right?

Currently the content of code cell outputs is passed through the filters `escape_latex` and `ansi2latex`, which are both provided by `nbconvert`:

https://github.com/jupyter/nbconvert/blob/b0209dfd029da8797a144023b008198c3e36aa32/nbconvert/filters/latex.py#L30-L64

https://github.com/jupyter/nbconvert/blob/b0209dfd029da8797a144023b008198c3e36aa32/nbconvert/filters/ansi.py#L64-L74

`escape_latex` is apparently doing fewer replacement than `texescape.py`, but I guess that's an `nbconvert` issue.

> ```
> \makeatletter
> \def\sphinx@verbatim@nolig@list{\do\'do\`\do\-}
> \makeatother
> ```

This doesn't seem to work.

I don't know why exactly, but I get an error from LaTeX:

```
! Improper alphabetic constant.
<to be read again> 
                   \global 
l.325 \begin{sphinxVerbatim}[commandchars=\\\{\}]
                                                 
? 
```

> A priori the `<` and `>`  can give TeX ligatures too but they need no handling as I expect them escaped to `r'\textless{}'` and ` r'\textgreater{}'`, if I understand correctly @mgeier context here.

No,  `<` and `>` are not replaced and e.g. `>>` becomes a ligature `»`!

> ```
> \fvset{codes*=\catcode`\'\active\sphinxbreaksattexescapedchars\sphinxbreaksviaactiveinparsedliteral}
> ```

This seems to work well!
(except for `<` and `>`)

> Beware however if you end up with math mode things in such cells? Is it at all possible?

No. This is only for plain text.
Math is handled separately in "fancy" outputs.

> Can you have a `\url` inside such environments?

No, actual LaTeX output is also handled in "fancy" outputs.
If there is LaTeX markup in "non-fancy" outputs, it should be shown verbatim in plain text.

This is how math/LaTeX output is supposed to look like (using "fancy" outputs): https://buildmedia.readthedocs.org/media/pdf/nbsphinx/0.4.3/nbsphinx.pdf#subsubsection.4.3.3

jfbu:
> > ```
> > \makeatletter
> > \def\sphinx@verbatim@nolig@list{\do\'do\`\do\-}
> > \makeatother
> > ```
> 
> 
> This doesn't seem to work.

Very sorry about typo (missing backslash in second `do`) it should have read

```latex
\makeatletter
\def\sphinx@verbatim@nolig@list{\do\'\do\`\do\-}
\makeatother
```

(but read further down)

> No, `<` and `>` are not replaced and e.g. `>>` becomes a ligature `»`!
> 
> > ```
> > \fvset{codes*=\catcode`\'\active\sphinxbreaksattexescapedchars\sphinxbreaksviaactiveinparsedliteral}
> > ```
> 
> This seems to work well!
> (except for `<` and `>`)

Taking into account the information you provided I can propose the following possibly not quite optimized:

```
\makeatletter
\renewcommand*\sphinx@verbatim@nolig@list{\do\'\do\`}
\makeatother
\renewcommand*\sphinxbreaksbeforeactivelist{\do\<}
\renewcommand*\sphinxbreaksafteractivelist{\do\.\do\,\do\:\do\;\do\?\do\!\do\/\do\>\do\-}
\fvset{codes*=\sphinxbreaksattexescapedchars}
```

I was worried about the hyphen character `-` when I saw mark-up such as ``\colorbox{ansi-yellow}{CD\strut}`` but examining a TeX trace I discovered that `xcolor` package has some sanitizing of its own when certain characters have been made active, and in particular the hyphen `-` as present here in `ansi-yellow` name.

The characters sanitized by `xcolor` in such color names seem to be `!:-+;">` but for some reason not the `<`. So you can use `>` in a color name (not that I recommend it...) but not a `<` is such color names end up in  environments "activating" `<`, `>` like is the case  of `sphinxVerbatim` with above config.

In the above proposed config, the single quotes ``'` `` are only make active to become straight quotes, but will not offer potential breakpoints. For this my impression is that a specific macro is to be added in place of simply like in the above recycling some already in `sphinx.sty`.

The list of active characters above can possibly be extended to include `+`, `:`, and  `"` because they are sanitized by `xcolor`. Already above I added the  colon `:`: for some reason `sphinxVerbatim` currently seems to have forgotten it from candidate characters for being made active (as they are not Pygmentized) to allow linebreaks. Possibly an oversight. Or perhaps some cautiousness for reasons I have forgotten.

One has to be careful with making characters active if the input contains macros where the characters may appear in some "string"-like argument (not typeset, but deciding what is done). And the ansi2latex produces such `\textcolor` and `\colorbox` usage as commented above. If no other LaTeX macros are added by mark-up we should be safe with above config.



> Currently the content of code cell outputs is passed through the filters `escape_latex` and `ansi2latex`, which are both provided by `nbconvert`:

Thanks for the links. Makes me think if Sphinx adopts some `\sphinxtextbackslash` mark-up it must be careful not to break `nbconvert` using `\textbackslash`.

mgeier:
Thanks, the new settings work very well, AFAICT. I've added them in  5803e465e08dfc934c30d06f37065fc2bf0dda4c

However, if you think that's a better solution, I could also create an additional custom filter to be applied after `nbconvert`'s `escape_latex`.

That way I could manually add some of the replacements from `sphinx.util.texescape.escape`.

I also tried using `sphinx.util.texescape.escape` directly as a filter (*instead of* `escape_latex`), but this replaced too many things. The problem was the literal `[` character which appears in all ANSI color sequences. And the ANSI decoding obviously doesn't work if `[` is replaced by `{[}`. But I guess all other replacements from `texescape.py` should be safe.
 
> So you can use `>` in a color name (not that I recommend it...)

I'm not planning to do that.
And I don't expect those names to change, so only the hyphen will be used as a special character.

> If no other LaTeX macros are added by mark-up we should be safe with above config.

All possible LaTeX markup should be in there: https://github.com/jupyter/nbconvert/blob/b0209dfd029da8797a144023b008198c3e36aa32/nbconvert/filters/ansi.py#L120-L168

jfbu:
Raw code such as `\textcolor[RGB]{255,0,0}{test}` would crash due to activation of `,` but this is already taken into account by the ansi2latex converter:

https://github.com/jupyter/nbconvert/blob/b0209dfd029da8797a144023b008198c3e36aa32/nbconvert/filters/ansi.py#L149-L154

For info, original LaTeX includes the comma `,` in the list of activated characters in verbatim output because `,, ` is a ligature in T1-encoded fonts. The default list is:

```
$ latexdef verbatim@nolig@list

\verbatim@nolig@list:
macro:->\do \`\do \<\do \>\do \,\do \'\do \-
```

The package `upquote` loaded automatically by Sphinx hacks into `\@noligs` in order to modify the way the single quotes are handled, thus not only are ligatures modified but the single quotes are rendered as straight quotes in output.

`sphinxVerbatim` wants to do more than preventing ligatures: it wants to allow linebreaks and for that it too needs to activate some characters but it assigns to them some definitions allowing the linebreaks (and they turn to also in passing prevent ligatures). As the processing by `fancyvrb` of `\verbatim@nolig@list` is done *after*  the Sphinx set-up, Sphinx has its own shorter `\sphinx@verbatim@nolig@list` (and I should have not used `@` letter in there) which locally replaces `\verbatim@nolig@list`.

Its default is only ``\do\` `` for following reasons:

- the `'`, `<`, `>`, and `-` should have been Pygmentized hence they were not expected to appear in contents, and then why activate them as it is always potentially dangerous to do that? And for the quote `'`, the Sphinx redefinition of `\PYGZsq` makes sure to also use `\textquotesingle` which gives a straight quote in PDF output.

- the comma `,` is not Pygmentized, hence it is handled in `\sphinxbreaksviaactive` to allow a linebreak after it, and this also prevents the T1 ``,, `` ligature,

- there only remained the `` ` ``, which is kept in the list to allow the "upquote" hack to work to let this become straight in PDF output.

In retrospect, Sphinx could have left the Pygmentized characters `'`, `<`, `>`, and `-` in the list: the rationale for not doing so was to avoid activating needlessly characters.

I am providing the above explanations as background info: what is to be done depends on whether you want or not to add to bare Verbatim treatement some potential linebreaks at specific characters.

Notice that `sphinxVerbatim` at any rate adds potential linebreaks at spaces. 

To limit its action to only that, you could redefine the `\sphinxbreaksbeforeactivelist`, `\sphinxbreaksafteractivelist` to be empty macros, and do in preamble

```
\makeatletter
\let\sphinx@verbatim@nolig@list\verbatim@nolig@list
\makeatother
\renewcommand*\sphinxbreaksbeforeactivelist{}
\renewcommand*\sphinxbreaksafteractivelist{}
```

in which case you end-up with almost same as `OriginalVerbatim` (aka `Verbatim`) with only the addition of linebreaks at spaces and consistent interline distances (and framing of course). And the dangers of the active characters in this context (active to avoid ligatures) are taken into account by the `ansi2latex` which uses `\detokenize`  to handle the comma, and also by `xcolor` (I am not sure `nbconvert` maintainers are aware of that) which sanitizes the hyphen `-` which may appear in color names as used by `ansi2latex`.

What is to be done depends very much on exactly what kind of contents can appear in these environments. For ANSI colors, it looks as if proposal of previous comment is ok, but if you want to be conservative you can use the  above which preserves the original `\verbatim@nolig@list` and does not activate anything thing else (hence no extra `\fvset{codes*=...}`).

jfbu:
(I posted previous comment prior to having read yours...)

jfbu:
> and do in preamble
> 
> ```
> \makeatletter
> \let\sphinx@verbatim@nolig@list\verbatim@nolig@list
> \makeatother
> \renewcommand*\sphinxbreaksbeforeactivelist{}
> \renewcommand*\sphinxbreaksafteractivelist{}
> ```
> 
> in which case you end-up with almost same as `OriginalVerbatim`

But this proposal is bad as it affects **all** `sphinxVerbatim`. Of course you could include it inside the "raw" directives only to limit it to the `sphinxVerbatim` for non-code contents.

On the other hand my earlier proposal

```
\makeatletter
\renewcommand*\sphinx@verbatim@nolig@list{\do\'\do\`}
\makeatother
\renewcommand*\sphinxbreaksbeforeactivelist{\do\<}
\renewcommand*\sphinxbreaksafteractivelist{\do\.\do\,\do\:\do\;\do\?\do\!\do\/\do\>\do\-}
\fvset{codes*=\sphinxbreaksattexescapedchars}
```

although affecting **all** `sphinxVerbatim` a priori does not alter the way the regular `sphinxVerbatim` containing Pygmentized contents do their job.

jfbu:
> Thanks, the new settings work very well, AFAICT. I've added them in [5803e46](https://github.com/spatialaudio/nbsphinx/commit/5803e465e08dfc934c30d06f37065fc2bf0dda4c)
> 
> However, if you think that's a better solution, I could also create an additional custom filter to be applied after `nbconvert`'s `escape_latex`.
> 
> That way I could manually add some of the replacements from `sphinx.util.texescape.escape`.
> 

Replacements of `sphinx.util.texescape.escape` are a legacy thing, I have not looked in detail at the rationale for each.

> I also tried using `sphinx.util.texescape.escape` directly as a filter (_instead of_ `escape_latex`), but this replaced too many things. The problem was the literal `[` character which appears in all ANSI color sequences. And the ANSI decoding obviously doesn't work if `[` is replaced by `{[}`. But I guess all other replacements from `texescape.py` should be safe.
> 

The `{[}` of `sphinx.util.texescape.escape` is probably related to prevent mis-interpretation in some contexts where a naked `[` would be interpreted as start of some optional argument.

For Unicode aspects, there is https://github.com/sphinx-doc/sphinx/pull/6750 

> > So you can use `>` in a color name (not that I recommend it...)
> 
> I'm not planning to do that.
> And I don't expect those names to change, so only the hyphen will be used as a special character.
> 
> > If no other LaTeX macros are added by mark-up we should be safe with above config.
> 
> All possible LaTeX markup should be in there: https://github.com/jupyter/nbconvert/blob/b0209dfd029da8797a144023b008198c3e36aa32/nbconvert/filters/ansi.py#L120-L168

OK, then things are fine I hope. The config I propose adds `:`  to activated characters, compared to current Sphinx which somehow overlooked it. (it is not Pygmentized, why did I not activate it? I was cautious I guess). You can extend even more list of activated characters (non letters, non digits), but do not do anything with braces `{`, `}`, and the backslash `\ `, and also don't activate the square brackets `[`, `]` because they appear in https://github.com/jupyter/nbconvert/blob/b0209dfd029da8797a144023b008198c3e36aa32/nbconvert/filters/ansi.py#L120-L168 syntax.

Thus linebreaks at `[`, `]`, and backslash are not possible. For backslash it was added recently at https://github.com/sphinx-doc/sphinx/pull/6799 but this works only because it is TeX-escaped. One can not do anything with the ``\ `` naked character which must conserve its meaning of staring TeX control sequences.

mgeier:
> what is to be done depends on whether you want or not to add to bare Verbatim treatement some potential linebreaks at specific characters.

TBH, I don't know.
This could be any output from some Python script, or from Bash, or whatever ...

Since it could be anything, I don't really know where line breaks make sense.
It would probably be best to try to mimic whatever `sphinxVerbatim` does, or what do you think?
I guess it would be strange to have different line breaks in code cell inputs (which use `docutils.nodes.literal_block` and therefore `sphinxVerbatim`) and code cell outputs, right?

> [...] but if you want to be conservative you can use the  above which preserves the original `\verbatim@nolig@list` and does not activate anything thing else (hence no extra `\fvset{codes*=...}`).

I don't know whether I want to be conservative or not.
But I think I prefer the less conservative option, because it seems to be closer to the behavior of input code cells.

> The config I propose adds `:`  to activated characters, compared to current Sphinx which somehow overlooked it.

OK, apparently this is consistent between input and output code cells, very good!

> Thus linebreaks at `[`, `]`, and backslash are not possible.

OK, that seems to be consistent with input code cells.

> For backslash it was added recently at https://github.com/sphinx-doc/sphinx/pull/6799 but this works only because it is TeX-escaped. One can not do anything with the ``\ `` naked character which must conserve its meaning of staring TeX control sequences.

I'm not sure if I completely understand, but backslashes are replaced by `escape_latex` and therefore there will never be a backslash on its own.

I've tried it with Sphinx' `2.0` branch, and lines are correctly broken after `\`:

image

So I guess everything is fine now, right?

mgeier:
I've just found one difference: When I use `^`, this is replaced by `\^{}`, and no line break happens.
In the input, this is replaced by `\PYGZca{}`, and a line break happens afterwards.

jfbu:
>  I don't really know where line breaks make sense.
> It would probably be best to try to mimic whatever `sphinxVerbatim` does, or what do you think?

I tend to agree

> I've just found one difference: When I use `^`, this is replaced by `\^{}`, and no line break happens.
> In the input, this is replaced by `\PYGZca{}`, and a line break happens afterwards.

The problem is that nbconvert latex escaping is not like sphinx.util.texescape.escape the former uses `\^{}` the latter uses `\textasciicircum{}`, thus the action of `\sphinxbreaksattexescapechars` is not on right input. 

I will refactor in future to more easily adapt what `\sphinxbreaksattexescapechars`  does but here the simplest is to abuse that `\do` has correct meaning immediately after it so we can add it like below (very last line):


```
\makeatletter
\renewcommand*\sphinx@verbatim@nolig@list{\do\'\do\`}
\makeatother
\renewcommand*\sphinxbreaksbeforeactivelist{\do\<\do\"}
\renewcommand*\sphinxbreaksafteractivelist{\do\.\do\,\do\:\do\;\do\?\do\!\do\/\do\>\do\-}
\fvset{codes*=\sphinxbreaksattexescapedchars\do\^\^}
```

I also added the `\do\"` to `\sphinxbreaksbeforeactivelist` because normal sphinxVerbatim with Pygmentized contents does it with `\PYGZdq{}`.

The difference between `"` and `^` is that the former has to be made a TeX-active character but the latter was TeX-escaped to some macro `\^{}` and we only need redefine how that macro expands.

In both cases the above mimicks how they would behave if Pygmentized although now they are not Pygmentized.

And the config does not cancel or modify behaviour of sphinxVerbatim with Pygmentized contents because the macro `\sphinxbreaksatspecials` which handles Pygmentized contents is executed automatically in addition to the above. (I should have given it another name like `\sphinxbreaksatpygments`).


> I'm not sure if I completely understand, but backslashes are replaced by `escape_latex` and therefore there will never be a backslash on its own.


Yes, I was sort or speaking in a kind of ultimate generality which was near non-sense. That's the whole charm of my verbosity.

> 
> I've tried it with Sphinx' `2.0` branch, and lines are correctly broken after `\`:
> 
> image
> 
> So I guess everything is fine now, right?

Indeed! thanks for testing this with Sphinx dev repo.

jfbu:
edit: one would like to get the `'` to behave like in Pygmentized sphinxVerbatim. But there is a difficulty because the meaning of its active self is reset very late by Verbatim after Sphinx customizations of active characters. To work around that needs some non-trival LaTeX, sorry.


```
\makeatletter
\renewcommand*\sphinx@verbatim@nolig@list{\do\'\do\`}
\begingroup
\catcode`'=\active
\g@addto@macro\@noligs{\let'\PYGZsq}
\endgroup
\makeatother
\renewcommand*\sphinxbreaksbeforeactivelist{\do\<\do\"\do\'}
\renewcommand*\sphinxbreaksafteractivelist{\do\.\do\,\do\:\do\;\do\?\do\!\do\/\do\>\do\-}
\makeatletter
\fvset{codes*=\sphinxbreaksattexescapedchars\do\^\^}
\makeatother
```

jfbu:
The idea of previous trick in comment is that `\PYGZsq` is already re-defined by Sphinx to allow a linebreak at the time the `\let'\PYGZsq` will be done, so this avoids having to type the whole complicated `\discretionary...` thing.

jfbu:
This makes me think an even better way would be to generalize the `\let'\PYGZsq` trick to other characters so that really @mgeier `sphinxVerbatim` for plain-text inherits `sphinxVerbatim` Pygmentized behaviour *and* obeys user modifications of `\sphinxbreaksbeforelist/\sphinxbreaksafterlist` but I propose we call a truce here. And I was planning to have dinner...

jfbu:
memo: the modification to `\@noligs` in code above influences also inline-literals (`\sphinxupquote`) but there is no change in PDF output, instead of having active `'` a strict synonym of `\textquotesingle` (done by `upquote` package) it becomes synonym to `\PYGZsq` which is a macro expanding to `\textquotesingle` (defined in `sphinxhighlight.sty`) hence only very slight negligible overhead.

If the modification of `\@noligs` is to be strictly limited to `sphinxVerbatim`:

```
\makeatletter
\renewcommand*\sphinx@verbatim@nolig@list{\do\'\do\`}
\begingroup
\catcode`'=\active
\let\nbsphinx@noligs\@noligs
\g@addto@macro\nbsphinx@noligs{\let'\PYGZsq}
\endgroup
\makeatother
\renewcommand*\sphinxbreaksbeforeactivelist{\do\<\do\"\do\'}
\renewcommand*\sphinxbreaksafteractivelist{\do\.\do\,\do\:\do\;\do\?\do\!\do\/\do\>\do\-}
\makeatletter
\fvset{codes*=\sphinxbreaksattexescapedchars\do\^\^\let\@noligs\nbsphinx@noligs}
\makeatother
```

Sorry. The problem is that a priori we don't want our modifications to perturbate `\sphinxupquote`.

mgeier:
Wow, thanks for all this TeX magic!

I've squashed the previous commits into 791c6eae8d73f260a33ccbc9d5a321131d7f681f and added you as a co-author, because I clearly wouldn't have been able to come up with any of this.

jfbu:
@mgeier thanks for the honor! I suddenly dearly hope nothing catastrophic will result from merging this (my insolvency is already in place, in case some litigation ensues :) )

I will be careful if something evolves at Sphinx side it does not break this...

mgeier:
Don't worry, if something breaks we can fix it again!

And to avoid litigation, we have this:

https://github.com/spatialaudio/nbsphinx/blob/5faf5bb84110f01591f24f09df2315e213181263/LICENSE#L13-L19

PR #364, opened 2019-12-11
s/prolog/prologue, s/epilog/epilogue

chmreid:
Fix two typos. Prolog and epilog should be spelled prologue and epilogue. Misspelling is confusing, partly because Prolog is a programming language.


mgeier:
I've chosen those words in reference to the already long-existing Sphinx options `rst_prolog` and `rst_epilog`: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_epilog

I think if any of the 4 options is changed, all of them should be changed.

But this would be a breaking change, and would be much more confusing and annoying than thinking for a second about the Prolog programming language.

Also, I'm not even sure it is a misspelling in the first place.

I'm not a native English speaker, so I don't have my own opinion about that, but a quick web search yields:

* prologue or less commonly prolog: https://www.merriam-webster.com/dictionary/prologue
* prologue also prolog: https://www.thefreedictionary.com/Prolog
* prologue or prolog: https://www.dictionary.com/browse/prolog
* prologue or prolog: https://en.wikipedia.org/wiki/Prologue

Clearly, "prologue" seems to be the more common spelling, but I didn't find a source that ruled out "prolog" as a misspelling.

Did you find a dictionary that doesn't allow "prolog"?

chmreid:
OK - I figured this PR might break some things. It was a quick sed one-liner to make this PR so I figured I'd submit it anyway, but it sounds like there are good reasons for going with prolog/epilog instead of prologue/epilogue.

PR #380, opened 2020-01-14
Use PyPi version of sphinx-material

s-weigand:
`sphinx-material` has a [release on PyPi](https://pypi.org/project/sphinx-material/) now, no need to install it from the repo anymore.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
-git+https://github.com/bashtage/sphinx-material.git
+sphinx-material
 ipykernel
 numpy
 matplotlib


s-weigand:
I guess they just forgot to change it.

Btw. the tests fail since the package is `python>=3. 6`, and it can't be installed from PyPi for lower versions. But that would be a problem user need to take care themselves anyway. Since the code might break in other cases, i.e. if the theme code uses formatted-strings. So the failing should be rated as intended.

Maybe mention it in the docs for building themes?

mgeier:
I've made a PR: https://github.com/bashtage/sphinx-material/pull/26.

I wouldn't worry about the Python 3.6 dependency. That's not our problem.

Most people should use >= 3.6 by now anyway ...

PR #382, opened 2020-01-17
Add 'alabaster-jupyterhub' to usage docs

### LINKED ISSUE #381: Add missing themes to docs
[The docs](https://nbsphinx.readthedocs.io/en/0.5.0/usage.html#HTML-Themes) only mention 27 out of 31 themes, which are build by RTD.

s-weigand:
Added missing 'alabaster-jupyterhub' to usage docs.

closes #381


doc/usage.ipynb
```diff
@@ -653,6 +653,10 @@
     "\n",
     "### 3rd-Party Themes\n",
     "\n",
+    "* [alabaster-jupyterhub](https://github.com/jupyterhub/alabaster-jupyterhub):\n",
```
mgeier: I was using the actual `html_theme` name for the first link. In this case this would be `alabaster_jupyterhub`.

If you think the PyPI name or the Python module name would make more sense, we can change this, but then we should change it for all themes (in a separate PR).

I must admit that the branch names are not consistent.
Probably I should re-name the branches?

mgeier: I think the canonical name is the one used in `html_theme`. All themes have that. Everything else (PyPI, Github, even Python module) is optional.

I just wanted to start re-naming all branches, but I found a problem: Some theme names end with "theme", which would lead to branch names like e.g. `sphinx_typlog_theme-theme`, which looks a bit strange, doesn't it?

Do you have an idea for a consistent naming scheme that doesn't suck?

s-weigand: 🤔
Tough one. 
How about `theme-<html_theme >`? 
Like that, at least the `theme-theme` can be prevented and the theme branches are all grouped together.

mgeier: Yeah, that sounds slightly better. The problem is then I would have to break all links (not just a few). I'm not sure if that's worth the disruption?

Another disadvantage with longer branch names is that the list of versions/themes in the sidebar on RTD is already quite unwieldy and it would become significantly longer.

I'm leaning towards not changing the branch names ...

But we can change the links on the "usage" page if you like?

PR #387, opened 2020-01-22
theme comparison tool: use "git worktree" to build themes

mgeier:
This is not meant to be merged!

This is just a proof of concept, to be able to actually try if the "git worktree" approach could work.

Of course a lot of error checking, comments and other details are missing.

The goal was to implement the essential functionality with as little code as possible.

The version in #379 has many more features, but I just wanted to try the basic stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ nbsphinx.egg-info/
 .vscode
 doc/_build
 doc/gallery/a-local-file.png
+/theme_comparison/
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -56,7 +56,36 @@ Again, you'll probably have to use ``python`` instead of ``python3``.
 The generated files will be available in the directories ``build/sphinx/html/``
 and ``build/sphinx/latex/``, respectively.
 
+Building Themes
+---------------
+
+The ``nbsphinx`` documentation is available in over 30 different `HTML themes`_,
+with each having its own branch ending in ``-theme``.
+
+To simplify the building and testing of themes,
+which is especially needed when changing CSS,
+we provide you with command line tool to build all themes
+or a user specified subset.
+The tool is located at ``dev_utils/theme_builder.py`` and can be run with::
+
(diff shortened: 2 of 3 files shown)


s-weigand:
Just tried it and it works nicely 😄 

But I found 2 Problems:
1. If the user removed `devtools/_worktree` manually, git still knows it and won't create a new worktree.
This can be solved with (the replace is to support windows):
```python
if not WORKTREE_DIR.exists():
    if str(WORKTREE_DIR).replace("\\", "/") in repo.git.worktree("list"):
        repo.git.worktree("remove", WORKTREE_DIR)
    repo.git.worktree("add", WORKTREE_DIR, "--detach")
```
2. New unstaged files are ignored .

mgeier:
Thanks for trying this out!

ad 1: I've also seen this problem in the meantime. Your solution should work fine, but probably we could use [as_posix()](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix) to avoid the manual replace?
We could also try `git worktree prune` instead of explicitly removing it?

ad 2: I know. As mentioned in https://github.com/spatialaudio/nbsphinx/pull/379#issuecomment-577029875, `--include-untracked` doesn't seem to be available here.
I would just keep this as a known limitation (unless you know a *simple* solution), I don't think it is a big deal.

Anyway, I just wanted to see here if what I suggested in #379 would actually work and how much code it would need.

Obviously this is much shorter than #379, but of course there are several features missing here.

Do you want to try to make your solution in #379 shorter (or at least simpler) or probably use a few things from here over there?

s-weigand:
Using `as_posix()` sounds good, as you saw in my code I use `os.path` all the time, guess I should use `pathlib` more often.

I will refactor #379 and use some of your code.

mgeier:
I've squashed the commits and made this ready for review.

@s-weigand I've tried to keep your authorship for the documentation.

I think it works really well, except the `custom-formats` page doesn't seem to be cached successfully (but I guess that's not the fault of this PR).

Issue #310, opened 2019-07-25
Doesn't have a title: no link will be generated


jetheurer:
I have an issue where the following warning comes up:

```
WARNING: toctree contains reference to document 'example' that doesn't have a title: no link will be generated
```

Where example is the name of my notebook `example.ipynb`

and this is the command I run: `python3 -m sphinx -b html -j 4 docs/source/ $BUILD`

When looking at the generated files, I dont get the link to example in my documentation.

My toctree looks like this:

```rst
.. toctree::
   :maxdepth: 2
   :caption: Contents:

   example.ipynb
   readme
   modules
```

To clarify, the html file: example.html builds, but the link in index.rst is dropped.

jetheurer:
This seems related to: https://stackoverflow.com/questions/55297443/including-of-notebook-with-nbsphinx-fails

jetheurer:
Here is a minimal example: [minimal_nbsphinx.zip](https://github.com/spatialaudio/nbsphinx/files/3463361/minimal_nbsphinx.zip)

Please run using `bash build.sh`

jetheurer:
For those that run into this again,

Here is a minimal working notebook:

```json
{
 "cells": [
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
     "# Test"
    ]
   }
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.7.3"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 2
}
```

The most relevant component being:
```json
    "source": [
     "# Test"
    ]
```
All notebooks need a markdown section with a title!

TCACastelijns:
I know this issue is closed, but I get still get the error above. Even if my notebooks contains titles.

My index.rst contains of
```

- predict.ipynb
- train.ipynb
- evaluate.ipynb
```

I wrote documentation for predict step with markdown and code examples:

```
{
 "cells": [
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "# Predict"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": 1,
   "metadata": {},
[...]

PR #401, opened 2020-02-27
Update README.rst

davidbailey:
I had this problem. And it looks like others have too (https://github.com/spatialaudio/nbsphinx/issues/310). I think a nice note is warranted.

### DIFF
--- a/doc/usage.ipynb
+++ b/doc/usage.ipynb
@@ -42,7 +42,7 @@
     "For an example, see this project's  [conf.py](conf.py) file.\n",
     "\n",
     "Once your `conf.py` is in place, edit the file named `index.rst` and add the file names of your notebooks (without the `.ipynb` extension) to the [toctree](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree) directive.\n",
-    "For an example, see this project's `doc/index.rst` file."
+    "Make sure these notebooks begin with a metadata cell that contains a title (e.g. # Title). For an example, see this project's `doc/index.rst` file."
    ]
   },
   {
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ Quick Start:
     #. Edit your ``conf.py`` and add ``'nbsphinx'`` to ``extensions``.
 
     #. Edit your ``index.rst`` and add the names of your ``*.ipynb`` files
-       to the ``toctree``.
+       to the ``toctree``. (Note: your *.ipynb files must begin with a metadata cell that has a ``# Title``.)
 
     #. Run Sphinx!
 
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ Quick Start:
     #. Edit your ``conf.py`` and add ``'nbsphinx'`` to ``extensions``.
 
     #. Edit your ``index.rst`` and add the names of your ``*.ipynb`` files
-       to the ``toctree``. (Note: your *.ipynb files must begin with a metadata cell that has a ``# Title``.)
+       to the ``toctree``.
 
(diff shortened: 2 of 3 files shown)


mgeier:
Thanks for this suggestion!

However, I'm very picky about what should go into the "quick start" (see e.g. https://github.com/spatialaudio/nbsphinx/pull/118#issuecomment-318324064). I think it's already too long as it is.

What about adding some more content to the "Usage" notebook instead?

I think it could also be worthwhile to try generating a more helpful warning message, don't you think?

davidbailey:
Agreed. Don't want the quick start to become 100 lines long.

You do get a sphinx error (https://github.com/sphinx-doc/sphinx/blob/master/sphinx/environment/adapters/toctree.py#L172), but this doesn't really tell you how to fix it. I tried to create an additional warning if there is no title when the exporter runs, but there may be a better way to handle that case.

mgeier:
Thanks for the update!

Your check is a bit brittle, though:

* The title doesn't have to be in the first cell
* It doesn't have to be in a Markdown cell (could also be a raw reST cell)
* The title doesn't have to be in the first line of the cell
* The `#` character can be preceded by up to three spaces
* The title can also be in "Setext" format (i.e. underlined with `=====` or `-------`)
* Probably more things?

It is often more robust to check something like this on the in-memory document structure instead of directly in the source code.

I've just tried a different approach in cc1695cf0150844ad596226ef832d1f486b496c9, what do you think about it?

doc/usage.ipynb
```diff
@@ -42,7 +42,7 @@
     "For an example, see this project's  [conf.py](conf.py) file.\n",
     "\n",
     "Once your `conf.py` is in place, edit the file named `index.rst` and add the file names of your notebooks (without the `.ipynb` extension) to the [toctree](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree) directive.\n",
-    "For an example, see this project's `doc/index.rst` file."
+    "Make sure these notebooks begin with a metadata cell that contains a title (e.g. # Title). For an example, see this project's `doc/index.rst` file."
```
mgeier: I guess you mean `Markdown` instead of `metadata`?

The notebook doesn't have to *begin* with it.

You should probably escape the example: `` `# Title`  ``?

When you insert this here, the following sentence doesn't work anymore, right?

davidbailey:
Awesome. I like the warning although it is pretty repetitive of the warning generated by Sphinx (https://github.com/sphinx-doc/sphinx/blob/master/sphinx/environment/adapters/toctree.py#L172).

I guess my problem was not actually the lack of a warning, but rather a lack of understanding how titles are generated and what to do about the warning. However as I still don't understand this I don't think I'm the right person to update the documentation.

mgeier:
Yes, it is a bit repetitive. But I think it's still worth adding the warning because it is more specific:

* it mentions the source file where the title is missing (which has to be fixed), while the other warning mentiones the source file with the `toctree` directive (which doesn't have to be fixed).
* it mentions "notebook", which the other warning doesn't

The repetitive warnings might be confusing, but at least the new warning would be the first one that appears in the output.

> I guess my problem was not actually the lack of a warning, but rather a lack of understanding how titles are generated and what to do about the warning.

Yeah, I can understand that. It isn't obvious.

I hope the new warning from #402 gives more clues about what to do.
Is the new warning clear enough or do you know a way to improve it further?

I think one problem is that you never explicitly give a "title" to a notebook. You only use some "section titles" (or "headlines") if you feel like it.

I guess the missing piece of information is that Sphinx uses the first "section title" as a "title" for the whole source file, and tables of contents don't really work if such a "title" is missing.

Issue #184, opened 2018-04-25
Readthedocs build error with nbsphinx 


JohnGriffiths:
Hi. 

I'm getting a build errors ('no module named nbsphinx') with readthedocs and am not sure why as I've set the conf, requirements, etc files to be exactly the same as those in nbsphinx (jupyter-alabaster-theme branch). 

Any ideas? 

Here is the readthedocs build output:

git remote set-url origin https://github.com/JohnGriffiths/ConWhAt.git
git fetch --tags --prune
git checkout --force origin/master
git clean -d -f -f
git show-ref --tags
git branch -r
conda env create --name master --file /home/docs/checkouts/readthedocs.org/user_builds/conwhat/checkouts/master/docs/readthedocs-environment.yml
conda install --yes --name master mock pillow sphinx sphinx_rtd_theme
python /home/docs/checkouts/readthedocs.org/user_builds/conwhat/conda/master/bin/pip install -U --cache-dir /home/docs/checkouts/readthedocs.org/user_builds/conwhat/.cache/pip recommonmark readthedocs-sphinx-ext
python setup.py install --force
cat docs/conf.py
python /home/docs/checkouts/readthedocs.org/user_builds/conwhat/conda/master/bin/sphinx-build -T -b readthedocs -d _build/doctrees-readthedocs -D language=en . _build/html
Running Sphinx v1.7.2
loading translations [en]... done

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/conwhat/conda/master/lib/python3.5/site-packages/sphinx/registry.py", line 315, in load_extension
    mod = __import__(extname, None, None, ['setup'])
ImportError: No module named 'nbsphinx'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/conwhat/conda/master/lib/python3.5/site-packages/sphinx/cmdline.py", line 303, in main
    args.warningiserror, args.tags, args.verbosity, args.jobs)
  File "/home/docs/checkouts/readthedocs.org/user_builds/conwhat/conda/master/lib/python3.5/site-packages/sphinx/application.py", line 191, in __init__
    self.setup_extension(extension)
  File "/home/docs/checkouts/readthedocs.org/user_builds/conwhat/conda/master/lib/python3.5/site-packages/sphinx/application.py", line 411, in setup_extension
    self.registry.load_extension(self, extname)
[...]

PR #404, opened 2020-03-05
support nbsphinx on Read the Doc #184

LovelyBuggies:
nbspinx is not in "./doc", leading into the problem in #184.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -6,3 +6,4 @@ sphinxcontrib-bibtex
 sphinxcontrib-svg2pdfconverter
 ipywidgets
 sphinx-copybutton
+nbsphinx


mgeier:
Thanks for this PR!

Your change is not necessary, though, because it already works on RTD, see https://nbsphinx.readthedocs.io/

How exactly did you experience this problem?

Adding `nbsphinx` might even be counter-productive in a development setting, because it might inadvertently install the latest release instead of the locally checked-out version.

I've just realized that `doc/requirements.txt` is not mentioned in the docs, so I created #405.

LovelyBuggies:
> I've just realized that `doc/requirements.txt` is not mentioned in the docs, so I created #405.

That's fine. Thanks!

PR #408, opened 2020-03-13
Allow keyword arguments in "nbsphinx_custom_formats"

mgeier:
I previously suggested to use a `lambda` function as configuration value, but it turned out that this breaks Sphinx's ability to cache source files, which means that all notebooks are (re)executed in every Sphinx run even when there are no changes, which is obviously bad.

This PR adds the possibility to add keyword arguments to the custom conversion function, which in turn should make it possible to specify the function itself as a string, which avoids the aforementioned problem.

Rendered docs: https://161-210404706-gh.circle-artifacts.com/0/html/custom-formats.html

I'm not sure if that's the best way to solve this, therefore I've marked this as "draft" for now.

If anybody has a better idea, please speak up!

I've also asked on the Sphinx-dev mailing list: https://groups.google.com/d/topic/sphinx-dev/aGfJb-TC8PI/discussion


mgeier:
OK, I couldn't come up with a better way ...

If nobody proposes a different way, I'll merge this soon.

PR #430, opened 2020-04-11
Import ABC from collections.abc instead of collections for Python 3 compatibility

### LINKED ISSUE #429: Importing ABC directly from collections is deprecated and will be removed in Python 3.10
https://github.com/spatialaudio/nbsphinx/blob/eed1356f3e68acf1e27be4274e0c6fb7c926910f/src/nbsphinx.py#L931

tirkarthi:
Fixes #429

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -25,7 +25,7 @@ https://nbsphinx.readthedocs.io/
 """
 __version__ = '0.6.0'
 
-import collections
+import collections.abc
 import copy
 import html
 import json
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR!

Sadly, I've just found a bug, and its fix (#431) would get rid of this whole `Sequence` thing.

If that PR is merged, this PR becomes obsolete.

Or do you know a better fix?

mgeier:
OK, I've updated #431.

Thanks again!

tirkarthi:
You're welcome. Thanks for the review :)

PR #431, opened 2020-04-11
nbsphinx_custom_formats: proper handling of strings

mgeier:
This was broken before because strings are sequences!

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -928,7 +928,8 @@ class NotebookParser(rst.Parser):
         else:
             raise NotebookError(
                 'No converter was found for {!r}'.format(srcfile))
-        if isinstance(converter, collections.abc.Sequence):
+        if (isinstance(converter, collections.abc.Sequence) and
+                not isinstance(converter, str)):
             if len(converter) != 2:
                 raise NotebookError(
                     'The values of nbsphinx_custom_formats must be '


tirkarthi:
Why not just add a check for strings like `not isinstance(convertor, str)`? This restricts the convertor to being a list/tuple and a convertor implementing Sequence protocol might be skipped.

mgeier:
I didn't really like the complicated `if`-condition.

But I guess you are right, it's better to allow `Sequence` and make sure it's not `str`.
I've updated this PR: c31716d48a794bce30dfd782ad54038061586412

tirkarthi:
Feel free to cherry pick my PR to use collections.abc .

mgeier:
I've just merged your PR and re-based this one.

tirkarthi:
Given that str was previously allowed it could be an incompatible change based on the API usage. I will leave it up to you to decide on tests, changelog.

mgeier:
> str was previously allowed it could be an incompatible change based on the API usage.

It was indeed allowed, but it wasn't possible to provide a valid string value.
Only two-element sequences were allowed, e.g. `'xy'`, but the second element was then used as `**kwargs`, which would have raised an exception.

So there is no previously successful usage that's now impossible.

In this sense it's not an incompatible change.
Only if you relied on having an exception raised when passing a string. Which you shouldn't have.

> I will leave it up to you to decide on tests

Well tests would be great of course (then I would have spotted this in the first place), but currently there are none for this case ...

Issue #303, opened 2019-05-21
Hide Cell


julesy89:
Hey,

For me hiding the cell's input but not the output did not work. 

nbconvert                     5.5.0
nbformat                      4.4.0
nbsphinx                      0.4.2

Always the input AND the output was hidden. When I move the
{%- if cell.metadata.nbsphinx != 'hidden' %}
line into the {% block input -%} it works.

However, in some cases it might make sense to hide both.

Is it possible to introduce a tag like "hide_input", "hide_output" and "hide"?

Please let me know if you are not able to reproduce the hide behaviour I had.

Julian

mgeier:
> Is it possible to introduce a tag like "hide_input", "hide_output" and "hide"?

Sure, it's possible and it has been requested before, see e.g. #65.

There are even some unfinished pull requests: #86,  #185.

Feel free to work on either one of them or to create a new one!

psychemedia:
Would it make sense for any PR relating to this to use a similar tags to those used for similar effect in Jupyter Book [ [__hiding cells__ docs](https://jupyterbook.org/features/hiding.html#Hiding-page-elements-and-displaying-a-button-to-show-them) and [__removing cells__ docs](https://jupyterbook.org/features/hiding.html#Removing-inputs-from-the-HTML)] to allow some element of portability between publishing workflows?

- to hide a markdown cell or the inputs of a code cell: `{ "tags": [ "hide_input" ] }`
- to hide the outputs of a code cell: `{ "tags": [  "hide_output",  ]}`
- to hide the whole of a code cell: `{ "tags": ["hide_input",  "hide_output" ] }`
- to remove the inputs of a code cell: `{ "tags": [ "remove_input" ] }`
- to remove the outputs of a code cell: `{ "tags": [  "remove_output",  ]}`
- to remove a markdown cell or an entire code cell: `{  "tags": ["remove_cell", ]}`

Tag based filtering in `nbconvert` is possible [[docs](https://nbconvert.readthedocs.io/en/latest/changelog.html#tag-based-element-filtering)] using the `TagRemovePreprocessor` and commands of the form `jupyter nbconvert my.ipynb --TagRemovePreprocessor.remove_cell_tags='{"my_removetag"}'` with the following traitlets:

- `remove_cell_tags`: removes cells
- `remove_input_tags`: removes inputs
- `remove_all_outputs_tag`: removes all outputs
- `remove_single_output_tag`: removes individual outputs

[...]

PR #436, opened 2020-04-15
[WIP] allow toggling inputs and outputs

machow:
This (WIP) PR addresses #303. 

I took a quick look over nbsphinx, and wanted to highlight some challenges and decisions involved.  (I'll clean up / flesh out the doc notebook before wrapping up).

### Changes

* NbOutput directive accepts :class: option, puts it on **outer classes**
* template checks for "hide_input" tag, if found adds the "toggle" class.
* sphinx-toggle added to extensions

### Example

Hidden:

image

Toggled:

image

### Questions

* the toggle class has to be set on the outer container. Right now I use the :class: option, but NbInput uses that option for the inner container. Is it worth adding a toggle option to the NbOutput and NbInput directives?
* does using the metadata.tags "toggle_input" and "toggle_output" sound good?
* other things to look out for?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -13,6 +13,7 @@ extensions = [
     'sphinxcontrib.bibtex',  # for bibliographic references
     'sphinxcontrib.rsvgconverter',  # for SVG->PDF conversion in LaTeX output
     'sphinx_gallery.load_style',  # load CSS for gallery (needs SG >= 0.6)
+    'sphinx_togglebutton',
 ]
 
 # Default language for syntax highlighting in reST and Markdown cells:
--- a/doc/hidden-cells.ipynb
+++ b/doc/hidden-cells.ipynb
@@ -79,6 +79,62 @@
    "source": [
     "This is the cell after the hidden cell."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Toggling Cells\n",
+    "\n",
+    "You can make a cell toggle-able by adding \"hide_input\" to the cell metadata tags\n",
+    "\n",
+    "TODO: check how adding a tag is explained in other parts of doc"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {
+    "tags": [
+     "toggle_input"
+    ]
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "15"
+      ]
+     },
+     "execution_count": 9,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "1 + 14"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {
+    "tags": [
+     "hide_output"
+    ]
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "my output is hidden\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(\"my output is hidden\")"
+   ]
   }
  ],
  "metadata": {
@@ -97,9 +153,9 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.7"
+   "version": "3.6.8"
   }
  },
  "nbformat": 4,
- "nbformat_minor": 2
+ "nbformat_minor": 4
 }
--- a/doc/hidden-cells.ipynb
+++ b/doc/hidden-cells.ipynb
@@ -93,31 +93,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 9,
+   "execution_count": 3,
    "metadata": {
     "tags": [
      "toggle_input"
     ]
    },
-   "outputs": [
-    {
-     "data": {
-      "text/plain": [
-       "15"
-      ]
-     },
-     "execution_count": 9,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
     "1 + 14"
    ]
   },
   {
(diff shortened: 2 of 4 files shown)


mgeier:
Thanks for this PR!

What I *really* like about it, is that the inclusion of the `sphinx_togglebutton` extension is optional.
This way it is open for users to create their own implementation of the HTML/CSS/JavaScript details.

I think the same flexibility should be used for selecting the relevant notebook tags.

I think there are 3 categories of tags:

* affecting code inputs
* affecting code outputs
* affecting whole cells (including code cells, but also others)

On top of those 3 categories, there could be an option to choose whether the toggleable parts are initially visible or not.

Or do you think they should always be initially hidden?

Or should they be initially hidden by default, but with an additional tag this can be changed to initially visible?

I think it would be great to make those 4 (or 3) sets of tags configurable (each accepting either a single string or multiple strings).

Those options should of course have reasonable defaults.
I think we should look around and check what tag names are used in other projects.
If we use sets of tags, we could even support multiple default tags for each category.

> the toggle class (or another class name nbsphinx might want)

I think it would make sense to keep the CSS class name used by `sphinx_togglebutton`.
Otherwise it wouldn't work out-of-the box, right?

We could probably make the CSS class(es) configurable as well, but I think this might be overkill?

> has to be set on the outer container. Right now I use the :class: option, but NbInput uses that option for the inner container. Is it worth adding a toggle option to the NbOutput and NbInput directives?

It would sure be a possibility. Feel free to add options there.

However, I can think of another way to implement this:

You could insert the CSS classes directly into the RST source like this:

```
{%- if ... whatever ... %}
.. rst-class:: toggle
{%- endif %}

{%- if ... some condition ... %}
.. rst-class:: toggle-hidden
{%- endif %}

.. nboutput::
    ...
```

Those `rst-class` classes are automatically applied to the following HTML element.
This way you wouldn't need any change in the implementation of `NbInput` and `NbOutput`.

I'm not sure which approach would be simpler in the end.

> As a general note, a generic metadata option to specify classes to set on either output or input containers would also allow toggling.

I don't understand how this is supposed to work, can you please elaborate?

Finally, there is one thing I'm wondering about:
If there are multiple outputs in one code cell, should they be individually toggleable or all at once?
I guess former would be easier to implement. For latter we would have to wrap all outputs into an additional container, right?

machow:
Thanks for your quick feedback. I'll take a closer look Monday--but wanted to quickly follow-up :).

> What I really like about it, is that the inclusion of the sphinx_togglebutton extension is optional. This way it is open for users to create their own implementation of the HTML/CSS/JavaScript details.

This really stood out to me while looking through possible implementations, too. I think the most remarkable thing is that it's basically a question of how to put a CSS class in the right spot. If users could add CSS classes to parts of a code cell (eg input, output, whole cell), nbsphinx could push this feature to "userland"!

For example, if users can add CSS classes, nbsphinx gets this feature below for free...

> Or should they be initially hidden by default, but with an additional tag this can be changed to initially visible?

Since sphinx-togglebutton has a class to show the toggled section by default!

image

In a sense then, the ability for users to add CSS classes in 3 code cell spots would make up an interface for extensions, like sphinx-togglebutton, which tells...

* **extension writers**: if your extension is configured via CSS classes, it works with nbsphinx.
* **users**: you can use any extensions that work as above, and use their documentation.
* **many feature requesters**: make or modify an extension.

This is what I was stewing on with the comment below :).

>> As a general note, a generic metadata option to specify classes to set on either output or input containers would also allow toggling.

> I don't understand how this is supposed to work, can you please elaborate?

But I realize nbsphinx has the broader jupyter notebook metadata ecosystem to be mindful of. I'm definitely on board with helping out on an implementation that uses a more general interface, or is sphinx-togglebutton specific.

As a concrete example, suppose that nbphinx added tags (or some other metadata field) as classes on both the .nbinput and .nboutput containers (not saying this is the best idea!), and set this in conf.py.

```
# sphinx-togglebutton option
togglebutton_selector = ".nbinput.toggle_input, .nb_output.toggle_output"
```

Then users adding the tags `toggle_input` and `toggle_output` would enable toggling cells.

machow:
cc @choldgraf, since he mentioned 

> All you should need to do is modify the selectors for the toggle buttons so that they map on to nbsphinx HTML structure, as described here: https://sphinx-togglebutton.readthedocs.io/en/latest/#control-the-selector-text-used-to-make-elements-toggle-able

And I'm wondering now if he was thinking about how it would look if nbsphinx added CSS classes to parts of the cell, based on something in metadata (see description above). Curious how well this has worked for jupyterbook!

choldgraf:
@machow quick thoughts from Jupyter Book - I think the tags work nicely, but it's still very early days there. Jupyter doesn't have a strong "tagging" culture yet, I think because only until recently the UIs didn't make it easy to tag things. In Jupyter Book, we are recommending that people use Directives for controlling things like "toggle this content" with the book content, and they use tags to trigger the same behavior just for code cells (since that's the only way I know of to attach metadata/triggers to a code cell)

mgeier:
@machow 

> suppose that nbphinx added tags (or some other metadata field) as classes on both the .nbinput and .nboutput containers (not saying this is the best idea!)

I'm not sure if I understand.
Are you saying that `nbsphinx` would in this scenario take *all* cell tags and unconditionally emit them as (potentially `tag_`-prefixed) CSS classes?

I think it would make more sense to have configurable lists of tags that `nbsphinx` would turn into some pre-defined CSS classes. But I'm open for other suggestions!

sandipde:
Following. I think this would be super useful. Please have an option to add custom tags list too.

mgeier:
@machow Any news on this?

mgeier:
@machow Do you want to continue working on this?

In the meantime there has been a PR (#543) that removes cells instead of hiding/toggling.

It would be great if in the end both things would work, hopefully without interfering with each other ...

nvaytet:
I would also really welcome this addition to `nbsphinx`. Thanks in advance

Issue #437, opened 2020-04-17
struggling with images


parmentelat:
as per your request in https://github.com/chrisjsewell/ipypublish/issues/133

I have given a try to nbsphinx to convert towards readthedocs.io a set of slideshow notebooks
those slides contain images that come in a variety of ways

* as often as possible, using pure markdown `![](pictures/image.png)`
* sometimes as a html tag when I need to resize `<img src="pictures/image.png" height="400px">`
* and even sometimes as more elaborate html tags, when I want to have them float on the right or left hand side of the page; this last usage is admittedly quite rare

in any case it looks like all the instances where I used the 2nd approach just vanish from the final html result after I run sphinx; this is a pain because there are a lot of the pictures, and I don't feel comfy with messing with them all, b/c because it would jeopardize the slideshow rendering, which is its first, and until now foremost, purpose :)

mgeier:
Thanks for the report!

It is known behavior that `<img>` tags don't survive the conversion (and their `src` files are not copied to the HTML directory).

Support for `<img>` tags has been requested quite a few times already:

#52
#284
#321

I've said in https://github.com/spatialaudio/nbsphinx/issues/284#issuecomment-603391556 that I'm not planning to work on this, but probably I should finally implement it (since nobody else seems to volunteer)?

mgeier:
OK, I've given it a try ...

Sadly, I couldn't make the `align` attribute work, but specifying the size should work with #438.

mgeier:
I've just made a new `nbsphinx` release (0.7.0) which includes the `<img>` parsing feature.

As mentioned above, the `align` attribute is not supported, but you can assign a `class` and then handle the alignment with [custom CSS](https://nbsphinx.readthedocs.io/en/0.7.0/custom-css.html).

@parmentelat Does that work for you?

mgeier:
I guess this can be closed.

Issue #321, opened 2019-09-11
GIF is missing


quasiben:
I am trying to use `nbsphinx` with a notebook which contains a GIF.  However, on conversion from ipynb to HTML the GIF is not copied to an images directory nor is there a line in the corresponding HTML which references a GIF.  It seems as if nbsphinx is completely ignoring the file.  Is there some setting I may be missing for the `conf.py` to pull in GIFs ?

IPYNB: https://github.com/dask/dask-tutorial/blob/master/01x_lazy.ipynb
HTML OUTPUT: https://tutorial.dask.org/01x_lazy.html

TomAugspurger:
As a note of followup, the GIF is handled well when markdown syntax is used

```
![img](path/to/gif.gif)
```

It failed to be included in the built HTML when writing HTML directly

```
<img src="path/to/gif"/>
```

mgeier:
@quasiben Does that answer your question?

quasiben:
Not exactly.  Ideally, nbsphinx should pick up the markdown syntax of a gif as it does with a png/jpeg.  But if you want to close that's ok with me

mgeier:
@quasiben No, I don't necessarily want to close this, I just wanted to know if I have to look into this.

Right now, the GIF if shown in your HTML link above, so can you please clarify what's the actual problem?

mgeier:
This may or may not be fixed by #438. I'm closing it anyway due to lack of response.

hakonanes:
> It failed to be included in the built HTML when writing HTML directly
> 
> ```
> <img src="path/to/gif"/>
> ```

This HTML syntax displayed my GIF nicely (bottom of this section): https://kikuchipy.org/en/latest/visualizing_patterns.html#Plot-multiple-signals

---

BTW, thanks *a lot* for `nbsphinx`, it makes writing Sphinx documentation much more fun :tada:

Issue #284, opened 2019-02-12
working with <img> tags in nbsphinx


phaustin:
(all figures below were created from https://github.com/phaustin/nbsphinx_issue/tree/issue)

* jupyter nbconvert is able to produce pretty good-looking html files
  from notebooks using scaled `<img>` elements, e.g.

  - https://github.com/phaustin/nbsphinx_issue/blob/issue/python/img_fig_eq.py#L43
    is converted to:

    https://clouds.eos.ubc.ca/~phil/github/nbsphinx/nbconvert/img_fig_eq_nbconvert.html

* nbsphinx workinng on the same notebook omits the image:

  - https://clouds.eos.ubc.ca/~phil/github/nbsphinx/img_fig_eq.html

    but does work when I use a markdown image tag with a width attribute:

    https://github.com/phaustin/nbsphinx_issue/blob/issue/python/md_fig_eq_aligh.py#L44
    
    produces:  https://clouds.eos.ubc.ca/~phil/github/nbsphinx/md_fig_eq_aligh.html

Since jupyter notebooks ignore attributes like width, height for markdown images, at the moment I'm
producing two versions of each notebook:  one for jupyter ipynb files with `<img>` tags 
and one for html nbsphinx conversion with `![altfig]()` markdown tags.

Is there a way to get nbsphinx to handle img tags the way that nbconvert does?

mgeier:
What does `nbconvert` do when you convert your notebook to LaTeX/PDF?

Both `nbconvert` and `nbsphinx` use `pandoc` to parse the Markdown text, but I guess the difference is that `nbconvert` converts directly to HTML (or LaTeX), while `nbsphinx` converts to RST first.

There may be a way to force `pandoc` to keep the original `<img>` tag, but it would most likely not work for LaTeX output.

See also #52.

phaustin:
for future googlers -- this is the translation code I'm currently using to go back and forth between jupyter markdown and rst scaled images:

[jupytext#305](https://github.com/mwouts/jupytext/issues/305#issuecomment-515186222)

mluerig:
similar issue here - are you planning on integrating support for images as html elements soon?

phaustin:
This should be possible soon using [myst_nb](https://github.com/ExecutableBookProject/MyST-NB).  This is very close to being a drop-in replacement for nbsphinx, although stil alpha.  (e.g. see /ExecutableBookProject/MyST-NB#98)

mluerig:
[...]

PR #438, opened 2020-04-18
Parse <img> tags

mgeier:
Sadly, with the current Pandoc-based implementation I couldn't make the `align` attribute work. I had to use "inline" images (because Pandoc stores the `` tag in a `RawInline` object), but `:align:` is not supported in reST for inline images.

This fixes (at least partially) #52, #284, #437 and probably #321.

Rendered HTML: https://208-210404706-gh.circle-artifacts.com/0/html/markdown-cells.html#Using-the-HTML-%3Cimg%3E-tag

Rendered PDF: https://208-210404706-gh.circle-artifacts.com/0/nbsphinx.pdf#subsubsection.3.5.1

And rendered HTML on RTD: https://nbsphinx--438.org.readthedocs.build/en/438/markdown-cells.html#Using-the-HTML-%3Cimg%3E-tag

### DIFF
--- a/doc/markdown-cells.ipynb
+++ b/doc/markdown-cells.ipynb
@@ -246,8 +246,39 @@
     "\n",
     "Remote image: ![Python logo (remote)](https://www.python.org/static/img/python-logo-large.png)\n",
     "\n",
-    "    ![Python logo (remote)](https://www.python.org/static/img/python-logo-large.png)\n",
+    "    ![Python logo (remote)](https://www.python.org/static/img/python-logo-large.png)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Using the HTML `<img>` tag\n",
     "\n",
+    "The aforementioned Markdown syntax for including images\n",
+    "doesn't allow specifying the image size.\n",
+    "\n",
+    "If you want to control the size of the included image,\n",
+    "you can use the HTML\n",
+    "[\\<img\\>](https://www.w3.org/TR/html52/semantics-embedded-content.html#the-img-element)\n",
+    "element with the `width` attribute like this:\n",
+    "\n",
+    "```html\n",
+    "<img src=\"images/notebook_icon.png\" alt=\"Jupyter notebook icon\" width=\"300\">\n",
+    "```\n",
+    "\n",
+    "<img src=\"images/notebook_icon.png\" alt=\"Jupyter notebook icon\" width=\"300\">\n",
+    "\n",
+    "In addition to the `src`, `alt`, `width` and `height` attributes,\n",
+    "you can also use the `class` attribute,\n",
+    "which is simply forwarded to the HTML output (and ignored in LaTeX output).\n",
+    "All other attributes are ignored."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "### SVG support for LaTeX\n",
     "\n",
     "LaTeX doesn't support SVG images, but there are Sphinx extensions that can be used for automatically converting SVG images for inclusion in LaTeX output.\n",
@@ -443,8 +474,7 @@
     "\n",
     "```\n",
     "[beginning of this section](#Links-to-Other-Notebooks)\n",
-    "```",
-    "\n",
+    "```\n",
     "It's also possible to create a\n",
     "[link to the beginning of the current page](#),\n",
     "by simply using a `#` character:\n",
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -33,6 +33,7 @@ import os
 import re
 import subprocess
 from urllib.parse import unquote
+import uuid
 
 import docutils
 from docutils.parsers import rst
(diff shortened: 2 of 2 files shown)


jayaddison:
Hi @mgeier - an unusual quirk of this feature is that when an `` element _without_ an `alt` attribute is processed from a notebook, `docutils` picks up the emitted inline image substitution and uses the 'name' of that substitution -- the `UUID4` value -- as the `alt`-text.

I guess it seemed reasonable for `docutils` to pick the substitution-name as fallback-alt-text because usually subsitution names are somewhat-human-readable, but in the case of these ephemeral substitutions it causes document reproducibility problems (ref: [Debian bug #1064575](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1064575)) and I doubt it's useful for accessibility either.

Do you have any thoughts on how we could/should handle that better?

mgeier:
Thanks for the report @jayaddison!
I wasn't aware of that behavior.

But doesn't that sound more like a bug in `docutils`?
Why would it set the `alt` attribute to anything if nothing is given?

> I guess it seemed reasonable for `docutils` to pick the substitution-name as fallback-alt-text

Maybe it seemed reasonable to someone, but it isn't.
The whole concept of a "fallback-alt-text" doesn't make any sense. The `alt` text should be written specifically for the purpose, because it is supposed to describe that image for people who can't see it. If no intentional text is available, the `alt` attribute should be omitted.

> I doubt it's useful for accessibility either.

Definitely not.
It's actively harmful.

----

Maybe related: #500, #162, https://github.com/jgm/pandoc/issues/6194

jayaddison:
You're welcome @mgeier - thanks for the considered response.

> > I guess it seemed reasonable for `docutils` to pick the substitution-name as fallback-alt-text
> 
> Maybe it seemed reasonable to someone, but it isn't. The whole concept of a "fallback-alt-text" doesn't make any sense. The `alt` text should be written specifically for the purpose, because it is supposed to describe that image for people who can't see it. If no intentional text is available, the `alt` attribute should be omitted.

Yep, `alt`-text is best written for-purpose.  When that doesn't happen, though, it'd be nice not to introduce accessibility/comprehensibility regressions or other unintended side-effects.

People are capable of making the most of the info provided to them - but UUIDs are -- by-design -- low-information, while human-language labels are likely (hopefully) to be more meaningful.

Are `pandoc` and `docutils` the only two downstream processors for `nbsphinx`'s markdown cell processing, or are there other places to check for compatibility for any suggested changes?

mgeier:
> Yep, `alt`-text is best written for-purpose. When that doesn't happen, though, it'd be nice not to introduce accessibility/comprehensibility regressions or other unintended side-effects.

Exactly.
AFAIK, the only reasonable thing to do in absence of an intentional `alt` text, is omitting the `alt` attribute in the generated HTML.

> People are capable of making the most of the info provided to them - but UUIDs are -- by-design -- low-information, while human-language labels are likely (hopefully) to be more meaningful.

They might be more meaningful, but that doesn't necessarily mean that they are suitable as `alt` text.
The `image` directive has an `:alt:` option, and if that's not specified, `docutils` shouldn't emit an `alt` attribute either, IMHO.

> Are `pandoc` and `docutils` the only two downstream processors for `nbsphinx`'s markdown cell processing, or are there other places to check for compatibility for any suggested changes?

Yes, Markdown cells are processed with Pandoc via `markdown2rst`:

https://github.com/spatialaudio/nbsphinx/blob/6145052c4265d6ed4eef19c4a26c3de74d3b4d1e/src/nbsphinx/__init__.py#L180

... which is defined here:

https://github.com/spatialaudio/nbsphinx/blob/6145052c4265d6ed4eef19c4a26c3de74d3b4d1e/src/nbsphinx/__init__.py#L959

The Markdown cells are converted to reST, which is then parsed by `docutils`.
Note that I consider the detour via reST as a temporary solution, see #36.

The intermediate reST document for `` looks something like this:

```rst
|38d18b0a1469490f86f4570a48f1693d|

...

.. |38d18b0a1469490f86f4570a48f1693d| image:: images/notebook_icon.png
    :width: 300
    :class: no-scaled-link
```

With `alt` attribute, i.e. `Jupyter notebook icon`, it looks like this:

```rst
|0689d12a29f54e41a224a6998da89e99|

...

.. |0689d12a29f54e41a224a6998da89e99| image:: images/notebook_icon.png
    :alt: Jupyter notebook icon
    :width: 300
    :class: no-scaled-link
```

jayaddison:
Thanks @mgeier - yep, I agree with your assessment, and the illustrative examples of the intermediate rST format look correct too.

> AFAIK, the only reasonable thing to do in absence of an intentional alt text, is omitting the alt attribute in the generated HTML.

I've been wondering about -- but haven't yet confirmed -- two ways to do this on the `docutils` -> `Sphinx` path, the output I currently care about (although I don't want to forget the `pandoc` path too): either begin including an empty `:alt:` option in the intermediate rST (ugly, and not perfect - `docutils` would emit an `alt=""` attribute on the HTML, I believe, but could be updated to drop those), AND/OR remove the fallback-`alt`-text logic in `docutils` entirely.

mgeier:
> ugly, and not perfect - `docutils` would emit an `alt=""` attribute on the HTML, I believe, but could be updated to drop those),

That might be very problematic!

The important thing to know is that `alt=""` has very different semantics compared to omitting the `alt` attribute:

`alt=""` means that the image is purely decorative: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt#decorative_images
This in turn means that screen readers will not even mention it.

A non-existing `alt` attribute means that the author was too lazy to provide help.
This in turn means that screen readers will try to provide other helpful information about the image, for example they might read the file name instead (but different screen readers might behave differently).

Therefore, `alt=""` should never be used as an automatic fallback, because it will essentially hide the image from users of screen readers.

> AND/OR remove the fallback-`alt`-text logic in `docutils` entirely.

Yes, as I said above (https://github.com/spatialaudio/nbsphinx/pull/438#issuecomment-1984385064), the only reasonable fallback is to completely omit the `alt` attribute from the generated HTML.

jayaddison:
Thank you @mgeier - I wasn't previously aware of the meaning of an empty-string `alt` attribute.  In that case, I agree with omitting the attribute from the downstream HTML when none was configured at-source.

jayaddison:
I began looking at `docutils` to try to figure out what changes might be required -- but then decided to look at the history of the relevant lines.

It seems that the `alt` option handling changed here: https://sourceforge.net/p/docutils/code/1236

In particular: the logic that only sends an `alt=` kwarg to `self.directive(match, ...)` if the `alt` option is set on the parent element is roughly the change that I had in mind.

jayaddison:
> It seems that the `alt` option handling changed here: https://sourceforge.net/p/docutils/code/1236

Nope, ignore that; the test coverage of `alt` attributes didn't change, so I believe that revision is irrelevant.

In any case: I'll try to develop a `docutils` patch for this and may take that into a thread on the relevant `docutils` project mailing list(s).

jayaddison:
> In any case: I'll try to develop a `docutils` patch for this and may take that into a thread on the relevant `docutils` project mailing list(s).

Mailing list discussion/question opened here: https://sourceforge.net/p/docutils/mailman/message/58746996/

mgeier:
Thanks for taking care of this! Let's hope for a positive response.

PR #446, opened 2020-04-26
Check for '%s' in prompt template

### LINKED ISSUE #442: Error if %s is not present in prompt string format
Hey,

I was trying to hide the prompt by setting an empty string to `nbsphinx_input_prompt`. This results in a `TypeError` on:

https://github.com/spatialaudio/nbsphinx/blob/77f305b071b2045444602167510c1184dd3e6e21/src/nbsphinx.py#L1065

The blame is the `%` format. Would you consider switching to `format` asking user to set the string as `"In [{}]"` rather than `"In [%s]"`. This will be more robust, see:

```py
In [6]: "hello" % "dummy"
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-a2bd27e7e4d9> in <module>
----> 1 "hello" % "dummy"

TypeError: not all arguments converted during string formatting

In [7]: "hello".format("dummy")
Out[7]: 'hello'

In [8]: "hello {}".format(12)
Out[8]: 'hello 12'
```

Another possibility would be to test for `%s`.

Note: I use that [workaround](https://github.com/spatialaudio/nbsphinx/pull/185#issuecomment-393260779) for my initial purpose.

fcollonval:
This fixes #442.

@mgeier yes this was what I meant. Here is a proposal.

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -1076,7 +1076,10 @@ def _create_code_nodes(directive):
 
     outer_node = docutils.nodes.container(classes=outer_classes)
     if execution_count:
-        prompt = prompt_template % (execution_count,)
+        if '%s' in prompt_template:
+            prompt = prompt_template % (execution_count,)
+        else:
+            prompt = prompt_template.format(execution_count)
         prompt_node = docutils.nodes.literal_block(
             prompt, prompt, language='none', classes=['prompt'])
     else:


mgeier:
Thanks, I think this would work. There can be a few theoretical edge cases where this wouldn't work, but I think nothing that's problematic in practice.

But ... is this feature really necessary?

You were suggesting this in #442 because you wanted to use it to remove the prompts.
However, this is not a good way to remove the prompts anyway. If you want to remove the prompts, you should use something like that: https://nbsphinx.readthedocs.io/en/latest/custom-css.html

Is there any other situation where this PR would be needed?

fcollonval:
Good question indeed. The only case I see is if the user wants to print something generic like `"In"` and `"Out"` next to each cells. But is it really useful... no sure. So I let you judge on this one and if you choose to simply close this, not problem for me (less trick, less bugs 😉 - especially if you already have some edge cases in mind.)

mgeier:
> The only case I see is if the user wants to print something generic like `"In"` and `"Out"` next to each cells. But is it really useful... no sure.

I don't think so.

If you don't need it, there is no reason to implement this right now.

If anybody needs this at some point (which I don't think will happen), we can merge it then.

> less trick, less bugs

Very true!

> especially if you already have some edge cases in mind.

Those are only very vague edge cases ... a user might use different placeholders instead of `%s`. For example `%r`. Or `%x` if they want hexadecimal numbers!

mgeier:
I'm closing this for now, but if a real use case comes up, please let me know!

PR #451, opened 2020-04-30
Insert empty comment to make sure reST directives are closed

### LINKED ISSUE #450: Invalid reStructuredText for markdown cell with image followed by cell with leading block quote
If a notebook has consecutive markdown cells like:

1. one contains an image or any other structure that results in a reStructuredText (rST) directive at the end of the rST for that cell, including raw HTML like `<table></table>`
2. one that starts with a block quote `> ...`

Then the resulting rST is invalid, because the block quote text gets included with the content of the directive, since it's only marked via indentation (https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#block-quotes).

Reproducer: https://github.com/huonw/nbsphinx-consecutive-cell-reproducer 
- Clone that repo
- Open `_build/html/image_quote_more.html` (or view it without styling at <https://htmlpreview.github.io/?https://github.com/huonw/nbsphinx-consecutive-cell-reproducer/blob/master/_build/html/image_quote_more.html> )
- Compare it to the original notebook in the top level `image_quote_more.ipynb`: https://github.com/huonw/nbsphinx-consecutive-cell-reproducer/blob/master/image_quote_more.ipynb

Notably:

- the first image shows the wrong image (it should be the jupyter logo, not the jupyter hub one)
- the first quote disappears
- the second quote is formatted as raw HTML
- when everything is in a single cell, it all works correctly

See https://github.com/huonw/nbsphinx-consecutive-cell-reproducer/commit/6ea5829191d279c2488a17f1329f54e8e650ecec for exactly what was added over the raw `sphinx-quickstart` output (specifically, I ran `sphinx-quickstart -q -p test -a test -r 1.0.0 -l en`).

Sphinx version: 3.0.3
Nbsphinx version: 0.6.1

This is the same issue as https://github.com/jupyter/nbconvert/issues/1245 which explores the underlying rST in more detail. I'm filing it here too because, AIUI, nbsphinx has its own conversion pipeline.

(Thanks for nbsphinx! It's working really well other than this edge case :+1: )

mgeier:
Fixes #450.

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -230,6 +230,8 @@ RST_TEMPLATE = """
 {%- else %}
 {{ cell | save_attachments or super() | replace_attachments }}
 {% endif %}
+..
+{# Empty comment to make sure the preceding directive (if any) is closed #}
 {% endblock markdowncell %}
 
 
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -97,6 +97,8 @@ RST_TEMPLATE = """
 {% block any_cell %}
 {%- if cell.metadata.nbsphinx != 'hidden' %}
 {{ super() }}
+..
+{# Empty comment to make sure the preceding directive (if any) is closed #}
 {% endif %}
 {%- endblock any_cell %}
 
@@ -201,6 +203,8 @@ RST_TEMPLATE = """
 
 
 {% block nboutput -%}
+..
+{# Empty comment to make sure the preceding directive (if any) is closed #}
 {%- set html_datatype, latex_datatype = output | get_output_type %}
 {%- if html_datatype == latex_datatype %}
 {{ insert_nboutput(html_datatype, output, cell) }}
@@ -230,8 +234,6 @@ RST_TEMPLATE = """
 {%- else %}
 {{ cell | save_attachments or super() | replace_attachments }}
 {% endif %}
-..
-{# Empty comment to make sure the preceding directive (if any) is closed #}
 {% endblock markdowncell %}
 
 
@@ -254,12 +256,8 @@ RST_TEMPLATE = """
 
 {{ cell.source | indent }}
 {%- elif raw_mimetype == 'text/markdown' %}
-..
-{# Empty comment to make sure the preceding directive (if any) is closed #}
 {{ cell.source | markdown2rst }}
 {%- elif raw_mimetype == 'text/restructuredtext' %}
-..
-{# Empty comment to make sure the preceding directive (if any) is closed #}
 {{ cell.source }}
 {% endif %}
 {% endblock rawcell %}


huonw:
I think an edge case that wasn't directly covered by the report in #450 is a code cell with output followed by a cell with a leading quote.

I added a notebook like this to the reproducer repo from #450: https://github.com/huonw/nbsphinx-consecutive-cell-reproducer/commit/47dd814c57c30ffd1b1bf78775731cb7e96d12c5 .

Notebook: https://github.com/huonw/nbsphinx-consecutive-cell-reproducer/blob/47dd814c57c30ffd1b1bf78775731cb7e96d12c5/code_quote.ipynb

Rendered HTML: https://htmlpreview.github.io/?https://github.com/huonw/nbsphinx-consecutive-cell-reproducer/blob/47dd814c57c30ffd1b1bf78775731cb7e96d12c5/_build/html/code_quote.html

Relevant section: https://github.com/huonw/nbsphinx-consecutive-cell-reproducer/blob/47dd814c57c30ffd1b1bf78775731cb7e96d12c5/_build/html/code_quote.html#L267-L273

```html
<div class="output_area docutils container">
    <blockquote>
        <div>
            <div class="highlight"><pre>
some output
</pre></div>
        </div>
    </blockquote>
    <p>starting with a quote (is included as part of the output of the previous cell)</p>
</div>
```

That is, the quote ends up grouped with the output, and isn't even within a `<blockquote>` element.

The relevant section of the rST from `jupyter nbconvert --to rst` looks like:

```rst
.. code:: ipython3

    print("some output")


.. parsed-literal::

    some output


   starting with a quote (is included as part of the output of the
   previous cell)
```

(Notably, it looks like `nbsphinx` handles the `display(Markdown("> ..."))` case correctly, but `jupyter nbconvert --to rst` does not. 👍 )

Given all this, I don't think the change in this PR will quite handle this case? I can think of a few possible approaches:

- a comment at the start of each markdown cell
- a comment at the end of every cell type
- a comment at the start and end of every cell type

mgeier:
Thanks for caring about the edge cases, I very much appreciate this!

The new example is worse, because it doesn't cause a warning (except for the missing title, which is unrelated).

> a comment at the start and end of every cell type

I don't think the start *and* end is necessary, one of them should suffice.

I've re-discovered the `any_cell` block in the template, which is great for this.

And then I guess we should insert a comment between code input and output and, more importantly, between outputs, if there are multiple.

Those empty comments don't hurt anyone, so it doesn't really matter if we use too many of them.
But we might as well insert the correct amount while we are thinking about it.

I've added a commit: 39d6b9ee6ac4917a2e1b3217fe47e9cdc083a3b3

AFAICT, your example looks fine with it, do you see any further problems?

mgeier:
Thanks @huonw!

Issue #452, opened 2020-05-05
Output images are fractionally too large for their container, in Safari


huonw:
A cell that outputs an image, such as a `IPython.display.Image` or `matplotlib` output, results in an `img` tag that is very slightly larger than the surrounding `div.output_area`. This results the image scrolling slightly, and capturing the mouse when scrolling through the notebook (notice the scrollbars appearing to the right of each image in the following gif, and the "bouncing" as each one hits the bottom of its scroll).

https://nbsphinx.readthedocs.io/en/0.6.1/code-cells.html#Local-Image-Files

![nbsphinx-scrolling-issue](https://user-images.githubusercontent.com/1203825/81041574-89e57400-8ef1-11ea-8406-0a6de84d7c79.gif)

The following description focuses on the first image ([7]), which has HTML:

```html
<div class="output_area docutils container">
<img alt="_images/code-cells_20_0.png" src="_images/code-cells_20_0.png">
</div>
```

According to browser developer tools (Firefox is similar to Chrome):

| element | Safari computed height | Safari `.clientHeight` | Chrome computed height | Chrome `.clientHeight` | 
|---|---|---|---|---|
| `img` | 137.59375px | 138 | 137.594px | 138 |
| `div` | 137px | 137 | 137.594px | 138 |

The client heights were calculated with the following in the console:
```javascript
$("#Local-Image-Files > div:nth-child(3) > div.output_area.docutils.container")[0].clientHeight
$("#Local-Image-Files > div:nth-child(3) > div.output_area.docutils.container > img")[0].clientHeight
```

Browser versions:

- Safari: 12.1.1 (14607.2.6.1.1)
- Chrome: 81.0.4044.129
- Firefox: 68.0.1

mgeier:
That's interesting, any idea what we can do about this?

huonw:
I'm not really sure what the best fix is, but it seems like it may be caused by

https://github.com/spatialaudio/nbsphinx/blob/d41f7cd00b95667258692ffaaf1dcd7b2f5216aa/src/nbsphinx.py#L602-L609

that results in a fractional amount of padding (4.800000190734863px, in Safari, 4.8px in Chrome). As I understand it, the behaviour of sub-pixel rounding isn't specified as part of CSS. I've opened https://github.com/spatialaudio/nbsphinx/pull/454 that switches to rounding to the closest pixel value (5px) in that case.

mgeier:
Wow, thanks for finding this out!

PR #454, opened 2020-05-05
Use px padding to avoid Safari img-too-large scrolling

huonw:
The rem padding causes `` image elements to have fractional heights, like
137.59375px for the [7] image in
<https://nbsphinx.readthedocs.io/en/0.6.1/code-cells.html#Local-Image-Files>. In
Safari (and only Safari), this means that the "layout" height (`.clientHeight`)
of the `img` tag being 138px, while the height of the surrounding `div` is
137px. The slightly overlap means that the image scrolls slightly, causing
unsightly and disconcerting scroll bars to appear, and making scrolling "jerky"
as the mouse gets captured briefly.

AIUI, the behaviour of sub-pixel round isn't specified as part of CSS, and there
being behaviour differences isn't necessarily unexpected.

(This seems to only apply to Safari, other browsers like Chrome and Firefox have
both the `img` and `div` have height 138px.)

Fixes: #452

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -605,7 +605,7 @@ div.nboutput.container div.output_area div[class*=highlight].math,
 div.nboutput.container div.output_area.rendered_html,
 div.nboutput.container div.output_area > div.output_javascript,
 div.nboutput.container div.output_area:not(.rendered_html) > img{
-    padding: 0.3rem;
+    padding: 5px;
 }
 
 /* fix copybtn overflow problem in chromium (needed for 'sphinx_copybutton') */


mgeier:
Thanks for the PR!

I guess you should fix this one as well:

https://github.com/spatialaudio/nbsphinx/blob/d41f7cd00b95667258692ffaaf1dcd7b2f5216aa/src/nbsphinx.py#L490

Right?

mgeier:
I've changed the prompt padding in 4eda0ba179032e546620b5b7a7c9da0db3766f13.

Thanks again, @huonw!

huonw:
Thanks for merging and adjusting that other spot!

PR #455, opened 2020-05-06
Allow using alternative markdown renderers

huonw:
Shelling out to pandoc twice for each markdown cell can be very slow. Using a single-process/in-memory renderer for any markdown can be noticably faster.

This adds support for using [commonmark-py](https://commonmarkpy.readthedocs.io/en/latest/) via `nbsphinx_markdown_renderer = "commonmark".

For instance, builds of https://github.com/stellargraph/stellargraph's docs go from ~2 minutes to 40 seconds (https://github.com/stellargraph/stellargraph/pull/1517).

This is a draft PR because there's several open questions, that would be good to collaborate on/clarify before I spend more effort on this:

- is this a useful feature at all?
- is there an alternative (or additional) Python library we can use? Preferably one that covers more common extensions like tables (https://github.com/readthedocs/commonmark.py/issues/28) and equations (https://github.com/readthedocs/commonmark.py/issues/49). I briefly investigated:
  - [mistune](https://mistune.readthedocs.io):  the rST support and table extensions seem to only be available in the 2.0.0 alphas, but this cannot be installed as `nbconvert` depends on version 0.8.4
  - [Python-Markdown](https://python-markdown.github.io): has a lot of extensions, but seems to only support targeting HTML
  - [m2r](https://github.com/miyakogi/m2r): not maintained
  - [m2rr](https://github.com/qhua948/m2rr): maybe not popular enough?
  - [recommonmark](https://recommonmark.readthedocs.io): unclear to me how to use this for this particular case
- is it worth generalising this even more? e.g. this could allow `nbsphinx_markdown_renderer` to be set to any subclass of `nbsphinx.MarkdownRenderer` (in addition to a `str` name) to let users customise

### DIFF
--- a/doc/usage.ipynb
+++ b/doc/usage.ipynb
@@ -327,6 +327,18 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "### `nbsphinx_markdown_renderer`\n",
+    "\n",
+    "Use a particular way to renderer Markdown. Possible values:\n",
+    "\n",
+    "- `pandoc` (default): a featureful renderer using [Pandoc](https://pandoc.org)\n",
+    "- `commonmark`: a faster renderer that doesn't support some Pandoc extensions, using [the commonmark.py library](https://commonmarkpy.readthedocs.io)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "#### `nbsphinx_output_prompt`\n",
     "\n",
     "Output prompt for code cells. `%s` is replaced by the execution count.\n",
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -717,7 +717,8 @@ class Exporter(nbconvert.RSTExporter):
     """
 
     def __init__(self, execute='auto', kernel_name='', execute_arguments=[],
-                 allow_errors=False, timeout=None, codecell_lexer='none'):
+                 allow_errors=False, timeout=None, codecell_lexer='none',
+                 markdown_renderer='pandoc'):
         """Initialize the Exporter."""
 
         # NB: The following stateful Jinja filters are a hack until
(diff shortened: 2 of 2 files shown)


chrisjsewell:
I would point out: https://github.com/executablebooks/markdown-it-py, which has tables, among other extensions, and is used in https://myst-parser.readthedocs.io

huonw:
> Instead of converting from Markdown to reST and then from reST to a docutils "doctree" object, I would like to do the conversion in one step, see #36.

That issue has been open for a while, has there been more progress other than the comments on it? If not, is it worth pursuing this PR as an intermediate step?

One concern that I guess you might have is migration: at the moment I guess switching to docutils directly would be unobservable to a user (except for differences to Pandoc's rendering), whereas with this change (with the arbitrary-function version) it would be observable to users who have a custom `def nbsphinx_markdown_renderer(markdown): ...` function?

> Ideally this should be a generic extension point, so we wouldn't have to support a limited set of parsers.

If this is generalised, would you want built-in support for any parsers other than Pandoc? At least with the current MD -> reST approach, the function required for other libraries is tiny (e.g. the core of the CommonMark one is two lines, and mistune would be similar).

> I would point out: https://github.com/executablebooks/markdown-it-py, which has tables, among other extensions, and is already used in https://myst-parser.readthedocs.io, https://myst-nb.readthedocs.io/en/latest/ and https://jupyterbook.org 😁

@chrisjsewell I think this doesn't work with the current setup of nbsphinx, for the same reason as Python-Markdown (and mistune < 2): no reStructuredText rendering?

mgeier:
> > Instead of converting from Markdown to reST and then from reST to a docutils "doctree" object, I would like to do the conversion in one step, see #36.
> 
> That issue has been open for a while, has there been more progress other than the comments on it?

Not that I know of, except the things already discussed in this PR.

We would need an appropriate Markdown parser before #36 can be tackled.

The MyST/markdown-it-py project looks promising, but I think it doesn't have all the needed features (yet?).

> If not, is it worth pursuing this PR as an intermediate step?

No, not when it exposes an implementation detail like the intermediate reST representation.

But if it just "invisibly" switches the Markdown parser, I'm all for it!

> One concern that I guess you might have is migration: at the moment I guess switching to docutils directly would be unobservable to a user (except for differences to Pandoc's rendering), whereas with this change (with the arbitrary-function version) it would be observable to users who have a custom `def nbsphinx_markdown_renderer(markdown): ...` function?

Exactly.

> > Ideally this should be a generic extension point, so we wouldn't have to support a limited set of parsers.
> 
> If this is generalised, would you want built-in support for any parsers other than Pandoc?

I guess there should be some parser selected by default. I don't care which one, as long as it has the right features.
And ideally, Pandoc wouldn't be involved at all in the default workflow.

> At least with the current MD -> reST approach, the function required for other libraries is tiny (e.g. the core of the CommonMark one is two lines, and mistune would be similar).

You would still have to take care of the Markdown extensions used in Jupyter.

Do you already have a feature-complete parser to replace Pandoc?

I have no problem with switching from Pandoc to another parser, as long as we don't expose the intermediate reST representation.

Issue #447, opened 2020-04-28
Question: consistent highlighting for rst / markdown code blocks


tshead2:
I have a notebook containing a mixture of code cells, markdown cells containing code:

    ```
    # my code here
    ```

and raw rst cells containing code blocks:

     .. code-block::
    
        # my code here

However, the syntax highlighting for the markdown and raw rst does not match the highlighting for the notebook code cells.  Is there a trick to make them the same?  I assume that if I could see the rst generated by nbsphinx, it'd be clear how to make the manual markup match.

Thanks in advance,
Tim

mgeier:
If you want to see the generated RST, you can try out the `save-rst` branch: https://github.com/spatialaudio/nbsphinx/tree/save-rst.

You can use `highlight_language` to choose the default highlight language: https://nbsphinx.readthedocs.io/en/0.6.1/usage.html#highlight_language.

And in your code blocks, you can specify the language for each instance:

````
```python
# my code here
```
````

```
.. code-block:: python

   # my code here
```

Does that help?

If not, can you please provide screenshots?

tshead2:
Thanks for the quick response.  I've tried many different permutations with-and-without the highlight directive, with-and-without explicitly specifying the language, and-so-on, but I always get the following:

<img width="935" alt="image" src="https://user-images.githubusercontent.com/4997768/80508915-e4049900-8935-11ea-9d7c-718897699672.png">

The first (greenish) block of code comes from a `code-block` directive in a raw RST cell, while the second block is a regular code cell in the source notebook.  This is a local build of the documentation using the _sphinx_rtd_theme_ to match the appearance when it's deployed to readthedocs.io, in case that matters.

Cheers,
Tim

mgeier:
I can't reproduce the problem. I never get those greenish blocks.

Can you please provide a minimal example project?

My code blocks always look like those on the page https://nbsphinx.readthedocs.io/en/0.6.1/a-normal-rst-file.html. They look the same as this one: https://nbsphinx.readthedocs.io/en/0.6.1/markdown-cells.html#Code.

tshead2:
Well, my minimal example (created from scratch, following https://nbsphinx.readthedocs.io/en/0.6.1/usage.html) works correctly, so it must be something in the configuration at my end.  Think I'll copy the minimal conf.py over to my real project, and gradually reintroduce changes until the problem resurfaces.  Apologies for the wild goose chase.

mgeier:
[...]

PR #463, opened 2020-05-12
Don't override Pygments background color in code cells

mgeier:
See #447.

I'm very unsure if that's actually a good idea.

Interestingly, most themes force their own background color, so this would only work with a subset of themes (e.g. with the RTD theme).

Apart from this, I'm hesitating because many people use the Pygments style `'sphinx'` (because it's the one used by `sphinx-quickstart`), which has a quite ugly green background.
If we were to merge this PR, many people would suddenly get a green background on their code cells (which they probably don't want).

Any comments/suggestions?

### DIFF
--- a/doc/usage.ipynb
+++ b/doc/usage.ipynb
@@ -196,6 +196,22 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "#### `pygments_style`\n",
+    "\n",
+    "Use\n",
+    "[pygments_style](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-pygments_style)\n",
+    "to change the color/font theme that's used for syntax highlighting in source code.\n",
+    "\n",
+    "This affects both [code cells](code-cells.ipynb)\n",
+    "and [code blocks in Markdown cells](markdown-cells.ipynb#Code)\n",
+    "(unless overwritten by the\n",
+    "[html_theme](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_theme))."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "#### `suppress_warnings`\n",
     "\n",
     "Warnings can be really helpful to detect small mistakes,\n",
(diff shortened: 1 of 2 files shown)


mgeier:
Good news everyone: `sphinx-quickstart` doesn't use the style `'sphinx'` anymore, see https://github.com/sphinx-doc/sphinx/pull/5282 and https://github.com/sphinx-doc/sphinx/pull/5700.

This makes it less of a problem for new users, but people using an old `conf.py` might still be surprised when their code cells suddenly turn green.

mgeier:
Of all the (about 30) themes mentioned in https://nbsphinx.readthedocs.io/usage.html#HTML-Themes, the Pygments background seems to be only visible on those:

* `bizstyle`
* `dask_sphinx_theme`
* `sphinx_material`
* `sphinx_rtd_theme`
* `traditional`

I'm not sure whether that's a relevant argument for or against this PR, but I think it is remarkable anyway. I find it interesting that not even all the built-in themes respect the Pygments background.

Note that in the `julia` theme the syntax highlighting doesn't seem to work at all, and in the `sphinx_typlog_theme`, the Pygments style seems to be forced (and the `pygments_style` setting seems to be ignored).

mgeier:
In the `bootstrap-astropy` theme, the Pygments background is only visible in the very corners (see the black corners):

image

This is a hint that the CSS still isn't quite right.
Any ideas how to fix this?

DominiqueMakowski:
About the yellow background argument, yes some users might be surprised, although it is ultimately quite a specific issue to search for and easy to fix. A side effect is that these users will automatically be "notified" of the new possibility of further customizing the look of their converted notebooks ☺️ 

About the CSS issue, really sorry but that's completely out of my range! I send some positive vibes tho 🤞

mgeier:
I've created PRs to fix the built-in themes (https://github.com/sphinx-doc/sphinx/pull/7718, https://github.com/sphinx-doc/sphinx/pull/7721) and in the `alabaster` theme (https://github.com/bitprophet/alabaster/pull/169).

PR #485, opened 2020-08-21
DOC: Avoid nested markup

mgeier:
... because this is not supported by reST.


mgeier:
@chrisjsewell Yes, but for now reST is used in `nbsphinx`.

Probably I'll try to re-factor the notebook parsing in order to allow alternative parsers like MyST at some point ... see also #455.

chrisjsewell:
Actually, I did recently ask about supporting this in sphinx, because really it would only need a few changes there, rather than any changes to reST: https://github.com/sphinx-doc/sphinx/issues/8068
But, as you see, it came back a "wont-fix" 😞

PR #496, opened 2020-09-29
FIX: Execution problem with python>= 3.8 on windows

s-weigand:
This PR sets the right event loop policy when using `python>=3.8` on windows.
For more details see https://github.com/jupyter/jupyter_client/issues/583

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -25,7 +25,6 @@ https://nbsphinx.readthedocs.io/
 """
 __version__ = '0.7.1'
 
-import asyncio
 import collections.abc
 import copy
 import html
@@ -50,13 +49,11 @@ import sphinx.transforms.post_transforms.images
 from sphinx.util.matching import patmatch
 import traitlets
 
-# Ensure that notebooks will be executed properly with python>= 3.8 on windows
-# See: https://github.com/jupyter/jupyter_client/issues/583
-if (
-    sys.version_info[0] == 3
-    and sys.version_info[1] >= 8
-    and sys.platform.startswith("win")
-):
+
+if sys.version_info >= (3, 8) and sys.platform == 'win32':
+    # See: https://github.com/jupyter/jupyter_client/issues/583
+    import asyncio
+
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
 _ipynbversion = 4
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -25,6 +25,7 @@ https://nbsphinx.readthedocs.io/
 """
 __version__ = '0.7.1'
 
+import asyncio
 import collections.abc
 import copy
 import html
@@ -32,6 +33,7 @@ import json
 import os
 import re
 import subprocess
+import sys
 from urllib.parse import unquote
 import uuid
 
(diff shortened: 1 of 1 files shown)


src/nbsphinx.py
```diff
@@ -25,13 +25,15 @@
 """
 __version__ = '0.7.1'
 
+import asyncio
```
mgeier, line 28: What about moving this import into the `if`-clause where it is used?

src/nbsphinx.py
```diff
@@ -48,6 +50,15 @@
 from sphinx.util.matching import patmatch
 import traitlets
 
+# Ensure that notebooks will be executed properly with python>= 3.8 on windows
+# See: https://github.com/jupyter/jupyter_client/issues/583
+if (
+    sys.version_info[0] == 3
+    and sys.version_info[1] >= 8
```
mgeier, line 57: Is there a reason for this specific way of checking the version?

What about `sys.version_info >= (3, 8)`?

src/nbsphinx.py
```diff
@@ -48,6 +50,15 @@
...
+    and sys.platform.startswith("win")
```
mgeier, line 58: Why not `sys.platform == 'win32'`?

Is there another platform starting with `win`?

src/nbsphinx.py
```diff
@@ -48,6 +50,15 @@
 from sphinx.util.matching import patmatch
 import traitlets
 
+# Ensure that notebooks will be executed properly with python>= 3.8 on windows
+# See: https://github.com/jupyter/jupyter_client/issues/583
```
mgeier, line 54: Is this URL really relevant?

What about using https://bugs.python.org/issue37373 instead?

src/nbsphinx.py
```diff
@@ -48,6 +50,15 @@
 from sphinx.util.matching import patmatch
 import traitlets
 
+# Ensure that notebooks will be executed properly with python>= 3.8 on windows
```
mgeier, line 53: This comment isn't adding any information.

mgeier (commented):
Thanks for this PR!

I've added a few comments and questions.

src/nbsphinx.py
s-weigand: Just copy pasted the suggested solution, the tupleway is more concise.

src/nbsphinx.py
s-weigand: Just copy pasted the suggested solution and `sys.platform == 'win32'` should be equivalent.
The only other windows platform I know of would be `"cygwin"`, but I don't know if has the same problem.
Gitbash on Windows (MinGW), says it is `"win32"`.

src/nbsphinx.py
s-weigand: IMHO the `jupyter_client` issue is more informative, since besides information about the underlying problem, it contains the convention of leaving the fix to the application, which is why this needs to be addressed in `nbsphinx` and not in `jupyter_client`.

mgeier: OK, I've added the Python issue link to the `jupyter_client` issue.

mgeier:
Thanks for the update! You seem to have committed some unrelated auto-formatting, though.

s-weigand:
Oops, sorry for that, must have added the whole file instead of only the important lines.
Fixed it.

src/nbsphinx.py
```diff
@@ -48,6 +49,13 @@
 from sphinx.util.matching import patmatch
 import traitlets
 
+
+if sys.version_info >= (3, 8) and sys.platform == ("win32"):
```
mgeier, line 53: The parentheses around the string are not necessary (and not helpful, either).

Also, I normally use single quotes, i.e. `'win32'`.

mgeier (commented):
Sorry for being nit-picky, but there is one last thing ..

src/nbsphinx.py
s-weigand: Fixed it 😄

mgeier:
Thanks @s-weigand!

PR #512, opened 2020-10-20
Remove pydata theme because of build problems

mgeier:
This causes a build error, see https://github.com/pandas-dev/pydata-sphinx-theme/issues/177.


mgeier:
The problem has been fixed in the latest release, now it builds without errors.

Issue #525, opened 2021-01-04
Cell Attachment requires preceeding text


alanhdu:
The current cell attachment logic seems to be incomplete. When I try to render the notebook in the gist https://gist.github.com/alanhdu/6f520fc0fa553ed1e06044c7c607ab8c, I get an:

```
WARNING: image file not readable: attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png
```

error.

After debugging, I am fairly confident that the problem is that the markdown:

```
   "source": [
    "![logo-64x64.png](attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png)"
   ]
```

is being translated into the following rst

```rst
.. image:: attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png
   :alt: logo-64x64.png
```

Note that there is no file name preceeding the `image`!  That only seems to appear if the original markdown is something like:

```markdown
This is an image ![logo-64x64.png](attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png)
```

which then gets translated into

```rst
This is an image: |logo-64x64.png|

.. |logo-64x64.png| image:: attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png
```

PR #526, opened 2021-01-04
Fix attachment regex when no markdown text

alanhdu:
Right now, the regex is suited for things like:
```markdown
This is an image ![logo-64x64.png](attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png)
```
which gets translated into:
```rst
This is an image: |logo-64x64.png|

.. |logo-64x64.png| image:: attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png
```

However, if you have an image without an text in front of it like:
```markdown
![logo-64x64.png](attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png)
```

Then the rst looks like:
```rst
.. image:: attachment:cbe9739f-7fcd-4ed5-866f-80d693d30fe8.png
   :alt: logo-64x64.png
```

Crucially, this means that there is no `|<filename.png>| ` to match, which causes the regex to fail.  This is a simple solution which just makes the alt-text matching optional in the regex.

Fixes https://github.com/spatialaudio/nbsphinx/issues/525


alanhdu:
I'm not exactly sure the right way to test this -- I checked manually that this fixes rendering my specific notebook. I added an example of this behavior that to the `markdown-cells.ipynb` notebook -- is that the best way to test this change?

mgeier:
Thanks for this PR!

> is that the best way to test this change?

The "best way" is quite a strong claim, I'm not sure about that, but I think it's a good enough way to test it.
 
>  I haven't been able to build the docs locally based on some `bibtex` configuration error.

I've just merged #524 which should fix (or rather postpone) the error, can you please rebase?

alanhdu:
Rebased!

mgeier:
Thanks, the rendered pages is available at https://nbsphinx--526.org.readthedocs.build/en/526/markdown-cells.html#Cell-Attachments.

doc/markdown-cells.ipynb
```diff
@@ -332,7 +335,9 @@
     "\n",
     "    ![a stick figure](attachment:stickfigure.png)\n",
     "\n",
-    "This is a cell attachment: ![a stick figure](attachment:stickfigure.png)\n",
+    "These are cell attachments: ![a stick figure](attachment:stickfigure.png)\n",
+    "\n",
+    "![logo-64x64.png](attachment:98a753bb-02aa-42e8-81da-6a5c4f9b8eb5.png)\n",
```
mgeier, line 340: Does this file name have some meaning?
What about using a more descriptive name, like e.g. `logo-64x64.png`?

The `alt` text (between the brackets) should be for accessibility, I don't think that `logo-64x64.png` really helps with that, right?

alanhdu:
I believe the test failure is a flaky one:

```
(line 1573) broken    https://developer.mozilla.org/en-US/docs/Tools/Web_Console#Opening_the_Web_Console - Anchor 'Opening_the_Web_Console' not found
```

but when I click on the link, it seems to work fine.

mgeier:
Thanks a lot!

Yes, the test failure is unrelated, I'm trying to fix it in #528.

Issue #535, opened 2021-02-08
footcite support


clonker:
Hi there,
thank you for the great tool you got going on here! I have a request and would also be willing to contribute regarding that if it is of any interest to you.
I am quite heavily using citations inside notebooks (via `<cite data-cite=...>[..]</cite>`) but since I keep my bibliographies separate on a per-document basis, this means creating a unique label prefix for each document. In recent versions sphinxcontrib-bibtex introduced "footbibliographies", where instead of using the global sphinx referencing system, footnotes are used and thus all bibliographies are separate, no more unique labels for me, yay - at least in `rst` (or derivative) documents.
Now I think all that needs to be done is to add an option to the `CitationParser` for it to also handle `data-footcite` attributes.

What do you guys think? Thanks!
Moritz

mgeier:
Sure, that sounds like a great idea, I'm looking forward to a pull request!

clonker:
implemented in PR #536

PR #536, opened 2021-02-08
Add footcite feature

clonker:
Hi,
as mentioned in issue #535, this adds the `data-footcite` attribute to the `CitationParser`. One could check if there are both `data-cite` and `data-footcite` present inside the HTML tag, however I think that is a fairly exotic case. In this implementations' current state, the attribute that comes first would simply take precedence. I took the liberty to update the documentation notebooks accordingly.
Please let me know if you have any suggestions!

**Source**:

footbibliography_source

**Result**:
footbibliography_result

### DIFF
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -176,7 +176,7 @@ There are different ways of handling references, for example you could use the
 sphinxcontrib.bibtex_ extension.
 
 After installing the sphinxcontrib.bibtex_ extension, you have to enable it in
-your ``conf.py`` and select at least one BibTeX file you want to use:
+your ``conf.py`` and select the BibTeX file(s) you want to use:
 
 .. code-block:: python
 
@@ -189,7 +189,7 @@ your ``conf.py`` and select at least one BibTeX file you want to use:
     bibtex_bibfiles = ['my-references.bib']
 
 Afterwards all the references defined in the bibliography file(s) can be used
-throughout the documents as detailed in the following.
+throughout the Jupyter notebooks and other source files as detailed in the following.
 
 .. _standard Sphinx Citations: https://www.sphinx-doc.org/en/master/usage/
     restructuredtext/basics.html#citations
@@ -241,8 +241,8 @@ citations like :footcite:`perez2011python`.
 Also footnote citations can be used within Jupyter notebooks with a special HTML syntax,
 see the section about 
 `footnote citations in Markdown cells <markdown-cells.ipynb#Footnote-citations>`__.
-Footnote citations are restricted to their own document and the assembly of the 
-bibliography is (analogously to citations) invoked with the
+Footnote citations are restricted to their own source file and the assembly of the 
+bibliography is (analogously to normal citations) invoked with the
 
 .. code-block:: rst
 
@@ -254,8 +254,8 @@ look like this (in HTML output):
 .. footbibliography::
 
 In the LaTeX/PDF output, there is no list of references appearing right 
-here or anywhere else in the document. Instead, the footnote citations 
-are placed into the footnotes of their respective pages.
+here. Instead, the footnote citations are placed into the footnotes of 
+their respective pages.
 
 
 Thumbnail Galleries
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -167,36 +167,16 @@ Domain Objects
 
         :ref:`/markdown-cells.ipynb#Links-to-Domain-Objects`
 
-Citations
----------
-
-You could use `standard Sphinx citations`_, but it might be more practical to
-use the sphinxcontrib.bibtex_ extension.
-
-If you install and enable this extension,
-you can create citations like :cite:`perez2011python`:
-
-.. code-block:: rst
-
-    :cite:`perez2011python`
-
-You can create similar citations in Jupyter notebooks with a special HTML
-syntax, see the section about `citations in Markdown cells`_.
-
-.. _standard Sphinx Citations: https://www.sphinx-doc.org/en/master/usage/
-    restructuredtext/basics.html#citations
-.. _sphinxcontrib.bibtex: https://sphinxcontrib-bibtex.readthedocs.io/
-.. _citations in Markdown cells: markdown-cells.ipynb#Citations
-
-For those citations to work, you also need to specify a BibTeX file,
-as explained in the next section.
-
 
 References
 ----------
 
+There are different ways of handling references, for example you could use the 
+`standard Sphinx citations`_, but it might be more practical to use the 
+sphinxcontrib.bibtex_ extension.
+
 After installing the sphinxcontrib.bibtex_ extension, you have to enable it in
-your ``conf.py`` and select the BibTeX file(s) you want to use:
+your ``conf.py`` and select at least one BibTeX file you want to use:
 
 .. code-block:: python
 
@@ -208,7 +188,27 @@ your ``conf.py`` and select the BibTeX file(s) you want to use:
 
     bibtex_bibfiles = ['my-references.bib']
 
-Then you can create a list of references in any reStructuredText file
+Afterwards all the references defined in the bibliography file(s) can be used
+throughout the documents as detailed in the following.
+
+.. _standard Sphinx Citations: https://www.sphinx-doc.org/en/master/usage/
+    restructuredtext/basics.html#citations
+.. _sphinxcontrib.bibtex: https://sphinxcontrib-bibtex.readthedocs.io/
+
+Citations
+^^^^^^^^^
+
+You can create citations like :cite:`perez2011python`:
+
+.. code-block:: rst
+
+    :cite:`perez2011python`
+
+You can create similar citations in Jupyter notebooks with a special HTML
+syntax, see the section about 
+`citations in Markdown cells <markdown-cells.ipynb#Citations>`__.
+
+You can create a list of references in any reStructuredText file
 (or `reST cell <raw-cells.ipynb#reST>`_ in a notebook) like this:
 
 .. code-block:: rst
@@ -227,6 +227,9 @@ but at the end of the document.
 For a possible work-around,
 see https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/156.
 
+Footnote citations
+^^^^^^^^^^^^^^^^^^
+
 With a sphinxcontrib.bibtex_ version of ``>= 2.0.0`` it is 
 possible to create footnote bibliographies with footnote 
 citations like :footcite:`perez2011python`.
@@ -235,8 +238,11 @@ citations like :footcite:`perez2011python`.
 
     :footcite:`perez2011python`
 
-These citations are restricted to their own document and the assembly of the
-bibliography is invoked with the
+Also footnote citations can be used within Jupyter notebooks with a special HTML syntax,
+see the section about 
+`footnote citations in Markdown cells <markdown-cells.ipynb#Footnote-citations>`__.
+Footnote citations are restricted to their own document and the assembly of the 
+bibliography is (analogously to citations) invoked with the
 
 .. code-block:: rst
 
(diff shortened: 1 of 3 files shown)


clonker:
I am not sure about the CI failures but they seem unrelated to this proposed set of changes

mgeier:
Thanks for this PR!
I will have a closer look soon.

Rendered: https://nbsphinx--536.org.readthedocs.build/en/536/markdown-cells.html#Citations

Is this expected to not work with LaTeX/PDF output?

See: https://632-46379698-gh.circle-artifacts.com/0/nbsphinx.pdf#subsection.3.2

If that's expected, can you please mention that in the docs?

> I am not sure about the CI failures but they seem unrelated to this proposed set of changes

Yeah, I think they are unrelated.
The `linkcheck` error was just temporary, but the other error is related to `sphinxcontrib-bibtex`, I will look into that later.

clonker:
Interesting, I hadn't checked the PDF output. It seems to me that the example citation was correctly placed into the footnotes - just the directive itself didn't receive any specific output. Perhaps it should be mentioned more explicitly that this does not render a bibliography in the classical sense but rather uses footnotes, especially in LaTeX/PDF output. 
Anyways it seems a bit odd to me that the doi got another footnote within the footnote (ie footnotes 134 and 135). I am not sure right now whether this is expected behavior and/or caused by sphinxcontrib-biblatex or by the way things are extracted from the ntoebook.

clonker:
So the double footnote I was worried about is caused by the `latex_show_urls = 'footnote'` option in `conf.py`. The same effect with raw reST files and just sphinxcontrib-biblatex, so this is somewhat expected behavior.
What would you say if I added a remark to the documentation that the directive only renders as bibliography in html output, in LaTeX/PDF output it causes the footnote citations to be placed into the footnotes depending on the page they're on?

mgeier:
Yes, the additional footnote is strange, and it would be nice to get rid of it, but it isn't the fault of this PR.

> It seems to me that the example citation was correctly placed into the footnotes - just the directive itself didn't receive any specific output.

Oh, yes, I didn't see the footnote.

> Perhaps it should be mentioned more explicitly that this does not render a bibliography in the classical sense but rather uses footnotes, especially in LaTeX/PDF output.

Yes, please.

> What would you say if I added a remark to the documentation that the directive only renders as bibliography in html output, in LaTeX/PDF output it causes the footnote citations to be placed into the footnotes depending on the page they're on?

Yes please!

There are already a few things that are different in HTML and LaTeX output. That's fine, but I'd like to mention all differences in the docs. I've also mentioned the different behavior there: https://nbsphinx.readthedocs.io/en/0.8.1/a-normal-rst-file.html#references

mgeier:
I've come up with a different way to use this, without needing raw reST cells:

```python
nbsphinx_epilog = r"""
.. footbibliography::
"""
```

What do you think about this?

It's probably not appropriate to use this in the `nbsphinx` docs, but you could probably mention it?

clonker:
I'll get on it then, thank you for the feedback :slightly_smiling_face: 
I think the epilog is a brilliant idea, makes the whole thing so much cleaner and easier to use, for sure worth mentioning!

clonker:
I think I have incorporated all the suggestions, please let me know if there is anything that I missed or that can be improved.

mgeier:
Thanks for the update!

I think it might be less confusing if the new footcite feature gets a separate section in the "Markdown Cells" notebook.
Either as sub-section of "Citations" or as sibling section right after it.
What do you think?

The structure on the "normal reST files" page is also a bit confusing: There is a "citations" section followed by "references", which then includes some text about (foot) "citations" again.

AFAIU, `bibtex_bibfiles` is needed for both "normal" and "foot" citations, right?

So probably it would make sense to start with a "references" section explaining the use of `bibtex_bibfiles` and then add two sub-sections "Citations" and "Foot Citations", or something like that?

clonker:
> I think it might be less confusing if the new footcite feature gets a separate section in the "Markdown Cells" notebook.
> Either as sub-section of "Citations" or as sibling section right after it.
> What do you think?

That would help with clarity, I was just reluctant to make bigger changes to the documentation. Since it is a special case of citations perhaps a sub-section makes here more sense than sibling section. 
 
> The structure on the "normal reST files" page is also a bit confusing: There is a "citations" section followed by "references", which then includes some text about (foot) "citations" again.
> AFAIU, `bibtex_bibfiles` is needed for both "normal" and "foot" citations, right?
> So probably it would make sense to start with a "references" section explaining the use of `bibtex_bibfiles` and then add two sub-sections "Citations" and "Foot Citations", or something like that?

Fully agree, let me make an update to this pr tomorrow and we'll take it from there.

clonker:
I restructured the document as you suggested and in the process removed the reference to the unmaintained `sphinx-natbib` extension. The bitbucket repository seems to be gone and there is just one github fork of it with last commit 8 years ago. If you think it should remain in the docs regardless I will re-add it of course. What do you think?

doc/a-normal-rst-file.rst
```diff
@@ -208,7 +188,27 @@ your ``conf.py`` and select the BibTeX file(s) you want to use:
 
     bibtex_bibfiles = ['my-references.bib']
 
-Then you can create a list of references in any reStructuredText file
+Afterwards all the references defined in the bibliography file(s) can be used
+throughout the documents as detailed in the following.
```
mgeier, line 192: I'm not sure if the reader understands what the term "documents" refers to.
I think it is the official Sphinx/`docutils` term, but wouldn't "source files" or "Jupyter notebooks and other source files" be more clear?

doc/a-normal-rst-file.rst
```diff
@@ -227,9 +227,35 @@ but at the end of the document.
...
+citations like :footcite:`perez2011python`.
+
+.. code-block:: rst
+
+    :footcite:`perez2011python`
+
+Also footnote citations can be used within Jupyter notebooks with a special HTML syntax,
+see the section about 
+`footnote citations in Markdown cells <markdown-cells.ipynb#Footnote-citations>`__.
+Footnote citations are restricted to their own document and the assembly of the
```
mgeier, line 244: Again, I'm not sure if "document" is the best term to use here.

doc/a-normal-rst-file.rst
```diff
@@ -227,9 +227,35 @@ but at the end of the document.
...
+
+.. code-block:: rst
+
+    :footcite:`perez2011python`
+
+Also footnote citations can be used within Jupyter notebooks with a special HTML syntax,
+see the section about 
+`footnote citations in Markdown cells <markdown-cells.ipynb#Footnote-citations>`__.
+Footnote citations are restricted to their own document and the assembly of the 
+bibliography is (analogously to citations) invoked with the
```
mgeier, line 245: What is meant by "citations"? Do you mean "normal" citations as opposed to footnote citations?

doc/a-normal-rst-file.rst
```diff
@@ -227,9 +227,35 @@ but at the end of the document.
...
+
+    .. footbibliography::
+
+directive. For example, a footnote bibliography might 
+look like this (in HTML output):
+
+.. footbibliography::
+
+In the LaTeX/PDF output, there is no list of references appearing right 
+here or anywhere else in the document. Instead, the footnote citations
```
mgeier, line 257: Here the use of "document" is even more confusing.

mgeier, line 257: What about this:

```suggestion
here. Instead, the footnote citations 
```

doc/markdown-cells.ipynb
```diff
@@ -204,7 +204,42 @@
     "You'll also have to define a list of references,\n",
     "see [the section about references](a-normal-rst-file.rst#references).\n",
     "\n",
-    "There is also a Notebook extension which may or may not be useful: https://github.com/takluyver/cite2c."
+    "There is also a Notebook extension which may or may not be useful: https://github.com/takluyver/cite2c.\n",
+    "\n",
+    "### Footnote citations\n",
+    "\n",
+    "Since version `2.0.0` of sphinxcontrib-bibtex, [footnote citations](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#role-footcite) are possible. This generates footnotes for all foot-citations up to the point of the bibliography directive, which is typically placed at the end of the document. \n",
```
mgeier, line 211: Again, I'm not sure if readers know what is meant by "document".

doc/markdown-cells.ipynb
```diff
@@ -204,7 +204,42 @@
...
+    "\"\"\"\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
```
mgeier, line 241: I think it would be better to move this cell up a bit, because it is mentioned a few lines before (before `nbsphinx_epilog`).

mgeier (commented):
Thanks for the update!

I just have a few minor questions/comments:

doc/a-normal-rst-file.rst
```diff
@@ -167,36 +167,16 @@ Domain Objects
...
 References
 ----------
 
+There are different ways of handling references, for example you could use the 
+`standard Sphinx citations`_, but it might be more practical to use the 
+sphinxcontrib.bibtex_ extension.
+
 After installing the sphinxcontrib.bibtex_ extension, you have to enable it in
-your ``conf.py`` and select the BibTeX file(s) you want to use:
+your ``conf.py`` and select at least one BibTeX file you want to use:
```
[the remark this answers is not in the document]
clonker: Doesn't change the meaning and also is not particularly important. In my mind it emphasized that there needs to be one bibtex file for the extension to function. I reverted the change :slightly_smiling_face:

doc/a-normal-rst-file.rst
clonker: You are right, this is more clear!

doc/a-normal-rst-file.rst
clonker: Yes that is what i meant, added a "normal" to it

doc/a-normal-rst-file.rst
clonker: Yes that improves understanding, thank you!

doc/markdown-cells.ipynb
clonker: Moved it up a bit :slightly_smiling_face:

clonker:
Thank you for your remarks, I hope everything is in order now - but of course I am happy to incorporate further changes! 
I am not sure how your policy is with respect to the git log, if you like I can squash the commits into one.

doc/a-normal-rst-file.rst
clonker: ... nice suggestion feature by the way, first time i saw it in action :grinning:

mgeier: Yeah, I haven't used it often, but sometimes it is very handy!

doc/markdown-cells.ipynb
```diff
@@ -204,7 +204,49 @@
     "You'll also have to define a list of references,\n",
     "see [the section about references](a-normal-rst-file.rst#references).\n",
     "\n",
-    "There is also a Notebook extension which may or may not be useful: https://github.com/takluyver/cite2c."
+    "There is also a Notebook extension which may or may not be useful: https://github.com/takluyver/cite2c.\n",
+    "\n",
+    "### Footnote citations\n",
+    "\n",
+    "Since version `2.0.0` of sphinxcontrib-bibtex, [footnote citations](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#role-footcite) are possible. This generates footnotes for all foot-citations up to the point of the bibliography directive, which is typically placed at the end of the source file.\n",
```
mgeier, line 211: I think it would be nice to make "bibliography directive" a link to to the appropriate section in the "normal reST file".

mgeier (commented):
Thanks for the update! One last thing ...

mgeier:
> I am not sure how your policy is with respect to the git log, if you like I can squash the commits into one.

Unless the changes are intentionally split into multiple meaningful commits, I normally just squash them when merging.

If you want, you can squash them, but if not it isn't a problem for me to squash them ...

clonker:
The "bibliography directive" now links to the normal rst file "footnote citations" subsection. If it is not a problem for you to squash them then I'll leave that to you :slightly_smiling_face:

mgeier:
Perfect, thanks a lot!

clonker:
Also thanks to you for working with me on this, it'll come in very handy in my project(s)! 🎉

Issue #434, opened 2020-04-13
How to print cell outputs when running sphinx-build in notebooks


dmey:
I need to debug a notebook and looking for a way to print all output to console. I tried to increase verbosity in `sphinx-build` but this did not help. Although the exception is captured and printed to console, nothing before/after (e.g. in a try-except-finally) is printed.

mgeier:
I don't quite understand what you mean.

Can you please provide a minimal example?

dmey:
Assuming you have already set up a Sphinx project with `nbsphinx` add a cell with `print('hello')`, then run `sphinx-build -v`... My expectation was to see `hello` printed to stdout.

mgeier:
OK, I see.

There is currently no way to do that.

Can `nbconvert --execute` do that?

dmey:
> Can nbconvert --execute do that?

Running `jupyter nbconvert --execute <notebook>` does not print outputs. So I guess that this is probably related to https://github.com/jupyter/nbconvert/issues/648.

mgeier:
Yes, it looks like that has been requested for `nbconvert` but there has been no solution yet, but there have been suggestions to use alternative notebook runners.

So once `nbconvert` can do this, `nbsphinx` should be able to implement it, too.

I'm also open to suggestions for using an alternative notebook runner, but I'm not planning on working on that myself.

PR #543, opened 2021-03-02
process cell tags remove-input and remove-output

t-vi:
This PR adds removing cell input and output by using the `remove-input` and `remove-output` tags.

My understanding from #436 is that the behaviour (remove output rather than hide with css) is consistent with Jupyter Book use of the same tags.
This is complimentary to #436 in that it deals with `remove-` tags as opposed to `hide-*`.

Thank you for looking at this, and making nbsphinx in the first place!

### DIFF
--- a/doc/hidden-cells.ipynb
+++ b/doc/hidden-cells.ipynb
@@ -28,7 +28,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "nbsphinx": "hidden"
    },
@@ -47,9 +47,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "42"
+      ]
+     },
+     "execution_count": 2,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
    "source": [
     "answer"
    ]
@@ -79,6 +90,64 @@
    "source": [
     "This is the cell after the hidden cell."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Hidden input and output\n",
+    "\n",
+    "You can also only remove input and output of a notebook using the `remove-input` and `remove-output` cell tags.\n",
+    "This is different from the metadata change for hiding everything, but is compatible with Jupyter Book."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "tags": [
+     "remove-output"
+    ]
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "'Cell with hidden output'"
+      ]
+     },
+     "execution_count": 3,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "\"Cell with hidden output\""
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {
+    "tags": [
+     "remove-input"
+    ]
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "'Did you see the input?'"
+      ]
+     },
+     "execution_count": 4,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "\"Did you see the input?\""
+   ]
   }
  ],
  "metadata": {
@@ -97,7 +166,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.7"
+   "version": "3.9.1+"
   }
  },
  "nbformat": 4,
--- a/doc/hidden-cells.ipynb
+++ b/doc/hidden-cells.ipynb
@@ -28,7 +28,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {
     "nbsphinx": "hidden"
    },
@@ -47,20 +47,29 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "data": {
-      "text/plain": [
-       "42"
-      ]
-     },
-     "execution_count": 2,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
+   "source": [
+    "answer"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Another way to hide sphinx is by adding a `remove-cell` cell tag. This is compatible with Jupyter Book."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "remove-cell"
+    ]
+   },
+   "outputs": [],
    "source": [
     "answer"
    ]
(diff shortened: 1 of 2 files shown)


t-vi:
@mgeier  thank you for making nbsphinx it is just what I need to keep a large part of my documentation in notebooks.

So my use-case is that I have (graphviz) graphs in the documentation that I want to generate from the notebook but I don't want to include the source code in the HTML because it adds nothing in my case (and people can always jump to the notebook if they want).

I hope this can fill a gap, but if this is totally different from what you would have in mind for handling these tags, don't hesitate to reject the PR, too.

mgeier:
Thanks for this PR!

I'm open for this feature, but I think there are some details still to be ironed out.

Can you please remove the outputs from the example notebook?

In your example, the padding after an input cell without output is missing.
So is the padding before an output cell without input.

The implementation is fine, but did you think about re-using the tag-based filtering from `nbconvert`?
Would this be possible?

I think  `remove-input` and `remove-output` are good defaults, but what about allowing configuration options for specifying other tag names (like IIRC it is possible in `nbconvert`)?

> My understanding from #434 is that the behaviour (remove output rather than hide with css) is consistent with Jupyter Book use of the same tags.

Is there a link to some documentation?

> This is complimentary to #434

I don't see the connection.
Do you happen to mean #436?
Or #131?

BTW, did you have a look at  #185 and  #86?

t-vi:
> I don't see the connection.
> Do you happen to mean #436?

Oh sorry. Yes, I changed the text to #436.

> BTW, did you have a look at #185 and #86?

I didn't. So I am aware of `hide_input` but ended up implementing `remove-cell` as a tag (by adding `or "remove-cell" not in ...tags`) in my local branch (which I'd happily add here, too). I must admit that tags are about my limit w.r.t. discoverability (it's bad, but it's the best we have, and standardizing across multiple implementations might be best, so I took the JupyterBook tags based on my impression that it seemed reasonable in #436) the custom metadata is more a way to sell headache pills to me than a user interface (and this is not to take a dig at nbsphinx using them when they were the best thing we had, but just that I had a bad time with them when using them for global things and I think they're best avoided for per-cell things due to the complexity of the UI).

> Can you please remove the outputs from the example notebook?
Will do.

> In your example, the padding after an input cell without output is missing.
> So is the padding before an output cell without input.

> The implementation is fine, but did you think about re-using the tag-based filtering from nbconvert?
> Would this be possible?

This would be very elegant.
So I found [removing cells](https://nbconvert.readthedocs.io/en/latest/removing_cells.html#removing-pieces-of-cells-using-cell-tags) in the nbconvert docs. I tried to adapt the eample there to the style of defining config used by nbsphinx, but it seems to not work for me.
```
            config=traitlets.config.Config({
                'HighlightMagicsPreprocessor': {'enabled': True},
                # Work around https://github.com/jupyter/nbconvert/issues/720:
                'RegexRemovePreprocessor': {'enabled': False},
                'TagRemovePreprocessor': {
                    'enabled': True,
                    'remove_cell_tags': ('remove-cell',),
                    'remove_all_output_tags': ('remove-output',),
                    'remove_input_tags': ('remove-input',),
                },
            }),
```

If you have ideas, going via nbconvert config would seem much more elegant to me.

P.S.: [Our project](https://torchdrift.org/) using nbsphinx and also your insipid theme went public today. Thank you for covering us for our documentation.

mgeier:
Thanks for the update!

> the custom metadata is more a way to sell headache pills to me than a user interface (and this is not to take a dig at nbsphinx using them when they were the best thing we had, but just that I had a bad time with them when using them for global things and I think they're best avoided for per-cell things due to the complexity of the UI).

I agree.
I think I've already added support for tags for all features which were originally using custom cell metadata. If I've missed something, please let me know!

In fact, I would change the documentation to mention the tags first, and then in the end mention the custom cell metadata just as a legacy feature that still works for compatibility but is not recommended for new notebooks.
But we can also do this in a separate PR, if you prefer.

----

Can you please answer my remaining questions from https://github.com/spatialaudio/nbsphinx/pull/543#issuecomment-791614690?

----

> P.S.: [Our project](https://torchdrift.org/) using nbsphinx and also your insipid theme went public today. Thank you for covering us for our documentation.

That's great to hear!

If you have suggestions for improving the theme, please let me know!

t-vi:
> If you have suggestions for improving the theme, please let me know!

The most annoying thing was that Jupyter Notebooks don't need a blank line but nbsphinx generates funny stuff without (seems to be https://github.com/jupyter/nbconvert/issues/917, so not a nbsphinx bug).

Other than that I probably haven't used half the customization that is possible. One thing I thought about is whether to put the logo in the topbar instead of the sidebar. That likely is already possible but I took your advice and focus on the content. I really like how nbsphinx and your insipid theme made that possible!

mgeier:
> > I think remove-input and remove-output are good defaults, but what about allowing configuration options for specifying other tag
> > names (like IIRC it is possible in nbconvert)?
> 
> Personally, I don't really see the value.

I don't have a concrete use case myself, but I could imagine that somebody wants different versions of their docs, once with removed cells and another one where all cells are visible.
Or they are already using a set of tags for something else, and suddenly they want to hide all those tagged cells.

If you look at it from a cost/benefit perspective: It costs basically nothing to expose the configuration options, therefore it doesn't matter if the benefit isn't huge.

Having the tags as literal strings littered around the source code is bad anyway, so at least I would make global variables for them. But making them Sphinx configuration options is just as easy, and to me it is quite an obvious thing to do (even if I don't have a use for that myself).

> Jupyter seems to be unable to [specify common tags](https://github.com/jupyter/nbformat/issues/93),

Yeah, that's even more reason to make them configurable in `nbsphinx`!

> but in the end I see more value with having some sort of standardization and so aligning with Jupyter Book + Hardcoding might be a feature.

Sure, standardization would be great, but as you said, this didn't happen (except for `metadata.jupyter.source_hidden` and `metadata.jupyter.outputs_hidden`: https://nbformat.readthedocs.io/en/latest/format_description.html#cell-metadata).

Jupyter Book is just another third-party package, it isn't in the position to standardize anything.

But if there are no better suggestions, I think it's reasonable to use their settings as our defaults.

> That said, if you intended the question to express a preference for configuration, I'll add that.

Yeah, I guess I did.
Yes, please!

> > > My understanding from #434 is that the behaviour (remove output rather than hide with css) is consistent with Jupyter Book use of the same tags.
> 
> > Is there a link to some documentation?
> 
> https://jupyterbook.org/interactive/hiding.html#removing-code-cell-content
> 
> Would you add that link to the documentation? I can do that when I flip the order of the hide methods in the documentation.

Thanks for the link.
I think that adds no value to our documentation. We have to mention the default tag names anyway.

> I think I have the other questions, but I'd appreciate a hint if not.

I think those were indeed all the questions, but I think this comment hasn't been addressed yet:

"In your example, the padding after an input cell without output is missing.
So is the padding before an output cell without input."

----

> > If you have suggestions for improving the theme, please let me know!
> 
> The most annoying thing was that Jupyter Notebooks don't need a blank line but nbsphinx generates funny stuff without (seems to be [jupyter/nbconvert#917](https://github.com/jupyter/nbconvert/issues/917), so not a nbsphinx bug).

Yeah, the problem is that there are different Markdown implementation with slightly different behavior.
For now, `nbsphinx` is using `pandoc` (which itself supports multiple Markdown dialects), but in the future I'd like to use a Python library (which probably doesn't exist yet) for this.

> Other than that I probably haven't used half the customization that is possible. One thing I thought about is whether to put the logo in the topbar instead of the sidebar.

There is no specific setting for putting the logo there, but you can try to use the [left_buttons](https://insipid-sphinx-theme.readthedocs.io/en/0.2.3/configuration.html#theme-option-left_buttons) option for that.

But I think in general it's a waste of space to put the logo into the topbar.

You should think about users with small screens (like smartphones). Putting the logo there will make the topbar actually less useful for them.

You could of course add some CSS to avoid showing the logo on small screens, but if the logo is so important to you, you probably also want it to be shown on those small screens, right?

What about putting the logo at the top of each page (using the `page.html` template)?

> That likely is already possible but I took your advice and focus on the content. I really like how nbsphinx and your insipid theme made that possible!

That's great to hear!

mgeier:
The PDF output seems to have the same problem, there is no space between cells: https://648-46379698-gh.circle-artifacts.com/0/nbsphinx.pdf#section.6

t-vi:
> The PDF output seems to have the same problem, there is no space between cells: https://648-46379698-gh.circle-artifacts.com/0/nbsphinx.pdf#section.6

Yes, but so this happens when I use the nbconvert facility, so it would appear to be a bug there.

mgeier:
> > The PDF output seems to have the same problem, there is no space between cells: https://648-46379698-gh.circle-artifacts.com/0/nbsphinx.pdf#section.6
> 
> Yes, but so this happens when I use the nbconvert facility, so it would appear to be a bug there.

It very likely is a bug there, but `nbsphinx` doesn't use the LaTeX/PDF workflow that `nbconvert` uses (it uses the reST template which is later converted to LaTeX by Sphinx).

Either way, I would like it to work correctly in `nbsphinx`, but if it gets fixed in `nbconvert` as well, that's of course also good.

t-vi:
> Either way, I would like it to work correctly in nbsphinx, but if it gets fixed in nbconvert as well, that's of course also good.

I'm afraid my notebook foo isn't enough to implement these tags after all.

PR #552, opened 2021-04-07
Increase docutils line_length_limit

### LINKED ISSUE #549: failing to parse long lines with docutils 0.17
My notebooks contains outputs with embed javascripts that's super long. Before `nbsphinx` was happy to convert them. But after I update my build environment today I get the following warnings:

```
xxx.ipynb:: WARNING: Line xxx exceeds the line-length-limit.
```

which is immediately followed by

```
xxx.ipynb:1: WARNING: Each notebook should have at least one section title
```

and eventually this lead to all my notebook not included in the resulting html build since titles are missing.

I can confirm this can be fixed by `pip install docutils==0.16`, but thought it's worth raising it here in case anyone else came across the same issue.

mgeier:
Closes  #549.

I have no idea whether this is the right place to change this setting, but it seems to work ...

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -1876,6 +1876,7 @@ def load_requirejs(app):
 
 def builder_inited(app):
     env = app.env
+    env.settings['line_length_limit'] = 100_000_000
     env.nbsphinx_notebooks = {}
     env.nbsphinx_files = {}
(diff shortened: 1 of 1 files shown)


mgeier:
See also https://github.com/jupyter/nbconvert/issues/1544.

Issue #563, opened 2021-05-12
Jinja2 update to version 3 breaks nbsphinx


nikosavola:
As Jinja2 version is not frozen to 2.11 in [setup.py](https://github.com/spatialaudio/nbsphinx/blob/master/setup.py#L18) nbsphinx uses the new version 3.0.0.

With Jinja2 3.0.0 I've been getting the following error for all notebooks
```
'nbformat.notebooknode.NotebookNode object' has no attribute 'tags'
```
This does not happen if I freeze Jinja2 to 2.11.

kif:
I confirm: `pip install  Jinja2==2.11` fixes the bug for me too 
https://readthedocs.org/projects/pyfai/builds/13742631/

mgeier:
Thanks for reporting this!

It looks like the Jinja2 behavior changed, but I'm not entirely sure whether that was on purpose.

I'm not sure, but it's probably related to https://github.com/pallets/jinja/pull/1204

Jinja2 release notes: https://jinja.palletsprojects.com/en/3.0.x/changes/#version-3-0-0.

mgeier:
Thanks for the fix in #564!
I've merged it and I've just made a new release https://pypi.org/project/nbsphinx/0.8.5/.

If anybody has an idea how to fix the root cause of the problem, please let me know!

sammorley-short:
Just wanted to add that I am only seeing this bug for pip version 20.2 (well, specifically 20.2.4). Using `pip>=20.3.0` also solves the problem.

I'm not sure exactly what changed to cause this, but pip 20.3 made a load of changes to the resolver, so presumably it's one of the ones listed here: https://pip.pypa.io/en/stable/news/#v20-3

astrojuanlu:
I fail to see how the pip version affects this 🤔

sammorley-short:
Yeah, apologies, I wasn't particularly clear above. 

In my case the fix in 0.8.5 released yesterday still doesn't fix the problem if you are using pip < 20.3, since the release pegging was not obeyed. However, upgrading to pip >= 20.3 fixed the issue for me, presumably because of a change in the dependency resolution ordering between the two minor versions.

Not saying this because it's something that needs changing on the `nbsphinx` side, but just flagging this for anyone else who wasn't seeing 0.8.5 solve the problem. 🙂

marscher:
I'm still observing the reported behavior with nbsphinx=0.8.5

mgeier:
[...]

PR #564, opened 2021-05-12
Freeze Jinja2 version to 2.11

NikoDaGreat:
I mentioned in https://github.com/spatialaudio/nbsphinx/issues/563 that the new Jinja2 version breaks nbsphinx.

This is a simple workaround to freeze Jinja2 version to 2.11.
Refactoring nbsphinx to support Jinja2 3.x is probably wise at some point but this should work for now


mgeier:
Thanks!

PR #577, opened 2021-06-17
Improve MathJax compatibility with myst-parser

chrisjsewell:
Heya, this would meet me halfway with https://github.com/executablebooks/MyST-Parser/pull/395, so both packages will share the same `processClass` regex (and myst-parser won't complain about overriding it).
Plus anyway it is slightly better for compatibility with any advanced use cases, so people can still use the MathJax standard classes `tex2jax_process` (mathjax2) and `mathjax_process` (mathjax3).

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -1889,7 +1889,7 @@ def config_inited(app, config):
         mathjax3_config['tex'] = tex
         options = {
             'ignoreHtmlClass': 'document',
-            'processHtmlClass': 'math|output_area',
+            'processHtmlClass': 'tex2jax_process|mathjax_process|math|output_area',
         }
         options.update(mathjax3_config.get('options', {}))
         mathjax3_config['options'] = options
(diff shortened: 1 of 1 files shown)


chrisjsewell:
> line too long (83 > 79 characters)

get with the times, use black formatting and 88 character lines 😜

mgeier:
Thanks for this PR!

I have nothing against this change, but just for my understanding, who/what generates those classes `tex2jax_process` and `mathjax_process`?

> line too long (83 > 79 characters)

That's just the opinion of the bot, feel free to ignore it.

> use black formatting

I won't. I don't like it.

> and 88 character lines

Only in exceptional cases.

chrisjsewell:
cheers

> but just for my understanding, who/what generates those classes `tex2jax_process` and `mathjax_process`

well nothing in sphinx/docutils core, but these are the default classes that MathJax itself sets to mark elements whose contents should be processed (in v2 and v3 respectively) and so, one would imagine, that this should really be the canonical way that any third-party extensions and/or advanced use cases should (or already do) mark elements for processing.

mgeier:
Thanks, I wasn't aware that those options have default values. See http://docs.mathjax.org/en/v3.1-latest/options/document.html?highlight=mathjax_process#option-descriptions.

With the argument of better compatibility, shouldn't the value `ignoreHtmlClass: 'mathjax_ignore'` (and `'tex2jax_ignore'` for MathJax 2) be kept as well?

mgeier:
I've added the missing "ignore" defaults in #585.

PR #602, opened 2021-10-08
doc fix: --InlineBackend.rc=figure.dpi=96 => dict

deargle:
The docs suggest adding `--InlineBackend.rc=figure.dpi=96` to `nbsphinx_execute_arguments` in `conf.py` -- but this will lead to nbsphinx complaining about .rc not being a dict. This changes the suggestion to be `--InlineBackend.rc={'figure.dpi': 96}`


mgeier:
Do you happen to have a traitlets version < 5?

I've made the opposite change in #553.

mgeier:
I've created #627, which adds a `traitlets>=5` requirement and clarifies the usage of `ipython_kernel_config.py`.

PR #604, opened 2021-10-11
Add nbsphinx_assume_equations to documentation

### LINKED ISSUE #603: Documenting nbsphinx_assume_equations
Would it be possible to add `nbsphinx_assume_equations` to the documentation? I found the (useful!) option recently a bit by chance - when trying to look it up for more details using the `0.8.7` documentation, I could [only] find a reference in the changelog. 

The changelog entry was still enough, though (and looking at the related PR and an issue mentioning the option clarified its usage even more) - it's mostly for making it slightly easier to find.

[only]: https://nbsphinx.readthedocs.io/en/0.8.7/search.html?q=nbsphinx_assume_equations

diego-plan9:
Closes #603 

Add an entry to the `usage.ipynb` with information about `nbsphinx_assume_equations`. The description is basically copy-pasted from the changelog, trusting the original writer was more fluent in the flag than me :)


doc/usage.ipynb
```diff
@@ -248,6 +248,15 @@
     "See [Ignoring Errors](allow-errors.ipynb)."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### `nbsphinx_assume_equations`\n",
+    "\n",
+    "If `True`, do not force loading MathJax on HTML pages generated from notebooks."
```
mgeier, line 257: If I'm not mistaken, this is the opposite:

```suggestion
    "If `False`, do not force loading MathJax on HTML pages generated from notebooks."
```

mgeier (commented):
Sorry for the late response ...

doc/usage.ipynb
diego-plan9: Oops, sounds right!

diego-plan9:
Thanks for the follow-up - I also merged from `master` which seems to have fixed the previously failing `build-me-my-docs` stage.

mgeier:
Thanks!

PR #606, opened 2021-10-27
fix: avoid closing the script tag early by escaping a forward slash

maartenbreddels:
Same as https://github.com/jupyter/nbconvert/pull/1665

Requires a new release of nbconvert as well

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -284,7 +284,7 @@ RST_TEMPLATE = """
 .. raw:: html
 
     <script type="application/vnd.jupyter.widget-state+json">
-    {{ nb.metadata.widgets['application/vnd.jupyter.widget-state+json'] | json_dumps }}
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR and sorry for my late response!

I played around with this a bit and it turns out that the `</script>` tag is case-insensitive and that browsers (at least Firefox and Chromium) also have a problem with `</script ` (with a space at the end).

I've created an extended fix in #611 which should cover more cases. Or am I missing something?

maartenbreddels:
Yeah, even better, care to upstream this?

(from mobile phone)

Op wo 24 nov. 2021 22:25 schreef Matthias Geier ***@***.***>:

mgeier:
> Yeah, even better, care to upstream this?

I don't really want to undo your change from https://github.com/jupyter/nbconvert/pull/1665, which isn't wrong and works in most *real* cases. But if you prefer my solution, you can of course take it and apply it (or parts of it) to your original change.

PR #615, opened 2021-12-07
support text builder

jdknight:
Provides initial support to allow processing nbsphinx nodes using a Sphinx `text` translator.


mgeier:
Thanks! Looking forward to further improvements!

PR #628, opened 2022-01-10
Support .jp-RenderedHTMLCommon CSS class

flying-sheep:
If one e.g. uses `ipywidgets` to wrap a DataFrame, the output area has no `rendered_html` class, but `jp-RenderedHTMLCommon`

### Alternative

Using `:is(.rendered_html, .jp-RenderedHTMLCommon) table` would be a less verbose alternative that works in all current browser versions: https://caniuse.com/css-matches-pseudo

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -656,6 +656,7 @@ div.nboutput.container div.output_area > div[class^='highlight']{
 }
 
 /* Some additional styling taken form the Jupyter notebook CSS */
+.jp-RenderedHTMLCommon table,
 div.rendered_html table {
   border: none;
   border-collapse: collapse;
@@ -664,10 +665,14 @@ div.rendered_html table {
   font-size: 12px;
   table-layout: fixed;
 }
+.jp-RenderedHTMLCommon thead,
 div.rendered_html thead {
   border-bottom: 1px solid black;
   vertical-align: bottom;
 }
+.jp-RenderedHTMLCommon tr,
+.jp-RenderedHTMLCommon th,
+.jp-RenderedHTMLCommon td,
 div.rendered_html tr,
 div.rendered_html th,
 div.rendered_html td {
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR!

Do you have a small example for me to reproduce the problem?

Maybe we should include this in the example notebook in the docs?
There are already examples for DataFrame and for widgets, but not both at once.

Regarding `:is(...)`: I don't have a preference.

flying-sheep:
An example would be this:

```py
import pandas as pd
import ipywidgets as widgets

df = pd.DataFrame(...)

out = widgets.Output()
with out:
    display(df)

tabs = widgets.Tab(children=[out])
tabs.set_title(0, 'Test')
tabs
```

mgeier:
Thanks for the example, that's great!

I'll add it to the docs: #633

flying-sheep:
Perfect! Do you need anything more from me or is this ready then?

mgeier:
Thanks, that's all I need for now.

I'd like to have a look into the equations problem to see if there is any overlap, and then I'll come back to this.

mgeier:
I still don't understand why equation rendering doesn't work within widgets.
It should be enabled everywhere within `output_area`, but for some reason it doesn't seem to work.

Do you have an idea why that's the case?

flying-sheep:
No idea! Is that issue blocking this one? I’d assume this PR would be useful on its own.

mgeier:
> No idea! Is that issue blocking this one?

No it isn't!

> I’d assume this PR would be useful on its own.

Yes it is!

> I could take a look anyway if you give me a pointer to where equations are handled

Equations are handled via MathJax, which is supposed to be activated on elements with the `output_area` CSS class:

https://github.com/spatialaudio/nbsphinx/blob/be3e63ebbb2e2beb934661508ea056fe33f747ca/src/nbsphinx.py#L1910

AFAICT, the widget outputs have this class, but somehow it doesn't seem to work ...

Issue #495, opened 2020-09-27
CI: Move to Github Actions


mgeier:
This has already been asked for in #358 and #361.

I think it would good - to keep it simple - to have several configuration files:

- [ ] run HTML builder with many Python and Sphinx versions
  * this is currently done on Travis-CI (config: https://github.com/spatialaudio/nbsphinx/blob/d5d352c5440cc7a185751bdf0d1ba263b7a2026d/.travis.yml)
  * I don't think we need to do this on multiple operating systems? Ubuntu should suffice?

- [ ] run HTML builder on Windows, using default versions

- [ ] run HTML builder on macOS, using default versions

- [ ] run HTML builder with default versions (whatever OS), upload HTML artifacts
  * Github still doesn't allow non-zipped artifacts, see #361.
  * I don't think we need multiple versions of the HTML artifacts, one OS (using default versions) should suffice

- [ ] create PDF with TexLive, upload artifact
  * one OS should be sufficient?

- [ ] create PDF with `tectonic`, upload artifact
  * one OS should be sufficient?

- [ ] run linkcheck (only once), see #491

Any further suggestions?

cc @s-weigand

s-weigand:
I got pretty much all of this running, besides building of PDF on windows ([see this workflow, still WIP with an ugly commit history](https://github.com/s-weigand/nbsphinx/blob/gh-action-ci-switch/.github/workflows/test.yml)).
For tests on windows to pass I had to use python<3.8 (#496 ) and add the `raises-exception` tag to the [bash-magic cell](https://nbsphinx.readthedocs.io/en/0.7.1/code-cells.html#Cell-Magics).

## Latex and Windows

The problem with creating PDFs on Windows is due to the difference in the created `nbsphinx.tex` file.

`textonic` and `pdflatex` both fail with:
```bash
LaTeX Warning: Reference `a-normal-rst-file:example_python_function' on page 23
 undefined on input line 2056.

[23]
! FancyVerb Error:
  Empty verbatim environment
.
\FV@Error ...ncyVerb Error:^^J\space \space #1^^J}

l.2292 \end{sphinxVerbatim}
```

When looking at the diff of the linux and windows generated `nbsphinx.tex` 
``` bash
$ diff -Z --color latex-linux/nbsphinx.tex  latex-win/nbsphinx.tex
```

you find the following:

```diff
2292,2294c2292,2339
< 1
< 2
< 3
---
> \end{sphinxVerbatim}
> }
> 
> {
[...]

PR #630, opened 2022-01-10
Test with range of Python versions

1kastner:
With newer versions, Python changes its behavior. This might affect how nbsphinx operates.


mgeier:
See #495.

See also https://github.com/spatialaudio/nbsphinx/blob/master/.travis.yml

mgeier:
I don't think it makes sense to do `linkcheck` on multiple Python versions, right?

Other workflows are still missing, for details see #495.

1kastner:
Yes, it might be reasonable to add those before.

Issue #625, opened 2022-01-05
Add to documentation of how to add raw cells with ReST in JupyterLab


1kastner:
Hello there and thanks a lot for this great project! In #624, I quickly discussed how I used and plan to use ReST cells in my Jupyter Notebooks to create Sphinx-compatible links with the help of Sphinx roles. In the UI Jupyter Notebook, raw cells can be easily selected as it can be seen in the screenshot at https://nbsphinx.readthedocs.io/en/0.8.8/raw-cells.html. The support of Jupyter Notebook is fading (see [this announcement](https://github.com/jupyter/notebook#notice)) and more and more people shift to JupyterLab. Other UIs exist as well. Maybe it could be possible to extend the documentation at this point to show how to do the same (setting the Raw Cell type) in other UIs, such as JupyterLab? Or just provide a link to other sources where people can read it?

With my current JupyterLab version, it appears in the right sidebar.
![grafik](https://user-images.githubusercontent.com/5236165/148265230-87a640e3-6746-4509-bcbe-d98768f4e2a0.png)

If it is accepted, I could take a screenshot similar to https://github.com/spatialaudio/nbsphinx/blob/master/doc/images/raw_cells.png (does the depicted notebook reside in this repository?) and it could be added to the raw cell description - discussing both the Jupyter Notebook UI and JupyterLab. Would that be acceptable for this project?

mgeier:
I'm painfully aware the the screenshot is becoming outdated.
However, I'm not willing to continuously update this screenshot (or even multiple of them).

If you want to update it (and/or add a different one for JupyterLab), I'd be happy to receive a PR from you.

Alternatively, the only thing I would be willing to change myself is to completely remove the screenshot because of the maintenance burden.

> does the depicted notebook reside in this repository?

Well, yes, it does (https://github.com/spatialaudio/nbsphinx/blob/master/doc/raw-cells.ipynb), but the screenshot depicts an old version.

That's yet another point to proof that screenshots will become outdated and have to be updated regularly ...

1kastner:
I can do an update this time for sure. I will also have a look whether there is a link (with intersphinx?) that could cross-reference to another documentation that could do the job here and the screenshots might not be necessary. I am not sure which one will create less work in the long run though.

1kastner:
[...]

PR #632, opened 2022-01-11
Update explanation regarding Raw cells

1kastner:
* Update screenshot for Jupyter Notebook
* Add screenshot for JupyterLab
* Extend explanation for Raw cells


1kastner:
Fixes #625

1kastner:
The result of the runner is not reproducible for me on my local machine even though I copied the command from the action output at https://github.com/spatialaudio/nbsphinx/runs/4773390871?check_suite_focus=true

1kastner:
> Duplicate substitution definition name: "Steps for converting cells to Raw formats in Jupyter".

I need to further research how that can be solved best.

mgeier:
The error might not be reproducible due to different Pandoc versions.

The solution should be to simply change one of the `Steps for converting cells to Raw formats in Jupyter`, which shouldn't be the same anyway.

1kastner:
> The error might not be reproducible due to different Pandoc versions.
> 
> The solution should be to simply change one of the `Steps for converting cells to Raw formats in Jupyter`, which shouldn't be the same anyway.

That sounds reasonable. I added "Notebook" or "Lab" depending on the UI.

doc/raw-cells.ipynb
```diff
@@ -15,17 +15,44 @@
    "source": [
     "# Raw Cells\n",
     "\n",
-    "The \"Raw NBConvert\" cell type can be used to render different code formats into HTML or LaTeX by Sphinx.  This information is stored in the notebook metadata and converted appropriately.\n",
+    "Any Jupyter Notebook consists of cells of three different types:\n",
+    "Code cells, Markdown cells, and/or a Raw cells.\n",
```
mgeier, line 19:
```suggestion
    "Code cells, Markdown cells, and/or Raw cells.\n",
```

doc/raw-cells.ipynb
```diff
@@ -15,17 +15,44 @@
...
+    "While most Jupyter Notebook users are very familiar with Code cells and Markdown cells in Jupyter Notebooks, Raw cells are less frequently used.\n",
+    "For Jupyter Notebook, they are introduced\n",
+    "[here](https://jupyter-notebook.readthedocs.io/en/stable/notebook.html?highlight=raw#raw-cells)\n",
+    "and for JupyterLab\n",
+    "[here](https://jupyterlab.readthedocs.io/en/stable/extension/notebook.html?highlight=raw#model).\n",
+    "The Raw cells are also sometimes referred to as Raw NBConvert cells in the context of\n",
+    "[nbconvert](https://nbconvert.readthedocs.io/en/latest/architecture.html?highlight=raw%20cell#a-detailed-pipeline-exploration).\n",
```
mgeier, line 26: I'm not sure if the highlighting is helpful at all, but highlighting "cell" is distracting, isn't it?

```suggestion
    "[nbconvert](https://nbconvert.readthedocs.io/en/latest/architecture.html?highlight=raw#a-detailed-pipeline-exploration).\n",
```

doc/raw-cells.ipynb
```diff
@@ -15,17 +15,44 @@
...
+    "To select a desired format from within JupyterLab, first activate the right sidebar by clicking on View in the Menu Toolbar.\n",
+    "Then you ensure that in front of Show Right Sidebar there is a tick.\n",
+    "Once the Right Sidebar is shown, you are ready to go.\n",
+    "\n",
+    "Now you select the cell containing your special code and choose options from the following dropdown menus:\n",
+    "\n",
+    "1. Select \"Raw\" in the Notebook Toolbar (just next to the symbols that run cells or reload the kernel)\n",
+    "2. Choose the appropriate \"Raw NBConvert Format\" in the Right Sidebar\n",
+    "\n",
+    "![Steps for converting cells to Raw formats in JupyterLab](images/raw_cells_jupyterlab.png)\n",
```
mgeier, line 55: The image is too wide when rendered with Sphinx: https://nbsphinx--632.org.readthedocs.build/en/632/raw-cells.html#JupyterLab

You should make it smaller or - if that's too much work - use `` to make it fit (see https://nbsphinx.readthedocs.io/en/0.8.8/markdown-cells.html#Using-the-HTML-%3Cimg%3E-tag).

Or, since this is about raw cells, you can use a raw reST cell?

1kastner: Again, maybe due to the different pandoc version it looks different when I build the docs locally. There, the width is always 100%

1kastner: I used the image tag because the image was cropped in ways that were unpredictable for me on my local machine.

1kastner:
I hope @mgeier I could fix all the minor issues you spotted.

mgeier:
Yes, thanks!

However, I'm not sure if scaling the images to 100% width is the right thing to do.

This might actually be a Sphinx theme problem ... many themes automatically scale images to 100% width, but that doesn't mean it's the best thing to do.

The problem is that on small screens (e.g. smartphone) the screenshot might become unreadable.
Granted, users can zoom in, but I think that's kinda complicated.

Also, the scaled version looks more ugly than the unscaled one.

What do you think about making the images scrollable?

I've implemented this in the `insipid` theme: https://github.com/mgeier/insipid-sphinx-theme/pull/66

1kastner:
Making the image scrollable sounds acceptable to me. You previously mentioned that the JupyterLab image was clipped and when I had a closer look, I saw that the Jupyter Notebook image was slightly clipped in the prior version, too. And having clipped images in the documentation is something I would prefer to avoid.

mgeier:
I had a little setback: https://github.com/mgeier/insipid-sphinx-theme/pull/66 doesn't seem to work as intended.

I had to come up with an alternative (https://github.com/mgeier/insipid-sphinx-theme/pull/68), which isn't perfect either, but it is much closer to other Sphinx themes.

What do you think about that?

> I would rather like to see this pull request merged earlier and move your totally valid concerns of not readable screenshots to a next working package. Would that be an acceptable procedure for you?

I would prefer to fix the theme issue first.
Because depending on that, this PR might have to be changed.

I think the `max-width: 100%` approach is the way to go for now, even though it has its disadvantages.

If we decide to do this, it doesn't make sense to set `100%` in this PR.
In this case, the old Markdown links can be restored because we don't need the special `` handling.

> especially as there are plenty of themes that are supported by nbsphinx and not only insipid.

That's a good point.
With https://github.com/mgeier/insipid-sphinx-theme/pull/68, the behavior of `insipid` should get closer to the other themes.

The rendering in different themes can be checked reasonably quickly with the https://github.com/spatialaudio/nbsphinx/blob/master/theme_comparison.py script.

1kastner:
The old markdown links have been restored. I would proceed this way:
1. I will go through all themes as they are generated by https://github.com/spatialaudio/nbsphinx/blob/master/theme_comparison.py - not more and not less.
2. I will have a look how they treat the screenshots. If the screenshots are clipped in my browser, then I will check whether `max-width: 100%` is set in the theme or not. 
3. In case it is currently missing, I would add `max-width: 100%` to the image tag (most likely defined in some CSS file) to avoid clipped graphics for now.

First of all, is that the workflow you intended?

I followed the steps of https://github.com/spatialaudio/nbsphinx/blob/efd08d8f777dcf7f435625b4a5726e240b5e2937/CONTRIBUTING.rst#building-themes and I encountered an issue that I opened at #634.

doc/raw-cells.ipynb
```diff
@@ -15,17 +15,44 @@
...
     "\n",
-    "1. Select \"Raw NBConvert\"\n",
-    "2. Switch the Cell Toolbar to \"Raw Cell Format\"\n",
-    "3. Chose the appropriate \"Raw NBConvert Format\" within the cell\n",
+    "### Jupyter Notebook\n",
     "\n",
-    "![Steps for converting cells to Raw formats in Jupyter](images/raw_cells.png)\n",
+    "To select a desired format from within Jupyter Notebook, select the cell containing your special code and choose options from the following dropdown menus:\n",
+    "\n",
+    "1. Select \"Raw NBConvert\" in the Menu Toolbar (just below the two menus \"Widgets\" and \"Help\")\n",
```
mgeier, line 38: This toolbar doesn't seem to be visible in the screenshot?

doc/raw-cells.ipynb
```diff
@@ -15,17 +15,44 @@
...
+    "\n",
+    "### JupyterLab\n",
+    "\n",
+    "To select a desired format from within JupyterLab, first activate the right sidebar by clicking on View in the Menu Toolbar.\n",
+    "Then you ensure that in front of Show Right Sidebar there is a tick.\n",
+    "Once the Right Sidebar is shown, you are ready to go.\n",
+    "\n",
+    "Now you select the cell containing your special code and choose options from the following dropdown menus:\n",
+    "\n",
+    "1. Select \"Raw\" in the Notebook Toolbar (just next to the symbols that run cells or reload the kernel)\n",
```
mgeier, line 52: This doesn't seem to be visible either?

mgeier:
Thanks for testing all the themes, very interesting!

> ... and instead the image is clipped

What do you mean by "clipped"?
Do you mean that the page has to be horizontally scrolled to be able to see the right edge of the image?

I have just merged https://github.com/mgeier/insipid-sphinx-theme/pull/68, which means that the `insipid` theme should now (once a release has been made) behave like most other themes with regards to shrinking large images.

> As the later themes are maintained by separate developers, the max-width must be passed through in some other way, e.g. by having additional CSS definitions that are added to the generation process.

I think it is the intention of the theme authors not to scale the images, so I would simply leave it as it is.

However, if you want to improve the overall quality of this PR, you could create the screenshots again with a bit smaller resolution.
We can't make it perfect for all themes, but we could choose a width that's displayed in most themes without shrinking.
And BTW, if you decide to do the screenshots again, you could try to close JupyterLab's left sidebar and make the right sidebar a bit narrower, which would leave more room for the middle column.

doc/raw-cells.ipynb
```diff
@@ -15,17 +15,44 @@
    "source": [
     "# Raw Cells\n",
     "\n",
-    "The \"Raw NBConvert\" cell type can be used to render different code formats into HTML or LaTeX by Sphinx.  This information is stored in the notebook metadata and converted appropriately.\n",
+    "Any Jupyter Notebook consists of cells of three different types:\n",
+    "Code cells, Markdown cells, and/or Raw cells.\n",
```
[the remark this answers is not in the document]
1kastner: It will be done with the next update

doc/raw-cells.ipynb
1kastner: My bad, it was toggled. It will be done with the next update.

doc/raw-cells.ipynb
1kastner: Here, the issue is that I tried to make the screenshot narrow enough so that the text is actually readable. The menu just disappeared at some point as the toolbar does not like to span two rows I guess.

1kastner:
> What do you mean by "clipped"? Do you mean that the page has to be horizontally scrolled to be able to see the right edge of the image?

I use a quite new version of Firefox and there is no scrollbar. Instead, the right part of the image is just not visable. Here is the screenshot of the screenshot as it is displayed to me in my browser (looking at <nbsphinx-project-root>/build/raw-cells.html):

grafik

> However, if you want to improve the overall quality of this PR, you could create the screenshots again with a bit smaller resolution.

I might try to resize the next screenshot slightly. I hope the quality does not suffer too much.

1kastner:
I added the resized images and used them in the Jupyter Notebook. The resized version was created with the following code:

```
<nbspinx-project-dir>/doc/images$ convert -resize 65% raw_cells_jupyter_notebook.png raw_cells_jupyter_notebook_resized.png
<nbspinx-project-dir>/doc/images$ convert -resize 65% raw_cells_jupyterlab.png raw_cells_jupyterlab_resized.png
```

The original images now open when you click on the resized images. I believe that would come quite handy for people who are visually impaired and who would have troubles using the resized images. Moreover, I came up with the 65% percent just by guess-work and in the future, as the insipid theme might evolve, so might its default pixel width change?

I hope with these last updates on the pull request I could fulfill all requested changes to your full satisfaction.

mgeier:
Thanks for the update!

> The original images now open when you click on the resized images. I believe that would come quite handy for people who are visually impaired and who would have troubles using the resized images.

I don't know much about the accessibility of screenshots, but I guess it's good to have the pixel-exact screenshot because the re-scaled image might be a bit blurry.

> Moreover, I came up with the 65% percent just by guess-work

But how did you come up with the width of the original screenshot?

Why did you not make that 65% narrower?

> as the insipid theme might evolve, so might its default pixel width change?

Sure, the default width might change.
I don't expect it to get narrower, but there are no guarantees.

> I hope with these last updates on the pull request I could fulfill all requested changes to your full satisfaction.

Well, I'm fine with them, and if you want I can merge them as is.

But I wouldn't call it "full satisfaction" ...

I think the two versions of each screenshot are overkill.

Why not make the original screenshot narrower, so it doesn't have to be scaled in the first place?

This was the situation of the previous screenshot and it looked very crisp.

1kastner:
> Why not make the original screenshot narrower, so it doesn't have to be scaled in the first place?

I just tried to re-take the screenshots in the narrower way. For the Jupyter Notebook UI, this is less of an issue. When I try to make the crisp screenshot of JupyterLab, the UI elements start to arrange in a way it usually does not when people use JupyterLab. 

This is what I would call the "normal" arrangement of the user interface but the screenshot is too wide:
raw_cells_jupyterlab

For me, this narrow arrangement of the user interface is already distorted even though it is ~800px and we are not even close to the 700px to crack (that is the width of the 65%-resized images). I had to manually enable the second toolbar so that the cell type dropdown menu was even visible.
raw_cells_jupyterlab

By default, the left menu has a fixed width, so I can only make the displayed Jupyter Notebook narrower. So here I cracked to border of 700 px. Only now we can use the screenshot in the theme without resizing. In this case, however, the content of the screenshot between Jupyter Notebook and JupyterLab deviate. I guess the JupyterLab UI was never designed for such narrow images.

raw_cells_jupyterlab

For me, this is a rather unusual way of how the JupyterLab UI is arranged and I am afraid that users might be confused. Furthermore, the two screenshots now display different content which I also previously tried to avoid. But I guess for every point there are pro and con arguments and I can definitively understand yours as well. If you feel that this is preferable, I will re-arrange the pull request.

mgeier:
That's strange. When I tried it, the buttons were not messed up:

image

Can you make the fonts smaller (by pressing Ctrl-minus or something)?

I find it strange that the right sidebar has a maximum width that's IMHO unnecessarily large. That should probably be reported, but I don't have enough energy to do that. I already have a litany of JupyterLab issues I would like to report some day ...

> I guess the JupyterLab UI was never designed for such narrow images.

I guess not.
But people are talking about it, e.g. https://github.com/jupyterlab/jupyterlab/issues/3275#issuecomment-618470517

> If you feel that this is preferable, I will re-arrange the pull request.

I'll leave it up to you.

1kastner:
In the screenshot you show, the dropdown to select "Raw cell" is missing - something you pinpointed in one of my previous screenshots.

1kastner:
I updated the font size as you suggested, removed the resized image, and made a small fix. With my local pandoc installation everything is fine again and so it does to me in the auto-triggered readthedocs. I guess there is no homework left and things are set now.

mgeier:
Thanks for your patience!

I think it is much better now.

> In the screenshot you show, the dropdown to select "Raw cell" is missing - something you pinpointed in one of my previous screenshots.

Yeah, sorry, I missed that!

PR #633, opened 2022-01-11
DOC: add example with rich content within Output widget

mgeier:
This illustrates the problem described in #628.

Rendered: https://nbsphinx--633.org.readthedocs.build/en/633/code-cells.html#Interactive-Widgets-(HTML-only)

Math is broken as well, BTW.

### DIFF
--- a/doc/code-cells.ipynb
+++ b/doc/code-cells.ipynb
@@ -557,6 +557,55 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


mgeier:
I've added some Markdown content with math and a table, which isn't working correctly, either.

Rendered: https://nbsphinx--633.org.readthedocs.build/en/633/code-cells.html#Interactive-Widgets-(HTML-only)

PR #637, opened 2022-03-09
Prevent adding a source suffix twice

martinRenou:
PR coupled with https://github.com/jupyterlite/jupyterlite-sphinx/pull/28

This prevents collisions between nbsphinx and jupyterlite-sphinx which both add the `".ipynb"` source prefix: if `app.registry.source_suffix` already contains "ipynb", then do nothing.

The behavior will then be the following:

`conf.py`
```python
# nbsphinx will handle .ipynb files and jupyterlite-sphinx won't
extensions = [
    'nbsphinx',
    'jupyterlite_sphinx',
]

# jupyterlite-sphinx will handle .ipynb files and nbsphinx won't
extensions = [
    'jupyterlite_sphinx',
    'nbsphinx',
]
```

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -28,6 +28,7 @@ __version__ = '0.8.9'
 import collections.abc
 import copy
 import html
+from itertools import chain
 import json
 import os
 import re
@@ -1929,13 +1930,14 @@ def patched_toctree_resolve(self, docname, builder, toctree, *args, **kwargs):
 
 
 def config_inited(app, config):
-    suffixes = [".ipynb", *config.nbsphinx_custom_formats]
-    for suffix in suffixes:
-        if (
-            suffix not in config.source_suffix
-            and suffix not in app.registry.source_suffix
-        ):
-            app.add_source_suffix(suffix, 'jupyter_notebook')
+    for suffix in config.nbsphinx_custom_formats:
+        app.add_source_suffix(suffix, 'jupyter_notebook')
+    if '.ipynb' in chain(config.source_suffix, app.registry.source_suffix):
+        # We don't interfere if '.ipynb' has been added by the user in conf.py
+        # or by another Sphinx extension that has been loaded earlier.
+        pass
+    else:
+        app.add_source_suffix('.ipynb', 'jupyter_notebook')
 
     if '**.ipynb_checkpoints' not in config.exclude_patterns:
         config.exclude_patterns.append('**.ipynb_checkpoints')
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -1929,10 +1929,13 @@ def patched_toctree_resolve(self, docname, builder, toctree, *args, **kwargs):
 
 
 def config_inited(app, config):
-    if '.ipynb' not in config.source_suffix:
-        app.add_source_suffix('.ipynb', 'jupyter_notebook')
-    for suffix in config.nbsphinx_custom_formats:
-        app.add_source_suffix(suffix, 'jupyter_notebook')
+    suffixes = [".ipynb", *config.nbsphinx_custom_formats]
+    for suffix in suffixes:
+        if (
+            suffix not in config.source_suffix
+            and suffix not in app.registry.source_suffix
+        ):
+            app.add_source_suffix(suffix, 'jupyter_notebook')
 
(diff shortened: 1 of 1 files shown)


martinRenou:
I can really see documentation taking advantages of both extensions.

src/nbsphinx.py
```diff
@@ -1886,10 +1886,13 @@ def patched_toctree_resolve(self, docname, builder, toctree, *args, **kwargs):
...
 def config_inited(app, config):
-    if '.ipynb' not in config.source_suffix:
-        app.add_source_suffix('.ipynb', 'jupyter_notebook')
-    for suffix in config.nbsphinx_custom_formats:
-        app.add_source_suffix(suffix, 'jupyter_notebook')
+    suffixes = [".ipynb", *config.nbsphinx_custom_formats]
+    for suffix in suffixes:
+        if (
+            suffix not in config.source_suffix
+            and suffix not in app.registry.source_suffix
```
mgeier, line 1893: I'm not familiar with `app.registry.source_suffix`. Is this where the stuff added by `app.add_source_suffix()` ends up?

src/nbsphinx.py
```diff
@@ -1886,10 +1886,13 @@ def patched_toctree_resolve(self, docname, builder, toctree, *args, **kwargs):
 
 
 def config_inited(app, config):
-    if '.ipynb' not in config.source_suffix:
-        app.add_source_suffix('.ipynb', 'jupyter_notebook')
-    for suffix in config.nbsphinx_custom_formats:
-        app.add_source_suffix(suffix, 'jupyter_notebook')
+    suffixes = [".ipynb", *config.nbsphinx_custom_formats]
```
mgeier, line 1889: I don't think that the `nbsphinx_custom_formats` have to be checked further. They are provided by the user, so it's the user's responsibility not to provide conflicting suffixes.

mgeier (commented):
Thanks for this PR!

src/nbsphinx.py
martinRenou: Yes, the app [calls `add_source_suffix` on the registry](https://github.com/sphinx-doc/sphinx/blob/f38bd8e9529d50e5cceffe3ca55be4b758529ff7/sphinx/application.py#L1170) then [the registry will throw if the suffix is already registered](https://github.com/sphinx-doc/sphinx/blob/f38bd8e9529d50e5cceffe3ca55be4b758529ff7/sphinx/registry.py#L266)

src/nbsphinx.py
martinRenou: I see. Does this `nbsphinx_custom_formats` config do anything else that sphinx's `source_suffix` don't do? https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix

> so it's the user's responsibility not to provide conflicting suffixes

Actually my understanding is that sphinx's `source_suffix` is a way for users to overwrite anything that extensions have setup, so it's not really the same thing? Or am I completely misunderstanding all this? (sphinx beginner here)

mgeier: > Does this `nbsphinx_custom_formats` config do anything else that sphinx's `source_suffix` don't do?

Yes. In addition to registering the source suffix, it also provides a function that transforms a string containing certain custom format into a Jupyter notebook object. See https://nbsphinx.readthedocs.io/custom-formats.html.

But I assume that if a user provides a custom format, they don't want their chosen suffix to be overwritten by some other Sphinx extension.

And if another extension is loaded after `nbsphinx`, I prefer an error over silently ignoring the source suffix.

And if a user provides `nbsphinx_custom_formats` *and* a conflicting `source_suffix`, it's their own fault.

mgeier: @martinRenou Do you want to update this PR or should I do it?

martinRenou: Please do it if you can :) Thanks a lot!

mgeier: I have rebased your branch and I've added fc79fd1ac5ab6df225b256ef4ca203c58c5ad92b that unconditionally adds `nbsphinx_custom_formats`.

Can you please test if that works for you?

martinRenou: Thanks a lot!

mgeier: I've merged this PR already, but please let me know if it doesn't work as expected, so we can fix that.

Issue #341, opened 2019-10-18
Using nbconvert.preprocessors.RegexRemovePreprocessor


ratnanil:
I read @mgeier's [answer](https://github.com/spatialaudio/nbsphinx/issues/285#issuecomment-464856891) to a question (#285)  regarding the skipping of some cells using a regex. I'm totally at a loss here and and don't know whether I'm look in the right place: But is there a way to use nbconvert.preprocessors.RegexRemovePreprocessor with `sphinx `/`nbsphinx ` on individual *.ipynb files to skip cells based on a regex? I would be very glad for a pointer!

mgeier:
There is currently no way to specify which preprocessors are to be used.

The `HighlightMagicsPreprocessor` is hard-coded to be enabled, the `RegexRemovePreprocessor` is hard-coded to be disabled (but latter is not necessary anymore, because the original problem has been fixed in https://github.com/jupyter/nbconvert/pull/784).

There is currently no way to specify additional preprocessors, but such a feature has been suggested in  #305. Feel free to add comments there and feel free to make a PR.

mgeier:
I forgot: The `ExecutePreprocessor` is used directly in the `nbsphinx` source code, without using `nbconvert`'s "config" mechanism.

Issue #305, opened 2019-05-28
Hook into preprocessing prior execution


marscher:
nbconvert is very flexible, when it comes to preprocessing notebooks. One can define the preprocessors execution order and so on and so forth. Some issues like hiding certain stuff (#303, #65 etc.) could be easily implemented via a custom preprocessor.
But it does not seem to be possible to add them prior the ExecutePreprocessor does its job.
What do you think? Would it be worth adding a mechanism like this to make it more flexible for users?

mgeier:
Yes, I think it would make sense to add some support for specifying custom preprocessors.

If you have a concrete idea how to do this, feel free to open a new PR or add some comments here.

Please note that I have hard-coded a few preprocessor settings here:

https://github.com/spatialaudio/nbsphinx/blob/880bf3a30dff75e16a2b1644e8293e13524a5af0/src/nbsphinx.py#L722-L726

If custom preprocessors are supported, this might have to be changed, too (or not?).

marscher:
I think the best option would be to implement this in the fashion of the nbconvert Exporter class:

https://github.com/jupyter/nbconvert/blob/a863787231f6e420d290acd98ef1e7af79b691e7/nbconvert/exporters/exporter.py#L48-L52

marscher:
My current workaround looks like this. In order to generalize this, we would need to add a preprocessors argument to the application and pass it into the right spot.

``` python
def _preprocess_notebooks():
    """ hooks into ExecutePreprocessor.preprocess to execute our own filters."""
    import nbsphinx
    from nbconvert.preprocessors import Preprocessor
    org_method = nbsphinx.Exporter.from_notebook_node

    class MyPreprocessor(Preprocessor):
        def preprocess(self, nb, resources):
            # ... 
            return nb, resources

    class MyPreprossor2(Preprocessor):
        def preprocess(self, nb, resources):
            # ...
            return nb, resources

    def my_from_notebook_node(self, nb, resources, **kwargs):
        self.log.info('patched preprocessing method')
        filters = [MyPreprocessor(),
                   MyPreprossor2(),
                   ]
        for f in filters:
            nb, resources = f.preprocess(nb, resources=resources)

        return org_method(self, nb, resources=resources, **kwargs)

    nbsphinx.Exporter.from_notebook_node = my_from_notebook_node
```

mgeier:
> we would need to add a preprocessors argument

Could you please provide a few examples for how such arguments might look like?

[...]

Issue #285, opened 2019-02-14
How to skip some cells?


wookayin:
This is really a great plugin for sphinx!

I am curious how we can skip some cells from ipython notebooks that we mark using some comments, directives, or [cell metadata](https://medium.com/@mjspeck/presenting-code-using-jupyter-notebook-slides-a8a3c3b59d67) to be excluded in converting. This would be super useful in writing documentations when we want to exclude some preambles or dirty data preparation codes from the rendered sphinx docs.

I don't think ipython nbconvert does have this feature :)

mgeier:
Thanks!

Is https://nbsphinx.readthedocs.io/en/0.4.2/hidden-cells.html what you need?

> when we want to exclude some preambles or dirty data preparation codes

Well it would probably be better to clean up the preparation code, wouldn't it?

And did you consider moving the (clean) preparation code to a separate `.py` file and then using `import` to work with it in your notebook?

> I don't think ipython nbconvert does have this feature :)

In fact, there are multiple ways to do this in `nbconvert`:

By regex with https://nbconvert.readthedocs.io/en/latest/api/preprocessors.html#nbconvert.preprocessors.RegexRemovePreprocessor

By tags with `TagRemovePreprocessor.remove_cell_tags` and `TagRemovePreprocessor.remove_input_tags` (see https://nbconvert.readthedocs.io/en/latest/config_options.html)

wookayin:
This works so great, sorry for missing that!

ratnanil:
I'd be very interested to use nbconvert.preprocessors.RegexRemovePreprocessor. Can this funcionality be used within nbsphinx? I cant find anything in the docs..

ratnanil:
It seems that `nbconvert.preprocessors.RegexRemovePreprocessor` [is disabled](https://github.com/spatialaudio/nbsphinx/blob/880bf3a30dff75e16a2b1644e8293e13524a5af0/src/nbsphinx.py#L722-L726) (see also https://github.com/spatialaudio/nbsphinx/issues/305#issuecomment-505484450). Does this mean that I cannot use `RegexRemovePreprocessor`?

mgeier:
@ratnanil See my answer at https://github.com/spatialaudio/nbsphinx/issues/341#issuecomment-544119378.

PR #650, opened 2022-06-22
Use nbconvert's preprocessor framework

jasongrout:
As noted in https://github.com/spatialaudio/nbsphinx/issues/341#issuecomment-544119639, nbsphinx directly uses the nbconvert execute preprocessor before the preprocessor queue is invoked. This changes the logic to use the nbconvert preprocessor queue. It also enables the user to give a configuration for preprocessors (for example, I used this to remove cells with certain tags).

The first commit in this PR appears to work just fine. However, with the second commit applied, building the ipywidgets docs with nbsphinx ends with some errors about files (notebooks) not found in the html-collect-pages step. Unfortunately, I ran out of time to debug this, but am opening a draft PR with my current in-progress work in hopes it will be of use.

This begins to address comments from https://github.com/spatialaudio/nbsphinx/issues/285, https://github.com/spatialaudio/nbsphinx/issues/305, https://github.com/spatialaudio/nbsphinx/issues/341.


mgeier:
Thanks for this PR and sorry for my late response!

It would be great if we could provide a hook for users to provide additional preprocessor settings.

Sadly, I don't have time to look into this either, but maybe someone else does?

I was wondering if the `RegexRemovePreprocessor` should be forced, but I guess this isn't necessary anymore since https://github.com/jupyter/nbconvert/pull/784 has been merged.

However, I think we should force `HighlightMagicsPreprocessor`, because otherwise people might inadvertently disable the highlighting of "magic" cells.

Issue #651, opened 2022-06-27
Notebook gallery with thumbnail links only


adonath:
We use the `.. nbgallery::` directive in the documentation of the Gammapy project to include tutorial notebooks in the normal documentation (https://docs.gammapy.org/0.20.1/tutorials/index.html). This feature works great and adds a lot of value and improved accessibility to the documentation. Thanks for this!

The way `.. nbgallery::` is implemented it works as a `toctree`. I think it would be a nice feature to be able to have the `nbgallery` work as links with thumbnails only. The use case we have in our documentation is to be able to link to tutorials using thumbnails from e.g. the user guide:

- https://docs.gammapy.org/0.20.1/user-guide/dl3.html#using-gammapy-data
- https://docs.gammapy.org/0.20.1/user-guide/maps/index.html#using-gammapy-maps

We already do this, however it leads to a broken navbar structure (https://github.com/gammapy/gammapy/issues/3985), because rendered tutorials are part of multiple `toctree` structures. This makes the content of the left sidebar dependent on the order the pages are being rendered. One way to resolved this issue for us would be to have a `nbgallery` directive, which does not act as a `toctree` but just as plain links. Is any feature like this planned? How could it be implemented most easily? 

Issue #647 might be related, but has different motivation.

mgeier:
Thanks for bringing this up!

> Is any feature like this planned?

It's not planned, but I think it would be a very good idea!

> How could it be implemented most easily?

I don't know. I think a separate directive (something like `nblinkgallery`maybe?) would have to be implemented.
The entries would have to be looked up in the known source files to check if they exist.
I'm not sure how globbing would work, but maybe this doesn't have to be implemented at first.
Then, given the list of source files, the appropriate `docutils` nodes would have to be created.

I currently don't have time to look into this, but if you need help, please let me know!

adonath:
Thanks @mgeier for the response! I'm actually interested to work in this, however I have minimal knowledge on implementing sphinx extensions.

The first part on parsing and checking the source files should be fine. But I'm not sure what the correct `docutils` nodes are to represent the structure. Looking at an example html from the Gammapy documentation:

```html
<div class="sphx-glr-thumbcontainer" tooltip="Data inspection and preliminary analysis with Fermi-LAT data.">
  <div class="figure align-center">
    <img alt="thumbnail" src="../_images/tutorials_data_fermi_lat_54_0.png" />
    <p class="caption">
      <span class="caption-text">
[...]

PR #654, opened 2022-07-22
Implement nblinkgallery directive

adonath:
This PR addresses issue #651 and includes the following changes:
- Implement new `NbLinkGallery` class and `nblinkgallery` directive
- Implement `GalleryLinks` node
- Adapt `doctree_resolve` to re-use the `GalleryNode` for `GalleryLinks` as well
- Add minimal docs example to `a-normal-rst-file.rst`

I consider this still as a draft. I'm happy to receive any comments for improvement, especial reduction of code duplication and improvement of documentation.

### DIFF
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -302,7 +302,7 @@ Thumbnail Links Galleries
 -------------------------
 
 In some case it is desired to just create thumbnail links to existing notebooks,
-already inlcuded in a ``toctree``. This can be used e.g. to link to a subset
+already included in a ``toctree``. This can be used e.g. to link to a subset
 of notebooks from API documentation to highlight the use of some functionality.
 
 For this there is a dedicated ``nblinkgallery`` directive.
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -299,18 +299,12 @@ The following example gallery was created using:
 
 
 Thumbnail Links Galleries (HTML only)
--------------------------
+-------------------------------------
 
 In some case it is desired to just create thumbnail links to existing notebooks,
 already included in a ``toctree``. This can be used e.g. to link to a subset
 of notebooks from API documentation to highlight the use of some functionality.
-
 For this there is a dedicated ``nblinkgallery`` directive.
-
-.. note::
-
-    The notes regarding LaTeX in :ref:`/subdir/gallery.ipynb`
-    and :ref:`/subdir/toctree.ipynb` also apply here!
 
 The following example gallery was created using:
 
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -298,7 +298,7 @@ The following example gallery was created using:
     gallery/*-rst
 
 
-Thumbnail Links Galleries
+Thumbnail Links Galleries (HTML only)
 -------------------------
 
 In some case it is desired to just create thumbnail links to existing notebooks,
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -296,3 +296,31 @@ The following example gallery was created using:
     :reversed:
 
     gallery/*-rst
+
+
+Thumbnail Links Galleries
+-------------------------
+
+In some case it is desired to just create thumbnail links to existing notebooks,
+already inlcuded in a ``toctree``. This can be used e.g. to link to a subset
+of notebooks from API documentation to highlight the use of some functionality.
+
+For this there is a dedicated ``nblinkgallery`` directive.
+
+.. note::
+
+    The notes regarding LaTeX in :ref:`/subdir/gallery.ipynb`
+    and :ref:`/subdir/toctree.ipynb` also apply here!
+
+The following example gallery was created using:
+
+.. code-block:: rest
+
+    .. nblinkgallery::
+        gallery/multiple-outputs.ipynb
+        gallery/no-thumbnail.ipynb
+
+.. nblinkgallery::
+    gallery/multiple-outputs.ipynb
+    gallery/no-thumbnail.ipynb
+
--- a/doc/a-normal-rst-file.rst
+++ b/doc/a-normal-rst-file.rst
@@ -311,10 +311,10 @@ The following example gallery was created using:
 .. code-block:: rest
 
     .. nblinkgallery::
-        gallery/multiple-outputs.ipynb
-        gallery/no-thumbnail.ipynb
+        gallery/multiple-outputs
+        gallery/no-thumbnail
 
 .. nblinkgallery::
-    gallery/multiple-outputs.ipynb
-    gallery/no-thumbnail.ipynb
+    gallery/multiple-outputs
+    gallery/no-thumbnail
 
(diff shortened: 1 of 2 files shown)


mgeier:
Thanks, this looks promising!

The HTML result is broken because of an unrelated problem with the bibliography. I'll try to find and report the problem, but in the meantime, you can just remove the `bibliography` directive during development and it should build fine.

The linkcheck error is also unrelated.

src/nbsphinx.py
```diff
@@ -1262,6 +1267,29 @@ def run(self):
...
+    has_content = True
+
+    def run(self):
+        """Check and collect notebook links"""
+        path = Path(self.env.srcdir) / self.env.docname
+
+        links = []
+
+        for entry in self.content:
+            filename = (path.parent / entry).resolve()
```
mgeier, line 1281: To be consistent with [toctree](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree), the entries should be handled as "document names", i.e. they don't require suffixes (but they are supported for backwards compatibility). Also, if an entry starts with a slash, it should be relative to the source directory. I'm not sure if there is already some Sphinx code that can be re-used for that.

I was wondering if `env.found_docs` can be used here, but maybe this is not populated yet at this point?

doc/a-normal-rst-file.rst
```diff
@@ -296,3 +296,31 @@ The following example gallery was created using:
     :reversed:
 
     gallery/*-rst
+
+
+Thumbnail Links Galleries
+-------------------------
+
+In some case it is desired to just create thumbnail links to existing notebooks,
+already inlcuded in a ``toctree``. This can be used e.g. to link to a subset
```
mgeier, line 305:
```suggestion
already included in a ``toctree``. This can be used e.g. to link to a subset
```

doc/a-normal-rst-file.rst
```diff
@@ -296,3 +296,31 @@ The following example gallery was created using:
     :reversed:
 
     gallery/*-rst
+
+
+Thumbnail Links Galleries
```
mgeier, line 301:
```suggestion
Thumbnail Links Galleries (HTML only)
```

doc/a-normal-rst-file.rst
```diff
@@ -296,3 +296,31 @@ The following example gallery was created using:
...
+In some case it is desired to just create thumbnail links to existing notebooks,
+already inlcuded in a ``toctree``. This can be used e.g. to link to a subset
+of notebooks from API documentation to highlight the use of some functionality.
+
+For this there is a dedicated ``nblinkgallery`` directive.
+
+.. note::
+
+    The notes regarding LaTeX in :ref:`/subdir/gallery.ipynb`
+    and :ref:`/subdir/toctree.ipynb` also apply here!
```
mgeier, line 313: I don't think those notes apply, because the whole thing doesn't work with LaTeX, right?

src/nbsphinx.py
```diff
@@ -2088,63 +2162,57 @@ def doctree_resolved(app, doctree, fromdocname):
             if doc in toctree['includefiles']:
                 if title is None:
                     title = app.env.titles[doc].astext()
+
                 uri = app.builder.get_relative_uri(fromdocname, doc)
-                base = sphinx.util.osutil.relative_uri(
-                    app.builder.get_target_uri(fromdocname), '')
 
                 # NB: This is how Sphinx implements the "html_sidebars"
                 #     config value in StandaloneHTMLBuilder.add_sidebars()
```
mgeier, line 2169: This comment should stay with the following code.

src/nbsphinx.py
mgeier: I didn't test this, but maybe you can use the existing `parse_content()` method?

https://github.com/sphinx-doc/sphinx/blob/f45dc5e1398f7850e23003cac200ad3433da2a78/sphinx/directives/other.py#L79

I guess you would have to override the `run()` method, because the options are not exactly equal.

https://github.com/sphinx-doc/sphinx/blob/f45dc5e1398f7850e23003cac200ad3433da2a78/sphinx/directives/other.py#L55

adonath: I was able to re-use the `TocTree.parse_content()` method, however just by passing the `NbLinkGallery` as `self` to  the `.parse_content()` method, which is somewhat unusual and maybe not very future proof. However it works and simplifies the code, while getting the possibility to work with the document names only. I adapted the documentation accordingly.

adonath:
Thanks @mgeier  for there review! I addressed all of your comments, please let me know if you have further ones. Currently it seems the latex build fails, I'm not sure what to do about this. Can I test the latex build locally?

adonath:
@mgeier Do you have time to look at this again? Thanks!

mgeier:
> Can I test the latex build locally?

Yes, sure: https://nbsphinx.readthedocs.io/en/0.8.9/contributing.html#building-the-documentation
I then normally change to `build/sphinx/latex/` and run `latexmk -pdf -pv`

> @mgeier Do you have time to look at this again? Thanks!

I currently have little time and next week I'm on holidays. But after that, I'll have a closer look.

src/nbsphinx.py
mgeier: Why can't you derive `NbLinkGallery` from `sphinx.directives.other.TocTree`?

I guess you would also have to override the `option_spec` (and maybe other things).

mgeier:
I've reported the bibliography-related error there: https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/309

src/nbsphinx.py
mgeier: To answer my own question:

> Why can't you derive `NbLinkGallery` from `sphinx.directives.other.TocTree`?

When doing that, links to the gallery notebooks are included in the TOC, which is not what we want.
I didn't find an elegant/clean solution, but I tried replacing the `toctree` node with a dummy node, which took care of the problem.

My experiments are available at 4bcd8dcc95ad80308cf23d625544cd73e47b1484, which is a bit hacky but I think it has quite good code re-use.

It also enables a few missing directive options and disallows the ones that don't work.

What do you think about it?

mgeier: I just created a little variation, which is hopefully somewhat less hacky: 9615c4e76cbe80f049a463e1dbe20c99125325cd

adonath: Thanks @mgeier! I agree your proposed solution is cleaner and even better re-use of code. How do we proceed maybe attach your commits to this PR? Or open another PR against mine?

mgeier: It would be great if you could continue with this PR.

If you like my commit, feel free to take it into your PR, if you like only some parts, just take some parts. And maybe you find some more things to improve.
You could also try to rebase or squash the commits, which would hopefully make the CI pass.

Once we agree on the functionality, I'd like to have a closer look at the documentation.

mgeier:
@adonath I took the liberty of using your work from here and building upon it in #696, I hope you don't mind!

src/nbsphinx.py
mgeier: I have continued working on this in #696.

adonath:
Thanks @mgeier! Yes that’s totally fine. For our project I ended up using a different solution so I didn’t give this PR any priority…

PR #657, opened 2022-07-30
Fix #656: Sphinx 5.1.0 has modified some internal LaTeX macros

jfbu:
Fix #656

I hesitated between leaving comments inside the LATEX_PREAMBLE or rather move them out as Python comments.  In the end I dif leave a latex comment about `\nbsphinxincludegraphics` but it should probably be removed.

This commit ensures compatibility with Sphinx 5.1.0 which has changed some internal macros related to rendering of code-boxes.  The legacy code is left exactly identical when it is detected Sphinx is `< 5.1.0`.  The new code opens up possibility to use some of the 5.1.0 added extensions regarding separate customizability of border widths and paddings.

I tested it by compiling the nbsphinx.pdf with Sphinx at least at https://github.com/sphinx-doc/sphinx/commit/cdde69918812b079a92325345478ac3d1afaf7c2 which adds to Sphinx 5.1.0 a fix related to code-blocks but unrelated to current nbsphinx (which does not use `verbatimwithframe=false`).  The pdf output looks ok.

About this:

>  I will provide a PR ... to refactor nbsphinxfancyoutput in a way not using internal Sphinx macros and keeping exact same behaviour otherwise.

It proved elusive for this task to not use any internal Sphinx LaTeX macro (i.e. with `@` in its name), if we want to benefit from the styling and properties of Sphinx Verbatim in a custom way (in particular without actually using sphinxVerbatim environement itself for this type of cell).

---

Additional notes not related to this PR, for archival.

<sub>
side notes:

There were some numerous slightly annoying LaTeX iterated warnings, urelated to this PR:
```
Package textcomp Warning: Symbol \textasciigrave not provided by
(textcomp)                font family zlmtt in TS1 encoding.
(textcomp)                Default family used instead on input line 7252.

```
They all seem to follow some earlier warning:
```
[10] [11] [12] [13] [14] [15]

Package textcomp Warning: Symbol \textasciigrave not provided by
(textcomp)                font family pplj in TS1 encoding.
(textcomp)                Default family used instead on input line 1267.

[16] [17] [18]
...[snip]...
LaTeX Font Warning: Font shape `T1/cmtt/l/n' undefined
(Font)              using `T1/cmtt/m/n' instead on input line 3591.

(/usr/local/texlive/2022/texmf-dist/tex/latex/base/ts1cmtt.fd)

LaTeX Font Warning: Font shape `TS1/cmtt/l/n' undefined
(Font)              using `TS1/cmtt/m/n' instead
[...]

PR #658, opened 2022-08-05
Fix missing backtick for literal mark-up in a markdown cell

jfbu:
relates https://github.com/spatialaudio/nbsphinx/pull/657#issuecomment-1206279663


mgeier:
Thanks!

PR #686, opened 2022-12-04
timings & timestamps for input cells

deeenes:
- This patch makes it possible to include under each cell the time of the last run and how long the execution took
- It does the same as the `execute_time` extension for Jupyter Notebook: https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/execute_time/readme.html
- The timings are taken from `cell.metadata.ExecuteTime`, I don't know if it exists in notebooks that don't use the extension
- The `nbsphinx_timings` option enables or disables showing the timings (not sure where to document it)

The HTML output looks like this (haven't tested LaTeX):

nbsphinx_timings

Having the timings is highly helpful for the readers as they know what to expect, how long does it take to run the code, and tells something about the age of the content (e.g. it's too new or too old to run with a certain version, etc).

### DIFF
--- a/src/nbsphinx.py
+++ b/src/nbsphinx.py
@@ -32,6 +32,7 @@ from itertools import chain
 import json
 import os
 import re
+import datetime
 import subprocess
 import sys
 from urllib.parse import unquote
@@ -122,6 +123,10 @@ RST_TEMPLATE = """
 {{ resources.codecell_lexer }}
 {%- endif -%}
 {{ insert_empty_lines(cell.source) }}
+{%- if cell.metadata.ExecuteTime %}
+    :execution-start: {{ cell.metadata.ExecuteTime.start_time }}
+    :execution-end: {{ cell.metadata.ExecuteTime.end_time }}
+{%- endif %}
 {%- if cell.execution_count %}
     :execution-count: {{ cell.execution_count }}
 {%- endif %}
@@ -532,7 +537,7 @@ div.nboutput.container {
     display: -webkit-flex;
     display: flex;
     align-items: flex-start;
-    margin: 0;
+    margin: 7px 0 0 0;
     width: 100%%;
 }
 @media (max-width: %(nbsphinx_responsive_width)s) {
@@ -620,7 +625,7 @@ div.nboutput.container div.output_area {
 div.nbinput.container div.input_area {
     border: 1px solid #e0e0e0;
     border-radius: 2px;
-    /*background: #f5f5f5;*/
+    background: #f5f5f5;
 }
 
 /* override MathJax center alignment in output cells */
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR!

How is this related to https://github.com/deshaw/jupyterlab-execute-time?

Will this work with notebook v7?

> The timings are taken from `cell.metadata.ExecuteTime`, I don't know if it exists in notebooks that don't use the extension

I guess not, I've never seen it.

> The `nbsphinx_timings` option enables or disables showing the timings (not sure where to document it)

I think this should be documented in a new notebook, maybe in this section: https://nbsphinx.readthedocs.io/executing-notebooks.html

It would be good if only that one notebook shows the timings, the rest of the docs should stay unchanged.

The actual option should also be mentioned at https://nbsphinx.readthedocs.io/en/0.8.10/usage.html#nbsphinx-Configuration-Values

> (haven't tested LaTeX)

We will see once this is part of the docs.
The PDF should be generated automatically by CI, but currently this seems to be broken (due to an unrelated problem, which I'll look into once I have time).

> [...] and tells something about the age of the content (e.g. it's too new or too old to run with a certain version, etc).

I think there are better ways to convey this information.
For example, you could include the versions of the used modules in the notebook.
If you want to know the date of the last change in a notebook (and if you happen to use Git), I can recommend my own extension https://github.com/mgeier/sphinx-last-updated-by-git.

Anyway, that's not really that important for the issue at hand.

deeenes:
Thanks Matthias for the reply,

> Thanks for this PR!
> 
> How is this related to https://github.com/deshaw/jupyterlab-execute-time?
> 
> Will this work with notebook v7?

This works with the classic Jupyter Notebook extension, but can be adapted easily to the extension above, only have to check how it records the timestamps in the `ipynb` data structure.

> 
> > The timings are taken from `cell.metadata.ExecuteTime`, I don't know if it exists in notebooks that don't use the extension
> 
> I guess not, I've never seen it.
> 
> > The `nbsphinx_timings` option enables or disables showing the timings (not sure where to document it)
> 
> I think this should be documented in a new notebook, maybe in this section: https://nbsphinx.readthedocs.io/executing-notebooks.html
> 
> It would be good if only that one notebook shows the timings, the rest of the docs should stay unchanged.
> 
> The actual option should also be mentioned at https://nbsphinx.readthedocs.io/en/0.8.10/usage.html#nbsphinx-Configuration-Values
> 
> > (haven't tested LaTeX)
> 
> We will see once this is part of the docs. The PDF should be generated automatically by CI, but currently this seems to be broken (due to an unrelated problem, which I'll look into once I have time).

Cool, maybe it's easier for you to edit the docs, or I can give it a try if you want

> 
> > [...] and tells something about the age of the content (e.g. it's too new or too old to run with a certain version, etc).
> 
> I think there are better ways to convey this information. For example, you could include the versions of the used modules in the notebook. If you want to know the date of the last change in a notebook (and if you happen to use Git), I can recommend my own extension https://github.com/mgeier/sphinx-last-updated-by-git.

Sure, a plugin for this (and support in nbsphinx) would be great, just like the `sessionInfo` in R.

> 
> Anyway, that's not really that important for the issue at hand.

mgeier:
> only have to check how it records the timestamps in the `ipynb` data structure.

Yes, please do that. I hope the two extensions are compatible.

> maybe it's easier for you to edit the docs, or I can give it a try if you want

I will try to fix the CI error (which is probably related to #687), but you should modify the docs for your new feature. The docs are at the same time tests to check if the feature works, so they should be updated together with the implementation of the feature.

mgeier:
I've merged #688, now CI should work.

PR #719, opened 2023-03-12
DOC: add example for including nbsphinx docs in another LaTeX file

mgeier:
This is an example for how to include Sphinx-generated content in a hand-written LaTeX file.

Rendered instructions: https://github.com/spatialaudio/nbsphinx/blob/7515477cfb0d029443cdae477b9aeef9f11a566c/include-in-latex/README.md

The resulting PDF file is here: https://output.circle-artifacts.com/output/job/538c4602-dd06-4181-adca-86293c169914/artifacts/0/my-latex-document.pdf


mgeier:
I have asked for feedback on the mailing list (https://groups.google.com/g/sphinx-users/c/yys9wg2OnkQ), but there hasn't been any response yet.

mgeier:
Latest version of PDF file, now with syntax highlighting: https://output.circle-artifacts.com/output/job/b50b8c06-562b-49e3-8e1b-64e38fcd8da0/artifacts/0/my-latex-document.pdf

mgeier:
Latest PDF: https://output.circle-artifacts.com/output/job/5cad65eb-3d43-4f75-a0a5-4fbab6c0d2c6/artifacts/0/my-latex-document.pdf

include-in-latex/my-latex-document.tex
```diff
@@ -0,0 +1,121 @@
+\RequirePackage{luatex85}
+\documentclass[a4paper]{book}
+
+% NB: This contains the "pxunit" config value (".75bp" by default):
+\ifdefined\pdfpxdimen
+   \let\sphinxpxdimen\pdfpxdimen\else\newdimen\sphinxpxdimen
+\fi \sphinxpxdimen=.75bp\relax
+\ifdefined\pdfimageresolution
+    \pdfimageresolution= \numexpr \dimexpr1in\relax/\sphinxpxdimen\relax
+\fi
```
mgeier, line 10: The lines above look quite complicated, is there some way to avoid having to include them?

include-in-latex/my-latex-document.tex
```diff
@@ -0,0 +1,121 @@
...
+
+\catcode`^^^^00a0\active\protected\def^^^^00a0{\leavevmode\nobreak\ }
```
mgeier, line 12: I'm not sure what that does (something with non-breaking spaces?), is this actually needed?

mgeier:
@jfbu Could you please have a quick look at the LaTeX stuff if you have some time?

AFAICT, this works quite well as it is, but maybe you have some ideas to make it simpler/better/more consistent/...?

BTW, this is the latest PDF: https://output.circle-artifacts.com/output/job/a68beaf5-d661-487c-abf9-fc7a9183117e/artifacts/0/my-latex-document.pdf

include-in-latex/my-latex-document.tex
jfbu: there are two distinct things here
- first one is to support user input of an image with using `px` unit for dimensions; this gets converted by Sphinx LaTeX into usage of `\sphinxpxdimen`.  The default of `.75bp` means `96px=1in`. Some engines (pdftex and luatex) have a `\pdfpxdimen` (actually for luatex only thanks to package luatex85, as luatex name is `\pxdimen`).  It looks natural to use that predefined register but one could simply drop the `\if..\fi` and do always `\newdimen\sphinxpxdimen` then `\sphinxpxdimen=.75bp`.  Making sure `\sphinxpxdimen` is same as `\pdfpxdimen` is important if your source will have direct usage of `px` as a dimensional unit anywhere. In the Sphinx context, this is authorized only in rest sources when specifying dimensions and then Sphinx replaces `px` by `\sphinxpxdimen` mark-up, because `px` is supported only by pdfTeX and luaTeX not other engines, else it could leave `px` as is in output, but one would need to set what it default is.  Notice that there is what I consider a bug in the pdfTeX default value of `px`, which differs by `1/65536` of a TeX point from the (correct) luaTeX default.  But last sentence was only for bragging because that default is supposed to make approximately `72px=1in` and at Sphinx we chose `96px=1in`. Maybe I said too much.
- second one is how the engines handle image inclusion like jpeg when the image file is lacking x- or y-resolution; it is a last resort value.  It works in pdftex, and perhaps in luatex (not sure). I *think* it is not available in xetex (whose documentation is incomplete).  However looking now at this code in the Sphinx LaTeX template I see that it is silly because `\sphinxpxdimen` has been set above in line 8.  So the whole `\numexpr\dimexpr` thing is silly, and `\pdfimageresolution = 96\relax` can replace it.  Sorry about that I will simplify upstream.  Thanks for ping.

include-in-latex/my-latex-document.tex
jfbu: This converts the Unicode character for a non-breaking space into suitable LaTeX mark-up.  It was needed at some point for xeLaTeX/luaLaTeX and the `^^^^` notation works ONLY with these Unicode-aware engines.  In the past it was definitely needed, but the LaTeX team has increased support for UTF-8 input in recent years, so I need to check if it is still needed with modern LaTeX.

jfbu: Hmm in distant past (TeXLive 2015) a basic LuaLaTeX document with no font configuration would report (silently) in the log a `Missing character: There is no   (U+00A0) in font cmr10!`, but this went away starting with TeXLive 2016 and besides I am not sure it ever applied with a document using fontspec and an OpenType font.  LuaLaTeX simply puts the U+00A0 literally in the PDF according to `\showoutput` log, but the PDF viewer seems to convert it to normal space.

I tested with DejaVu Sans and lualatex, TL2014 and TL2023 appear to act the same.  A space is there in PDF, but when copied pasted the original Unicode char is not recovered only a standard ascii space.  This is true for both text and math mode.

The difference when one checks the TeX logs with `\showoutput` is the explicit `\penalty 10000` inserted by the `\nobreak`.  There is no such penalty when the raw U+00A0 is left as is.  But in practice TeX will never break a line at character it needs glue or explicit zero or negative `\penalty`.  I have no remembrance whatsoever why the above was added but it must have been motivated by some context.  

Ah I do find a difference but not with lualatex, with xelatex.
```latex
\documentclass{article}
\usepackage{fontspec}
\setmainfont{DejaVu Sans}

% is this really needed?
% \catcode`^^^^00a0\active\protected\def^^^^00a0{\leavevmode\nobreak\ }
\begin{document}
\noindent\parbox{0pt}{%
\hskip0pt hyphenation hyphenation
}%
\end{document}
```

with LuaLaTeX both "hyphenation" are hyphenated.  With XeLaTeX only the first one, and I had not inserted the `\hskip0pt` also the first one would not be hyphenated (except with LuaLaTeX, TeX does not hyphenate first word in a paragraph; Sphinx LaTeX prefixes all text paragraphs, if not using LuaLaTeX, with a `\hskip0pt` to work around that). 

If one uses the `^^^^` code then both engines act the same.

At this stage, for documents using OpenType Fonts which do have an `U+00A0` (i think some Adobe public fonts do not, I don't recall details), there does not seem to be much reason to do the `^^^^` line, apart from assuring XeLaTeX and LuaLaTeX behave the same.

Sorry for being too long.

jfbu: ah ok I recall now.  The NON-BREAKING SPACE unicode char will not participate to the available stretch (and shrink) in the paragraph lines.  This is shown by this file that you can process with lualatex:
```latex
\documentclass{article}
\usepackage{fontspec}
\setmainfont{DejaVu Sans}

%%%is this needed
%%%\catcode`^^^^00a0\active\protected\def^^^^00a0{\leavevmode\nobreak\ }
\begin{document}
\hrule

using control-space (with \verb+\nobreak+ but changes nothing here)

xxxxxx $xxx\nobreak\ yyy$ $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 

xxxxx $xxx\nobreak\ yyy$ $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 
 $xxx\nobreak\ yyy$ 


using NON-BREAKING SPACE unicode char:

xxxxxx $xxx yyy$ $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 

xxxxx $xxx yyy$ $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx yyy$ 
 $xxx  yyy$ 
\end{document}
```

(sorry that in the above the non-breaking space are GONE from my copy paste you need to replace all `xxx yyy` by `xxxU+00A0yyy`)

Output with lualatex:
Capture d’écran 2023-05-21 à 17 18 54

One sees that Unicode NON-BREAKING SPACE does not participate to the overall stretchability of the lines.

Sorry for being too long...

include-in-latex/my-latex-document.tex
jfbu: ah about my remark about the `\pdfimageresolution` complicated `\numexpr\dimexpr` being unnecessary, the original template is
```
\sphinxpxdimen=<%= pxunit %>\relax
\ifdefined\pdfimageresolution
    \pdfimageresolution= \numexpr \dimexpr1in\relax/\sphinxpxdimen\relax
\fi
```
so in original context it is definitely needed to support `'pxunit'` [latex_elements](https://www.sphinx-doc.org/en/master/latex.html#the-latex-elements-configuration-setting) key.  If you impose `\sphinxpxdimen=.75bp\relax` then as explained you should then simply use
```
\ifdefined\pdfimageresolution \pdfimageresolution= 96\relax\fi
```
With LuaLaTeX, the `luatex85` package must be loaded for this syntax.

mgeier: Sorry for my late response! And thanks for the detailed explanation!

I have simplified it as you suggested, but I'm still wondering if it's possible to avoid these cryptic incantations in the main hand-written LaTeX file.

Would it be possible to move this from the main file into one of the separate Sphinx style files?

Issue #745, opened 2023-05-22
Download buttons/links


wpbonelli:
Hi, does `nbsphinx` offer a similar feature to `sphinx-gallery` where [notebooks are zipped and can be downloaded](https://sphinx-gallery.github.io/stable/configuration.html#disabling-download-button-of-all-scripts) from the rendered page? One option I see is linking in prolog/epilog to GitHub but I would like to keep notebooks as jupytext paired scripts and avoid version controlling ipynbs, while allowing the docs visitor to download either a `.py` or an `.ipynb`.

mgeier:
You can just change `.html` to `.ipynb`, for example:

* https://nbsphinx.readthedocs.io/en/0.9.1/custom-formats.html
* https://nbsphinx.readthedocs.io/en/0.9.1/custom-formats.ipynb

This will give you the *executed* notebook (if you didn't disable execution).

Does that help?

The problem is that, depending on the webserver configuration, the file might be displayed as JSON in the browser. This is the reason why I didn't provide a download link in the docs but instead created a "View in nbviewer" link: https://nbviewer.org/urls/nbsphinx.readthedocs.io/en/0.9.1/custom-formats.ipynb

https://github.com/spatialaudio/nbsphinx/blob/ce39da6798d9f654a3942b255cbe4d6a266d02d1/doc/conf.py#L55-L61

`.zip` files are currently not available.

I have thought about customizing the provided downloads, but up to now nobody has asked for it!

w-bonelli:
Thanks! Got it working with [`:download:` role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#referencing-downloadable-files) in the prolog, e.g. `prolog.rst`

```rst
.. only:: html

    .. role:: raw-html(raw)
        :format: html

    .. note::

        | Download :download:`Jupyter notebook (.ipynb) <{{ docname }}>` or :download:`Python script (.py) <{{ docname[:-6] }}.py>`
```

and in `conf.py`

```python
nbsphinx_prolog = r"""
{% set docname = env.doc2path(env.docname, base=None) %}
""" + Path("prolog.rst").read_text()
```

mgeier:
I'm glad you found a solution that works for you!

If your project is public, would you mind providing a link, just out of curiosity?

Does the `:download:` link work better for you than directly linking to the notebook file?
I guess it's not really a problem, but now you have two copies of all `.ipynb` files on your server, right?

[...]

PR #751, opened 2023-06-29
DOC: Add download link for executed .ipynb file

mgeier:
See discussion in #745.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -50,6 +50,7 @@ nbsphinx_prolog = r"""
       <a class="reference external" href="https://github.com/spatialaudio/nbsphinx/blob/{{ env.config.release|e }}/{{ docname|e }}">{{ docname|e }}</a>.
       Interactive online version:
       <span style="white-space: nowrap;"><a href="https://mybinder.org/v2/gh/spatialaudio/nbsphinx/{{ env.config.release|e }}?filepath={{ docname|e }}"><img alt="Binder badge" src="https://mybinder.org/badge_logo.svg" style="vertical-align:text-bottom"></a>.</span>
+      <a href="{{ env.docname.split('/')|last|e + '.ipynb' }}" class="reference download internal" download>Download notebook</a>.
       <script>
         if (document.location.host) {
           let nbviewer_link = document.createElement('a');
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Looks good to me. Continuing the discussion from the issue: I don't think `target=_blank` would be needed if we already have `download`, because it is used for opening links in a different tab.

If you decide to document a button or link for downloading notebooks as a feature as well (and thereby close the mentioned issue) I would also suggest mentioning about how downloads of this manner will not work on cross-origin links, e.g., on PR builds on readthedocs because the URL would not be the same as that of the stable documentation. The reason is that it is a security measure in modern browsers (the origin of which was discussed in [this Bugzilla report](https://bugzilla.mozilla.org/show_bug.cgi?id=676619): a cross-origin download link could lead to users downloading an unverified file of the same name hosted on a different server with malicious intent.

mgeier:
> If you decide to document a button or link for downloading notebooks as a feature as well (and thereby close the mentioned issue)

What do you mean by "document"?
I'm suggesting here to add links to the docs, are you talking about some additional documentation?

> I would also suggest mentioning about how downloads of this manner will not work on cross-origin links, e.g., on PR builds on readthedocs because the URL would not be the same as that of the stable documentation.

But isn't the `.ipynb` file always in the same domain as the HTML? It's right next to it, in the same directory, isn't it?

The download link also works on the PR build on RTD: https://nbsphinx--751.org.readthedocs.build/en/751/

agriyakhetarpal:
> What do you mean by "document"?
I'm suggesting here to add links to the docs, are you talking about some additional documentation?

Yes, if you decide to add some additional documentation for how others can use this feature, that is

> But isn't the .ipynb file always in the same domain as the HTML? It's right next to it, in the same directory, isn't it?

> The download link also works on the PR build on RTD: https://nbsphinx--751.org.readthedocs.build/en/751/

That's strange. It does not work on the [PR build](https://pybamm--3076.org.readthedocs.build/en/3076/source/examples/notebooks/solvers/ode-solver.html) where I had added it myself. Maybe it is due to the extra `target="_blank"` attribute (your method contains just the `download` attribute)

mgeier:
> In your case, it downloads the notebook from the PR build itself, but our build leads us to a different link (it downloads the notebook from [docs.pybamm.org](https://docs.pybamm.org) instead). So the point about cross-origin links remains valid.

Yeah, but that's definitely not a workflow that I would recommend.

You are linking to the `latest` downloads, and those links will likely become invalid over time.

I would recommend to link to the notebooks of the exact same version as the doc build.
This should be easy, because those notebooks are created by `nbsphinx` anyway.
No need to worry about cross-origin links.

agriyakhetarpal:
That sounds like a good idea to prevent potential link rot. Thank you for the recommendation!

PR #772, opened 2023-12-21
Import CellExecutionError from nbconvert from the right place

HealthyPear:
newer versions of nbconvert now define CellExecutionError only from nbconvert.preprocessors

I guess also a test is missing which made this part of code never trigger the crash ( I get it because in my sphinx build something is triggering it)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     install_requires=[
         'docutils',
         'jinja2',
-        'nbconvert!=5.4',
+        'nbconvert>=5.3,!=5.4',
         'traitlets>=5',
         'nbformat',
         'sphinx>=1.8',
--- a/src/nbsphinx/__init__.py
+++ b/src/nbsphinx/__init__.py
@@ -616,7 +616,7 @@ class NotebookParser(rst.Parser):
 
         try:
             rststring, resources = exporter.from_notebook_node(nb, resources)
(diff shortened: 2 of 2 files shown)


mgeier:
Thanks for this PR!

If we want to be very clear, we should change the dependency in `setup.py` to `nbconvert>=5.3,!=5.4`, because this was added in https://github.com/jupyter/nbconvert/commit/c62f6e4cd5d666a435b4e59ded6fc35b1cb81f87.

BTW, I still have to figure out the CI failure, but it has nothing to do with your changes.

HealthyPear:
> Thanks for this PR!
> 
> If we want to be very clear, we should change the dependency in `setup.py` to `nbconvert>=5.3,!=5.4`, because this was added in [jupyter/nbconvert@c62f6e4](https://github.com/jupyter/nbconvert/commit/c62f6e4cd5d666a435b4e59ded6fc35b1cb81f87).
> 

will do

> BTW, I still have to figure out the CI failure, but it has nothing to do with your changes.

I see that the error has to do with the absence of `pkg_resources`.
As far as I know that dependency to deal with resources is deprecated in recent python versions in favor of `importlib_resources`.

HealthyPear:
The CI issues seem gone anyway

mgeier:
The CI error has been fixed (or rather worked-around) in #773.

Can you please re-base?

mgeier:
Thanks!

PR #807, opened 2024-08-11
Handle doc2path() API change

### LINKED ISSUE #804: Sphinx 9 compatibility
Trying to build our docs with Sphinx 8, I see:

```
…site-packages/nbsphinx/__init__.py:1345:
  RemovedInSphinx90Warning:
  Sphinx 9 will drop support for representing paths as strings. Use "pathlib.Path" or "os.fspath" instead.
```

The problem is this line apparently:
https://github.com/spatialaudio/nbsphinx/blob/b0b5274d686fa7b66d159e1645634c3ff6959774/src/nbsphinx/__init__.py#L1361

mgeier:
Hopefully fixes #804.


flying-sheep:
seems all a bit hacky.

I think the idea of the deprecation is that Sphinx eventually wants to return pathlib paths from these APIs.

So maybe a better fix would be to wrap the instances of these that represent concrete filesystem paths in `pathlib.Path` and using its APIs instead of string manipulation. But I didn’t take a deeper look here.

mgeier:
It sure is hacky, but contrary to the Sphinx maintainers, I value backwards compatibility, and I am somewhat confident that my PR here will continue to work with old versions (as well as new versions), while if I switch everything to `pathlib`, it will break when using old Sphinx versions, right?

flying-sheep:
no I mean wrapping the return values of that API in `Path(...)`

then you don’t need to use string operations on paths anymore

callezenwaka:
Hey @mgeier,

Is this PR getting merged soon? I have a failing pipeline and do not want to `filterwarnings` the error.

src/nbsphinx/__init__.py
```diff
@@ -577,7 +577,7 @@ def parse(self, inputstring, document):
         formats = {
             '.ipynb': lambda s: nbformat.reads(s, as_version=_ipynbversion)}
         formats.update(env.config.nbsphinx_custom_formats)
-        srcfile = env.doc2path(env.docname, base=None)
+        srcfile = str(env.doc2path(env.docname, base=None))
         for format, converter in formats.items():
             if srcfile.endswith(format):
```
mgeier, line 582: I cannot replace `.endswith(...)` with `Path.suffix`, because latter would not work for multi-suffixes like `.pct.py`.

I might be able to use `Path.suffixes`, but this seems cumbersome?

src/nbsphinx/__init__.py
```diff
@@ -1327,7 +1327,7 @@ class CreateSectionLabels(docutils.transforms.Transform):
 
     def apply(self):
         env = self.document.settings.env
-        file_ext = env.doc2path(env.docname, base=None)[len(env.docname):]
+        file_ext = str(env.doc2path(env.docname, base=None))[len(env.docname):]
```
mgeier, line 1330: I cannot replace this by `Path.suffix`, because it wouldn't work with multi-suffixes. And if I would make it work with multi-suffixes, it would still not work if the `env.docname` has a dot in it.

src/nbsphinx/__init__.py
```diff
@@ -1358,7 +1358,7 @@ class CreateDomainObjectLabels(docutils.transforms.Transform):
...
```
mgeier, line 1361: Same as above.

mgeier (commented):
@flying-sheep 

> no I mean wrapping the return values of that API in `Path(...)`
> 
> then you don’t need to use string operations on paths anymore

Thanks for the suggestion!

However, it turns out that except for a single instance, I actually need the string operations. I have added comments below with my reasoning.

And for the Jinja stuff I need strings anyway, because paths are not supported, right?

Please let me know if I'm missing something!

mgeier:
@callezenwaka 

> Is this PR getting merged soon? I have a failing pipeline and do not want to `filterwarnings` the error.

Yes, I'm planning to merge this soon, I just wanted to respond to @flying-sheep's comments first.

It would be great if you could test this on your project(s) and report if it works or not!

mgeier:
I'm going to merge this now, because people seem to be waiting for a fix, but I'm still open for changes which I can incorporate at a later time.

flying-sheep:
Thanks for responding! Sorry for holding this up with my curiosity!

mgeier:
No worries, I like getting feedback!

PR #812, opened 2024-09-04
feat: change markdown standard to GFM

tachyonicClock:
Change markdown standard to Github Flavored Markdown (GFM) replacing Pandoc's default markdown. This is consistent with the Jupyter Notebook markdown cells and should improve compatibility (https://nbformat.readthedocs.io/en/latest/format_description.html#markdown-cells).


mgeier:
Thanks for this PR!

I wasn't aware that Jupyter uses GFM. First I thought this was changed recently, but apparently this has been supported for a long time!

So I'm all for also using it in `nbsphinx`, but this PR seems to break a few things:

* LaTeX environments and macros like `\ref{}`
* HTML elements
* "warning" box within "note" box

This might not be a full list, for now I only looked at `markdown-cells.ipynb`.

douglas-raillard-arm:
Drive-by comment: have you considered also making a similar PR in jupyter nbconvert ? Both projects unfortunately do not share the pandoc invocation, but I think most users would expect them to behave the same way (at least I did).

PR #813, opened 2024-09-11
markdown: Fix bullet list parsing

douglas-raillard-arm:
Jupyter markdown allows bullet list to start right after the previous element, without any blank line in between. The default settings of pandoc is to not recognize that as a bullet list.

Therefore, enable the pandoc markdown extension that makes the blank line optional.


mgeier:
Thanks for this PR!

This will probably also be fixed by #812, but I think it would be good to merge this PR as well, because it will also make it work for Pandoc < 2.

src/nbsphinx/__init__.py
```diff
@@ -1043,7 +1043,9 @@ def filter_func(text):
     input_format += '-implicit_figures'
     v = nbconvert.utils.pandoc.get_pandoc_version()
     if nbconvert.utils.version.check_version(v, '1.13'):
-        input_format += '-native_divs+raw_html'
+        input_format += '-native_divs'
+        input_format += '+raw_html'
+        input_format += '+lists_without_preceding_blankline'
```
mgeier, line 1048: Why is this guarded with version 1.13+?

AFAICT this was already available earlier: https://github.com/jgm/pandoc/commit/777226296b04fa37094ecb07eb33f8d3e05af036

I think this can be added without version check, right?

mgeier: For the record, this is an earlier commit where the option was added: https://github.com/jgm/pandoc/commit/5afd373ae45f525ff1eff5e54c1850fe2c614b4b, see https://github.com/jgm/pandoc/issues/972.

douglas-raillard-arm: The changelog states that this extension was fixed in pandoc 1.13.2 (2014-12-20):
> Fix Ext_lists_without_preceding_blankline bug ([#1636](https://github.com/jgm/pandoc/issues/1636), Artyom).

https://pandoc.org/releases.html

So since that is what I tested is the "new" behavior and that 1.13 is already ancient with a check already in place, I just piggy backed on it. I don't think anyone would use that old version anyway so it probably will make little difference in practice, I'm happy to remove the check if you want.

mgeier: > this extension was fixed in pandoc 1.13.2

If we don't want to use it before the fix, the version check for `'1.13'` would strictly speaking not be sufficient, we would have to add a check for `'1.13.2'`.
But I think we can allow it before the fix anyway, because list with buggy spacing are better then no lists at all, right?

> I don't think anyone would use that old version anyway so it probably will make little difference in practice

I agree that it won't make any difference in practice, but given that we know that the feature was introduced much earlier, we might just as well put it before that version check.

If you think the check for `'1.13'` has become irrelevant, we can also remove it. But I don't think it hurts, and I have no problem with keeping the check in there, even if nobody might ever use that version again.

douglas-raillard-arm: I'll move it outside of the block then, if someone is stuck on an old version it's probably best indeed to have something a bit buggy rather than nothing.

douglas-raillard-arm:
PR updated to enable the option outside of the pandoc version check

mgeier:
Thanks!

Issue #815, opened 2024-09-24
Xarray dataset html repr is not correctly formatted


sandorkertesz:
If I have a notebook with an Xarray html repr the generated Sphinx html documentation using NBSphinx looks rather different containing a lot of empty space. This is a concrete example:

```python
import xarray as xr
url="https://psl.noaa.gov/thredds/dodsC/Datasets/noaa.oisst.v2/sst.mnmean.nc"
ds = xr.open_dataset(url)
ds
```

Notebook:
<img width="500" alt="image" src="https://github.com/user-attachments/assets/4a26b9ed-52b5-4a5f-8bbc-4322421fddda">

Generated Sphinx html documentation:
<img width="500" alt="image" src="https://github.com/user-attachments/assets/52262446-4a3c-4a21-b2af-a0fbd2bec42c">

This happened to all the Xarray datasets I tried in my various projects. 

My Sphinx configuration can be found here: 
https://github.com/ecmwf/earthkit-data/blob/develop/docs/conf.py
It is using this CSS file: https://github.com/ecmwf/earthkit-data/blob/develop/docs/_static/style.css
However, even when not using this CSS file the problem still occurs.

In my latest tests I used this environment (Python 3.10.13):
docutils                      0.20.1
nbclient                      0.8.0
nbconvert                     7.16.4
nbformat                      5.9.2
nbsphinx                      0.9.3
Pygments                      2.17.2
Sphinx                        7.3.7
sphinx-autoapi                3.0.0
sphinx-issues                 4.1.0
sphinx-rtd-theme              2.0.0
sphinxcontrib-apidoc          0.3.0
sphinxcontrib-applehelp       1.0.7
sphinxcontrib-devhelp         1.0.5
sphinxcontrib-htmlhelp        2.0.4
sphinxcontrib-jquery          4.1
sphinxcontrib-jsmath          1.0.1
sphinxcontrib-qthelp          1.0.6
sphinxcontrib-serializinghtml 1.1.9
xarray-2024.9.0

mgeier:
Thanks for this report!

Can you please provide an example that doesn't depend on SciPy/NetCDF that I can add to the docs (because I would like to avoid dependencies that aren't strictly necessary)?

mgeier:
Thanks for the example, I've added it in #817.

The display problems seem to be theme-dependent.

I currently don't have time to look into that, but maybe someone else will?

There are already some CSS overrides for the RTD theme, more can be added as needed:
[...]

PR #817, opened 2024-10-01
DOC: add section about Xarray datasets

mgeier:
See #815.


mgeier:
Rendered: https://nbsphinx--817.org.readthedocs.build/en/817/code-cells.html#Xarray-Datasets

PR #819, opened 2024-10-28
Use force=True with copyfile

### LINKED ISSUE #818: Possible bug: misc.copy_overwrite warning
Sphinx 8 introduced a change to [not overwrite user-supplied data by default when copying](https://github.com/sphinx-doc/sphinx/pull/12647). It changed two utility functions, `copyfile` and `copy_asset_file`. They now skip any files that would result in a file overwrite and emit a "misc.copy_overwrite" warning unless called with `force=True`.

Nbsphinx calls `sphinx.util.copyfile` directly and it calls `copy_asset_file` indirectly via `sphinx.util.fileutil.copy_asset`, but it does not pass `force=True`. 

I am not familiar enough with Sphinx or nbsphinx to know if force=True should be passed. All I know is this issue surfaced as a bug in my own development workflow. I was working on the [PyData Sphinx Theme repo](https://github.com/pydata/pydata-sphinx-theme), which aborts its docs build if it encounters a warning that it does not expect. 

Here is the full warning:

>WARNING: Aborted attempted copy from /dev/pydata-sphinx-theme/docs/_build/html/.doctrees/nbsphinx/examples/pydata.ipynb to /dev/pydata-sphinx-theme/docs/_build/html/examples/pydata.ipynb (the destination path has existing data). [misc.copy_overwrite]

The reason why I suspect that this might be a bug in nbsphinx is that the PyData Theme docs use a number of other extensions and none of them trigger this warning; it's only nbsphinx.

gabalafou:
Fixes #818.

### DIFF
--- a/src/nbsphinx/__init__.py
+++ b/src/nbsphinx/__init__.py
@@ -1691,10 +1691,11 @@ def html_page_context(app, pagename, templatename, context, doctree):
 
 def html_collect_pages(app):
     """This event handler is abused to copy local files around."""
-    # From version 8 of Sphinx, we need to pass force=True to the copyfile function
-    # but not to older version of the same function.
-    import inspect
-    use_force = bool(inspect.signature(sphinx.util.copyfile).parameters.get('force'))
+    # From version 8 of Sphinx, we need to pass force=True to the copyfile
+    # function but not to older version of the same function.
+    from inspect import signature
+    use_force = bool(
+        signature(sphinx.util.copyfile).parameters.get('force'))
     files = set()
     for file_list in app.env.nbsphinx_files.values():
         files.update(file_list)
@@ -1704,7 +1705,10 @@ def html_collect_pages(app):
         sphinx.util.ensuredir(os.path.dirname(target))
         try:
             kwargs = {"force": True} if use_force else {}
-            sphinx.util.copyfile(os.path.join(app.env.srcdir, file), target, **kwargs)
+            sphinx.util.copyfile(
+                os.path.join(app.env.srcdir, file),
+                target,
+                **kwargs)
         except OSError as err:
             logger.warning(
                 'Cannot copy local file %r: %s', file, err,
--- a/src/nbsphinx/__init__.py
+++ b/src/nbsphinx/__init__.py
@@ -1710,7 +1710,6 @@ def html_collect_pages(app):
     files = set()
     for file_list in app.env.nbsphinx_files.values():
         files.update(file_list)
-    overwrite = backwards_compat_overwrite()
     for file in status_iterator(files, 'copying linked files... ',
                                 sphinx.util.console.brown, len(files)):
         target = os.path.join(app.builder.outdir, file)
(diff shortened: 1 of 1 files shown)


gabalafou:
I manually checked this in a few different ways, but the general idea was to start with a repo where I could reproduce #818. Then:

- set sphinx to version 8, set nbsphinx to my modified version, observe that the bug described in #818 goes way
- set sphinx to version 7, set nbsphinx to my modified version, everything ok

mgeier:
Thanks for this PR!

The test failures are unrelated, I'm trying to find out what to do about them in https://github.com/sphinx-doc/sphinx/issues/13083.

src/nbsphinx/__init__.py
```diff
@@ -1691,6 +1691,11 @@ def html_page_context(app, pagename, templatename, context, doctree):
 
 def html_collect_pages(app):
     """This event handler is abused to copy local files around."""
+    # From version 8 of Sphinx, we need to pass force=True to the copyfile
+    # function but not to older version of the same function.
+    from inspect import signature
+    use_force = bool(
+        signature(sphinx.util.copyfile).parameters.get('force'))
```
mgeier, line 1698: Wouldn't it be nicer to use something like `'force' in signature(...).parameters` here?

src/nbsphinx/__init__.py
```diff
@@ -1699,7 +1704,11 @@ def html_collect_pages(app):
         target = os.path.join(app.builder.outdir, file)
         sphinx.util.ensuredir(os.path.dirname(target))
         try:
-            sphinx.util.copyfile(os.path.join(app.env.srcdir, file), target)
+            kwargs = {"force": True} if use_force else {}
```
mgeier, line 1707: You are defining `kwargs` at two places. Woudn't it be better to define it once above (instead of defining `use_force`) and then using it twice?

I this case `kwargs` would probably not be a good name. You could probably use something like `overwrite` and use it like this: `copyfile(..., **overwrite)`?

gabalafou:
Thanks for the pointers!

Re: test failures. That looks frustrating!

src/nbsphinx/__init__.py
```diff
@@ -1689,17 +1689,37 @@ def html_page_context(app, pagename, templatename, context, doctree):
...
     for file in status_iterator(files, 'copying linked files... ',
                                 sphinx.util.console.brown, len(files)):
         target = os.path.join(app.builder.outdir, file)
         sphinx.util.ensuredir(os.path.dirname(target))
         try:
-            sphinx.util.copyfile(os.path.join(app.env.srcdir, file), target)
+            sphinx.util.copyfile(
+                os.path.join(app.env.srcdir, file),
+                target,
+                **overwrite)
```
mgeier, line 1722: Now that we have this as a function, you could get rid of the additional line above by directly calling the function here (and again below):

```suggestion
                **backwards_compat_overwrite())
```

What do you think about this?

mgeier:
Thanks!

zmbc:
Would love to see this released 😄

mgeier:
I was hoping for some movement in https://github.com/sphinx-doc/sphinx/issues/13083, but we don't have to wait for this.

Here's your release: https://pypi.org/project/nbsphinx/0.9.6/.

PR #822, opened 2025-02-17
Adapt ``ForceEquations`` for Sphinx 8.2

AA-Turner:
cc @mgeier xref https://github.com/sphinx-doc/sphinx/issues/13346

A

### DIFF
--- a/src/nbsphinx/__init__.py
+++ b/src/nbsphinx/__init__.py
@@ -27,6 +27,7 @@ import sphinx.directives
 import sphinx.directives.other
 import sphinx.environment
 import sphinx.errors
+import sphinx.transforms
 import sphinx.transforms.post_transforms.images
 from sphinx.util.matching import patmatch
 try:
@@ -1458,7 +1459,7 @@ class CopyLinkedFiles(docutils.transforms.Transform):
             env.nbsphinx_files.setdefault(env.docname, []).append(file)
 
 
-class ForceEquations(docutils.transforms.Transform):
+class ForceEquations(sphinx.transforms.SphinxTransform):
     """Unconditionally enable equations on notebooks.
 
     Except if ``nbsphinx_assume_equations`` is set to ``False``.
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR!

This works in most cases, but sadly not in all.

The problem is that there are multiple situations where math/LaTeX stuff can appear:

* Markdown cells, inline math
* Markdown cells, block math
* Code output cells

All three are handled separately and it's quite fiddly.

I tested it in #824 and this shows the problem with Markdown block math: https://nbsphinx--824.org.readthedocs.build/en/824/a-markdown-file-block.html

I don't know yet why that happens (I've actually expected problems with Markdown inline math, but that seems to work fine), and I have limited time to investigate ...

AA-Turner:
> I don't know yet why that happens (I've actually expected problems with Markdown inline math, but that seems to work fine), and I have limited time to investigate ...

The new behaviour includes MathJax if a `math` or `math_block` node is traversed in the document. Does the custom markdown parsing here construct `math` nodes?

A

mgeier:
> The new behaviour includes MathJax if a `math` or `math_block` node is traversed in the document. Does the custom markdown parsing here construct `math` nodes?

It turns out that the block math (that doesn't work correctly) ends up in custom `` `:nbsphinx-math:`...` `` roles, which I've defined there:

https://github.com/spatialaudio/nbsphinx/blob/215bf0d9b2108c739a224736d8763dffc91f3479/src/nbsphinx/__init__.py#L648-L650

How can I make those be recognized as math?

> Alternatively, you could call [`app.set_html_assets_policy('always')`](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.set_html_assets_policy), which forcibly includes the mathjax script in every page.

OK, thanks, I think I can use this as work-around (slightly worse than the current work-around) in case we don't find a better solution.

src/nbsphinx/__init__.py
```diff
@@ -1468,9 +1469,15 @@ class ForceEquations(docutils.transforms.Transform):
     default_priority = 900  # after checking for equations in MathDomain
 
     def apply(self):
-        env = self.document.settings.env
-        if env.config.nbsphinx_assume_equations:
-            env.get_domain('math').data['has_equations'][env.docname] = True
+        if sphinx.version_info[:2] >= (8, 2):
```
mgeier, line 1472: Out of curiosity: Why are you putting `[:2]` there?
Shouldn't it behave identically when you drop it?

gabalafou:
I created an alternative solution:

- #840

PR #823, opened 2025-02-17
Specify sphinx.configuration for RTD

mgeier:
... since this is now required.


mgeier:
I have included this in #828.

PR #828, opened 2025-03-02
Disable Sphinx 8.2

mgeier:
In general, I think pinning upper version bounds is not the right approach, but given the slew of incompatibilities introduced by Sphinx (in part involving 3rd party dependencies), I think it is best for my sanity and for most casual `nbsphinx` users to restrict Sphinx upgrades for now.

I hope it will become easier to support Sphinx upgrades at a later point.

This PR includes #823 in its entirety and parts of #826.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,8 @@ setup(
         'nbconvert>=5.3,!=5.4',
         'traitlets>=5',
         'nbformat',
-        'sphinx>=1.8',
+        # https://github.com/spatialaudio/nbsphinx/issues/825
+        'sphinx >= 1.8, < 8.2',
     ],
     author='Matthias Geier',
     author_email='Matthias.Geier@gmail.com',
--- a/.github/workflows/version-matrix.yml
+++ b/.github/workflows/version-matrix.yml
@@ -36,12 +36,11 @@ jobs:
 
           # a few Python versions using latest Sphinx release
           - env:
-              PYTHON_VERSION: "3.8"
-          - env:
               PYTHON_VERSION: "3.10"
           - env:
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -29,7 +29,8 @@
         'nbconvert>=5.3,!=5.4',
         'traitlets>=5',
         'nbformat',
-        'sphinx>=1.8',
+        # https://github.com/spatialaudio/nbsphinx/issues/825
+        'sphinx >= 1.8, < 8.2',
```
AA-Turner, line 33: See my comment in https://github.com/spatialaudio/nbsphinx/issues/825#issuecomment-2692944098, I think that Sphinx 8.2.2 shoud now work with `nbsphinx` (though the explicit imports in #826 remain the correct solution)
```suggestion
        'sphinx >= 1.8, != 8.2.0, != 8.2.1',
```

jenshnielsen: @AA-Turner @mgeier I can confirm that sphinx 8.2.2 passes with current nbsphinx for now

mgeier: Thanks for the update!

I'm still getting a warning related to https://github.com/felix-hilden/sphinx-codeautolink/issues/173.

I will un-pin Sphinx when this is done. I might also re-think the testing strategy for warnings, because keeping up with them is quite time-consuming.

I have just made a new `nbsphinx` release which still blocks Sphinx 8.2+.

AA-Turner: Indeed, to Matthias' point about testing strategy for warnings, enabling deprecation warnings for the documentation build (or non-core dependencies) probably isn't needed and may be unhelpful in this case.

Issue #417, opened 2020-03-24
vertical scrollbars for long code cells 


mluerig:
when cells with a lot of output (e.g. a long docstring) gets converted to html, the vertical scrollbar is missing - can I change that behavior?

seems like it's possible via nbconvert
https://stackoverflow.com/questions/31700364/when-exporting-an-ipython-notebook-to-html-long-output-doesnt-have-a-scroll-ba

mgeier:
This is not yet implemented, see #372.

Would you like to implement this?

mluerig:
I could give it a try, although I have no idea where to start

mgeier:
You could start by coming up with some CSS settings that implement the scrolling behavior (probably similar to `nbconvert` ... or not).

After that, we only need to make `nbsphinx` emit a certain CSS class for the right output cells. I can help you with that, if you want.
But for starters you could apply it to every single output cell, this way you don't have to worry about that part for now.

mgeier:
Thanks @tautomer!
Would you like to turn this into a PR?
It would also help to have an example (or several) for a long code cell in one of the notebooks in the docs.

Issue #372, opened 2020-01-08
HTML: Implement support for "scrolled" metadata


mgeier:
JupyterLab has a context menu item "Enable Scrolling for Outputs", which is stored in the `scrolled` field in the cell metadata.

I guess it would make sense to implement this in `nbsphinx`, too.

PR #831, opened 2025-03-27
Force long output text to scroll

bgmeulem:
Currently, long cell output does not scroll. This has been adressed in issues #417 and #372.

These cells are already correctly configured to scroll upon overflow from various parent classes.
The reason these outputs don't scroll vertically is simply because there is no limit on how much vertical space they are allowed to take, so they never overflow.

This PR simply implements a maximum height for all `nboutput container` classes, but only for those that contain text.
Checking if they contain text is done by checking if they have a child with `pre`. I'm not very accustomed to CSS, so I don't know if this is a good way to check.

### DIFF
--- a/src/nbsphinx/_static/nbsphinx-code-cells.css_t
+++ b/src/nbsphinx/_static/nbsphinx-code-cells.css_t
@@ -128,6 +128,11 @@ div.nboutput.container div.output_area {
         width: 100%;
     }
 }
+/* scroll long text outputs vertically */
+div.nboutput.container div.output_area:has(pre) {
+    max-height: 600px;
+}
+
 
 /* input area */
 div.nbinput.container div.input_area {


mgeier:
Thanks for this PR!

I'm not sure if an unconditional fixed threshold is the right thing to do, though.

The Jupyter Notebook has already "scrolled" metadata (see #372), so why not use this?

> I chose `600px` rather arbitrarily

Why not use whatever the Jupyter Notebook uses?

mgeier:
If you want to avoid the (unrelated) CI errors (which I fixed in #835), you can rebase this PR.

Xmaster6y:
I personally like this simple solution. Yet it should not be merged.

You can achieve the same using custom CSS:

```py
html_static_path = ["_static"]
html_css_files = [
    "css/nbsphinx.css",
]
```

```css
/*nbsphinx.css*/
div.nboutput.container div.output_area:has(pre) {
    max-height: 600px;
}
```

PR #836, opened 2025-05-01
DOC: add separate examples for inline, block and cell output math

mgeier:
This adds examples that show the problem that #822 doesn't quite solve.

They will also serve as regression tests once it does work.

A possible solution is to use [`app.set_html_assets_policy('always')`](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.set_html_assets_policy), as suggested in https://github.com/spatialaudio/nbsphinx/pull/822#issuecomment-2664315587.

### DIFF
--- a/doc/gallery/cinque-rst.ipynb
+++ b/doc/gallery/cinque-rst.ipynb
@@ -13,9 +13,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


mgeier:
This page shows the broken equation: https://nbsphinx--836.org.readthedocs.build/en/836/gallery/uno-rst.html

Issue #15, opened 2016-01-20
Hiding code input but not output?


jhconning:
The documentation has a nice discussion of [how to hide code cells and their output entirely](http://nbsphinx.readthedocs.org/en/0.2.2/hidden-cells.html).  It is however also frequently useful to hide *just the code* and *not the output* for some or all code-cells in a notebook (e.g. to make an HTML only report).  I've seen some rendered html that seems to go further by offering code-folding buttons on the rendered html.

  Is there a way to *hide just the code input* but not the output with nbsphinx?   For reference, with jupyter's nbconvert one can use templates and a command as such:
```
jupyter nbconvert your_notebook.ipynb --to html --template output_toggle_html
```
where the template is like the one adapted from [this SO answer](http://stackoverflow.com/questions/19524554/suppress-code-in-nbconvert-ipython)

mgeier:
I've heard and read about this request (with regards to `nbconvert`) a few times, it seems quite common.

I thought about implementing this in `nbsphinx`, but I wanted to wait until it's supported in the Jupyter live notebook and in `nbconvert`, because I wanted to re-use whatever notebook/cell metadata they are using.

I'm not really a fan of removing code completely, but I think the code-folding option would be great.

I have no clue how to do such a thing in HTML/CSS, can you provide an example for that?

Do you have an idea how this could be realized in LaTeX?
Folding would work in the literal sense on a piece of paper, but I don't think this would be a practical solution (except if you have some kind of Origami printer).
Probably the code could be collected and referenced in a "list of code cells" similar to a "list of figures"?

As a quick hack to remove all input cells, you can search for "block input" in `nbsphinx.py` and remove the following lines up to (but not including) the line with "endblock input".
Theoretically, you can also change this in your own template file called `nbsphinx-rst.tpl` in the `_templates/` directory with the following content (but I didn't test if that really works):

```jinja
{% extends 'nbsphinx-rst.tpl' %}
{% block input %}
{% endblock input %}
```

This only looses a little bit of CSS, namely `padding-top: 5px;`, which you should move to `div.nboutput`.

Completely removing the input on a per-notebook and per-cell basis would be simple, I'd just have to extend the supported notebook/cell metadata.
As I said, I would like to wait for the "official" metadata, but if you can't wait, we'd have to come up with `nbsphinx`-specific metadata for the time being. Any suggestions?

jhconning:
Thanks for the reply and sorry for my slow response. 

[...]

PR #837, opened 2025-05-17
Add option to hide cell input of a single cell using its metadata.

BenZickel:
Resolves https://github.com/spatialaudio/nbsphinx/issues/15.

### DIFF
--- a/doc/hidden-cells.ipynb
+++ b/doc/hidden-cells.ipynb
@@ -79,6 +79,17 @@
    "source": [
     "This is the cell after the hidden cell."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can also remove just the cell input from the HTML/LaTeX output by adding this to the cell metadata:\n",
+    "\n",
+    "```\n",
+    "\"nbsphinx\": \"hidden-input\"\n",
+    "```"
+   ]
   }
  ],
  "metadata": {
--- a/src/nbsphinx/__init__.py
+++ b/src/nbsphinx/__init__.py
@@ -125,6 +125,7 @@ RST_TEMPLATE = """
 
 
 {% block input -%}
+{%- if cell.metadata.nbsphinx != 'hidden-input' %}
 .. nbinput:: {% if cell.metadata.magics_language -%}
 {{ cell.metadata.magics_language }}
 {%- elif nb.metadata.language_info -%}
@@ -141,6 +142,7 @@ RST_TEMPLATE = """
 {%- endif %}
 
 {{ cell.source.strip('\n') | indent }}
+{%- endif %}
 {% endblock input %}
 
 


mgeier:
Thanks for this PR!

Can you please address the comments I made in #15?

BenZickel:
I've reviewd #15 and some of its references and I think that this PR is the least disruptive from all other propositions, as it utilizes the existing metadata keyword `nbsphnix` which is already in use.

My specific use case is for creating documentation using `nbsphnix` in which there are input cells which are not informative and confuse the reader. For example when generating diagrams, the reader is not interested in the code used to geenrate the diagram. This use case was also mentioned in https://github.com/spatialaudio/nbsphinx/pull/185#issuecomment-393275752.

If there are any other comments I'll make additional changes as required, and thanks for creating `nbsphnix`!

mgeier:
> ... there are input cells which are not informative and confuse the reader. For example when generating diagrams, the reader is not interested in the code used to geenrate the diagram.

Are you sure? I'm often very much interested in how a good-looking diagram was generated.

I agree that putting a whole blob of unstructured plotting code into a huge code cell is likely distracting to readers. I recommend moving most of the auxiliary code into functions in a separate `.py` file (or multiple) which you can then `import` in a code cell.

I'm doing this all the time, here's a random example: https://splines.readthedocs.io/en/0.3.2/euclidean/catmull-rom-non-uniform.html#Animation

This way you are not hiding how to generate the diagram (and make it easier for a reader to understand), but you are still avoiding (most of) the noise.

> If there are any other comments I'll make additional changes as required

Yes, there are many comments in #15, I don't want to copy all of them here, but for starters, here is one (from https://github.com/spatialaudio/nbsphinx/issues/15#issuecomment-173515626):

>> I thought about implementing this in `nbsphinx`, but I wanted to wait until it's supported in the Jupyter live notebook and in `nbconvert`, because I wanted to re-use whatever notebook/cell metadata they are using.

Or what about this one (from https://github.com/spatialaudio/nbsphinx/issues/15#issuecomment-515746378):

>>  I think the most reasonable course of action is to limit this feature to HTML output and implement a "toggle" button similar to the one used in https://jupyter.org/jupyter-book/features/hiding.

Admittedly, the link is broken, let me know if you don't find what I meant at that time.

mgeier:
> If you are creating them using dedicated `.py` files, that you don't mind users reading them, ...

On the contrary, I definitely want users to be able to read them.

> ... you would need to point to the source code of the notebook and dedicated `.py` files anyhow, ...

Yes, and I consider this good practice.

> ... in which case all cells are shown, including hidden and removed ones.

That's true, but I'd like to make it as easy as possible for readers. Having to open and sift through the source notebook is just an additional obstacle that I would like to avoid.

Also, if there is no hint in the HTML-rendered version (because the input cell was hidden), readers might not know that there's even something hidden from them they could discover. 

> Would using a different tag for hiding (or should I say removing) an input cell be sufficient (see for example https://jupyterbook.org/en/stable/interactive/hiding.html#removing-code-cell-content)?

It would very helpful if you could assemble a list of which tags are used in different tools to hide/remove which kinds of cells. This list could then be a basis for deciding what to do in `nbsphinx` and it could also go into the documentation.

Also, it would help if you could find out what this means (and whether HTML is considered an "exported format"):

https://github.com/jupyter/nbformat/blob/b950f02b9e898b2fb3c7c959e385607942b3c2b8/CHANGELOG.md?plain=1#L390-L391

```
  These fields should not be used to hide elements in exported
  formats.
```

PR #853, opened 2025-11-11
Update default version of require.js

### LINKED ISSUE #851: Update default version of require.js
It seems like that the default version of require.js hardcoded in the plugins source code leads to some issues, see https://github.com/spatialaudio/nbsphinx/issues/678 and https://github.com/SAP/swagger-plugin-for-sphinx/issues/343. For me it helped to update the version to 2.3.7 (latest one). While its possible to do this via the configuration, it would be great if you could update the default

kasium:
Closes #851

### DIFF
--- a/doc/code-cells.ipynb
+++ b/doc/code-cells.ipynb
@@ -728,7 +728,7 @@
     "\n",
     "```html\n",
     "<!-- require.js is a mandatory dependency for jupyter-widgets -->\n",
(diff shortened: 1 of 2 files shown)


mgeier:
Thanks for this PR!

Sadly, it doesn't seem to work, see the rendered documentation of "widgets": https://nbsphinx--853.org.readthedocs.build/en/853/code-cells.html#Interactive-Widgets-(HTML-only)

There's an error in the JavaScript console that's probably related:

```
Uncaught ReferenceError: define is not defined
    <anonymous> https://cdn.jsdelivr.net/npm/@jupyter-widgets/html-manager@^1.0.1/dist/embed-amd.js:2
```

kasium:
Mhm, there is no changelog and the changes don't look related: https://github.com/requirejs/requirejs/compare/2.3.4...2.3.7

However when lookin into other PRs like https://github.com/spatialaudio/nbsphinx/pull/849 I think your CI is broken :-)

PR #872, opened 2026-04-07
Log execution and time spend executing for all notebooks executed

jenshnielsen:
While trying to optimize the docs build time for my documentation it was useful for me to log the time taken to execute notebooks to identify notebooks that are slower than they should be. 

I don't know if this is useful for others but I figured I would open a pr in case it is.

The output can be activated by passing `-v` to sphinx when building the docs

### DIFF
--- a/src/nbsphinx/__init__.py
+++ b/src/nbsphinx/__init__.py
@@ -15,6 +15,7 @@ import os
 import re
 import subprocess
 import sys
+import time
 from urllib.parse import unquote
 import uuid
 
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks for this PR! I think this is very useful, however ... are you aware of https://www.sphinx-doc.org/en/master/usage/extensions/duration.html?

This seems to provide similar information for the 5 slowest notebooks (or other Sphinx sources). When setting `duration_n_slowest = 0`, this is supposed to show all durations, but that didn't work for me.

The displayed durations seem to be a bit longer than the ones from this PR, I guess this includes a bit more than just executing the notebooks.

jenshnielsen:
@mgeier Thanks for the pointer I was not aware of that!

jenshnielsen:
@mgeier I tested the durations extension and both the duration_n_slowest config and the json output seems to work for me using sphinx >=9 (I can't use that for my actual build due to an unrelated bug but that is not important in this context) 
I don't think there is much to gain from merging this so I will go ahead and close. Thanks for the pointer again

